PR #26, opened 2023-06-10
feat(api): add total aggregate

tothandras:
When `windowSize` query parameter is not defined, the total aggregate will be returned for the specified period for each subject and group.

**BREAKING CHANGES**
- Changes the behavior of the `windowSize` query param from being the default of the meter's window size.
- Changed the response field `values` to `data`

Possible improvements:
- specify `groupBy` query parameter for groups to be ignored or selected

TODO:
- update examples

### DIFF
--- a/api/openapi.yml
+++ b/api/openapi.yml
@@ -112,6 +112,8 @@ paths:
         - name: windowSize
           in: query
           required: false
+          description: |
+            If not specified, a single usage aggregate will be returned for the entirety of the specified period for each subject and group.
           schema:
             $ref: "#/components/schemas/WindowSize"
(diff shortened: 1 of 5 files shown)


api/openapi.yml
```diff
@@ -112,6 +112,8 @@ paths:
         - name: windowSize
           in: query
           required: false
+          description: |
+            If not specified, a single usage aggregate will be returned for the entirety of the specified period for each subject and group.
```
hekike, line 116: I think a separate endpoint would be more devex-friendly. But maybe an SDK can have multiple methods and abstract the logic away.

PR #70, opened 2023-06-21
feat(api): add healthcheck

hekike:
## Overview

Add GET /healthcheck endpoint.

Ideally, in the future, we would separate health checks of ingestion and query.
Ingestion only depends on `kafka`, query depend on `ksql` and later on `postgres`.
In a production system we want a separate  failover for ingestion and query instances.


api/openapi.yml
```diff
@@ -138,6 +138,20 @@ paths:
...
+      operationId: healthcheck
+    responses:
+      "200":
+        description: All checkers were in success
+      default:
+        description: Unexpected error
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/Error"
```
sagikazarmark, line 154: I doubt we want to expose this in the public API. I'd rather expose it on the telemetry interface since it's an internal endpoint.

PR #96, opened 2023-07-06
feat: support batch event ingestion

### LINKED ISSUE #54: Support batch event ingestion
## Problem

Currently, we only support [event ingestion](https://github.com/openmeterio/openmeter/blob/main/api/openapi.yml#L10) one at a time.

## Solution

- The cloudevents specification allows [batched HTTP content mode](https://github.com/cneijenhuis/spec/blob/d3f2edbd4837dbc6194aa418059d820669392523/http-transport-binding.md#33-batched-content-mode).
- Update the [OpenAPI document](https://github.com/openmeterio/openmeter/blob/main/api/openapi.yml#L17) to include `Content-Type: application/cloudevents-batch+json` with a body of an array of events.
- Use the JSON decoder for the content type: https://github.com/openmeterio/openmeter/blob/main/internal/server/router/router.go#L24
- Update the request handler to unmarshal the body based on the content-type: https://github.com/openmeterio/openmeter/blob/main/internal/server/router/router.go#L53
- Publish the events one by one (transaction handling would be nice, can be implemented later)

zsomborjoel:
Hi! 
I'm new to open source and Go. I would like to try to contribute to your project through this PR based on _'good first issue'_ tag. :)

Please feel free to review my changes and request additional. Thank you!

## Overview

-  Updated the code to include Content-Type: application/cloudevents-batch+json with a body of an array of events
-  Updated the request handler (on http ingestion level) to unmarshal the body based on the content-type
-  Published the events one by one (no transaction handling yet)
-  Test included 

Fixes #54 

## Notes for reviewer

Context was moved to Handler in order to use it globally for logger.

### DIFF
--- a/internal/ingest/httpingest/httpingest.go
+++ b/internal/ingest/httpingest/httpingest.go
@@ -1,6 +1,7 @@
 package httpingest
 
 import (
+	"context"
 	"encoding/json"
 	"net/http"
 	"time"
@@ -16,7 +17,8 @@ import (
 type Handler struct {
 	Collector Collector
 
-	Logger *slog.Logger
+	Logger  *slog.Logger
+	Context context.Context
 }
 
 // Collector is a receiver of events that handles sending those events to some downstream broker.
@@ -26,17 +28,65 @@ type Collector interface {
 
 func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	logger := h.getLogger()
+	h.Context = r.Context()
 
+	contentType := r.Header.Get("Content-Type")
+
+	if contentType == "application/cloudevents-batch+json" {
+		err := h.processBatchRequest(w, r)
+		if err != nil {
+			logger.ErrorCtx(h.Context, "unable to process batch request", "error", err)
+			_ = render.Render(w, r, api.ErrInternalServerError(err))
+			return
+		}
+	} else {
+		err := h.processSingleRequest(w, r)
+		if err != nil {
+			logger.ErrorCtx(h.Context, "unable to process single request", "error", err)
+			_ = render.Render(w, r, api.ErrInternalServerError(err))
+			return
+		}
+	}
+
+	w.WriteHeader(http.StatusOK)
+}
+
+func (h Handler) processBatchRequest(w http.ResponseWriter, r *http.Request) error {
+	var events []event.Event
+
+	err := json.NewDecoder(r.Body).Decode(&events)
+	if err != nil {
+		return err
+	}
+
+	for _, event := range events {
+		err = h.processEvent(event)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (h Handler) processSingleRequest(w http.ResponseWriter, r *http.Request) error {
 	var event event.Event
 
 	err := json.NewDecoder(r.Body).Decode(&event)
 	if err != nil {
-		logger.ErrorCtx(r.Context(), "unable to parse event", "error", err)
-
-		_ = render.Render(w, r, api.ErrInternalServerError(err))
-
-		return
+		return err
 	}
+
+	err = h.processEvent(event)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (h Handler) processEvent(event event.Event) error {
+	logger := h.getLogger()
 
 	logger = logger.With(
 		slog.String("event_id", event.ID()),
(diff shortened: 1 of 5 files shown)


internal/ingest/httpingest/httpingest.go
```diff
@@ -16,7 +17,8 @@ import (
 type Handler struct {
 	Collector Collector
 
-	Logger *slog.Logger
+	Logger  *slog.Logger
+	Context context.Context
```
tothandras, line 21: I think it would be better to pass the context to `processEvent` as the first argument and use `r.Context` in both `processBatchRequest`, and `processSingleRequest`.

internal/ingest/httpingest/httpingest.go
```diff
@@ -26,42 +28,84 @@ type Collector interface {
...
+			return
+		}
+	} else {
+		err := h.processSingleRequest(w, r)
+		if err != nil {
+			logger.ErrorCtx(h.Context, "unable to process single request", "error", err)
+			_ = render.Render(w, r, api.ErrInternalServerError(err))
+			return
+		}
+	}
```
tothandras, line 49: I'm not sure about this, @sagikazarmark ?

```suggestion
	var err error
	switch contentType {
	case "application/cloudevents+json":
		err = h.processSingleRequest(w, r)
	case "application/cloudevents-batch+json":
		err = h.processBatchRequest(w, r)
	default:
		_ = render.Render(w, r, api.ErrUnsupportedMediaType(errors.New("content type must be application/cloudevents+json or application/cloudevents-batch+json")))
	}

	if err != nil {
		logger.ErrorCtx(h.Context, "unable to process request", "error", err)
		_ = render.Render(w, r, api.ErrInternalServerError(err))
		return
	}
```

tothandras (changes requested):
Thank you for the contribution! Unfortunately, there is a bug in the OpenAPI generator. We need for it to be fixed before continuing. https://github.com/deepmap/oapi-codegen/issues/1127

internal/ingest/httpingest/httpingest.go
```diff
@@ -26,42 +28,84 @@ type Collector interface {
...
+	if err != nil {
+		return err
+	}
+
+	for _, event := range events {
+		err = h.processEvent(event)
+		if err != nil {
+			return err
+		}
+	}
```
sagikazarmark, line 67: This is probably OK for now, but generally batch ingestion endpoints are supposed to be able to process events asynchronously.

internal/ingest/httpingest/httpingest.go
```diff
@@ -26,42 +28,84 @@ type Collector interface {
 
 func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	logger := h.getLogger()
+	h.Context = r.Context()
```
sagikazarmark, line 31: This isn't thread safe. Context should be passed to the `process*` functions.

sagikazarmark (changes requested):

sagikazarmark:
Thanks @zsomborjoel! This is a great start. I generally agree @tothandras's comments. I had a few of my own.

internal/ingest/httpingest/httpingest.go
```diff
@@ -32,20 +32,20 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-			return
-		}
+	var err error
+	switch contentType {
+	case "application/cloudevents+json":
+		err = h.processSingleRequest(w, r)
+	case "application/cloudevents-batch+json":
+		err = h.processBatchRequest(w, r)
+	default:
+		_ = render.Render(w, r, api.ErrUnsupportedMediaType(errors.New("content type must be application/cloudevents+json or application/cloudevents-batch+json")))
```
tothandras, line 42: You need to add `ErrUnsupportedMediaType` to the `api/error.go`:
```go
func ErrUnsupportedMediaType(err error) *ErrResponse {
	return &ErrResponse{
		Err:        err,
		StatusCode: 415,
		StatusText: http.StatusText(415),
		Message:    err.Error(),
	}
}
```

internal/ingest/httpingest/httpingest_test.go
```diff
@@ -64,3 +65,43 @@ func TestHandler(t *testing.T) {
...
+	client := server.Client()
+
+	var events []event.Event
+	for i := 1; i <= 10; i++ {
+		id := strconv.Itoa(i)
+
+		event := event.New()
+		event.SetID("id" + id)
+		event.SetSubject("sub" + id)
+		event.SetSource("test" + id)
```
tothandras, line 84: You can use `fmt.Sprintf("id%s", id)` instead of strconv and concat.

sagikazarmark:
Looks good to me other than the above issue with the code generator mentioned by @tothandras.

Let me see if I can fix it.

tothandras:
@zsomborjoel Seems like the bug is fixed in https://github.com/deepmap/oapi-codegen/issues/1127.

zsomborjoel:
> @zsomborjoel Seems like the bug is fixed in [deepmap/oapi-codegen#1127](https://github.com/deepmap/oapi-codegen/issues/1127).

Thank you! Do you need me to do any more change? I have added the previously requested ones.

PR #97, opened 2023-07-07
feat(config): add auth support

hekike:
Support Kafka, KSQL, and Schema Registry authentications.

**Breaking change:**

- `ingest.kafka.schemaRegistry` => `schemaRegistry.url`

### DIFF
--- a/config.go
+++ b/config.go
@@ -29,16 +29,28 @@ type configuration struct {
 	// Ingest configuration
 	Ingest struct {
 		Kafka struct {
-			Broker         string
-			Partitions     int
-			SchemaRegistry string
+			Broker           string
+			SecurityProtocol string
+			SaslMechanisms   string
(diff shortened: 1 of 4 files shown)


hekike:
Note: config logging needs improvement to avoid logging secrets

sagikazarmark:
Not sure I understand why the schema registry config was moved out of the ingest section. As far as I can see, it's still exclusively used in the kafka ingest collector.

hekike:
Schema registry is a separate service primarily leveraged by the Kafka ecosystem (in this case Kafka client and ksqlDB) but it's technically independent from Kafka and could be used without it, for example: https://docs.aws.amazon.com/glue/latest/dg/schema-registry.html

hekike:
Discussed that schema registry can be more generic in future

PR #99, opened 2023-07-07
feat(schema): improve json schemas

hekike:
Add metadata to JSON schemas.


internal/ingest/kafkaingest/schema/event_key.json
```diff
@@ -1,4 +1,7 @@
 {
-    "$schema": "http://json-schema.org/draft-04/schema#",
-    "type": "string"
-}
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://openmeter.io/event_key.schema.json",
+    "type": "string",
+    "title": "Event Key",
+    "description": "Event Key Schema"
+}
```
hekike, line 7: weird, autoformat removes new line at end

internal/ingest/kafkaingest/schema/event_key.json
```diff
@@ -1,4 +1,7 @@
 {
-    "$schema": "http://json-schema.org/draft-04/schema#",
-    "type": "string"
-}
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "$id": "https://openmeter.io/event_key.schema.json",
```
tothandras, line 3: Do we need to serve these on their url?

hekike: Good questions. Maybe we should set them to their relative path in the repo, see:
https://json-schema.org/understanding-json-schema/structuring.html#id

hekike: Also, maybe it's fine:

> Even though schemas are identified by URIs, those identifiers are not necessarily network-addressable. They are just identifiers. Generally, implementations don’t make HTTP requests (https://) or read from the file system (file://) to fetch schemas. Instead, they provide a way to load schemas into an internal schema database. When a schema is referenced by it’s URI identifier, the schema is retrieved from the internal schema database.

tothandras (approved):

PR #116, opened 2023-07-20
Update OpenAPI

tothandras:
## Overview

- Update OpenAPI definition
- Move to `/v1alpha2/`
- Use RFC-7807 compliant errors
- Better `operationId`s
- Use meter `slug` and `id`
- Use named `groupBy`s for meters (closes #117)
- And more...

## TODO

- [x] Generate clients
  - [x] Publish node client to npm
- [x] Update examples
- [x] Update documentations


internal/server/logger.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+}
+
+func LogEntrySetFields(r *http.Request, fields map[string]interface{}) {
+	if entry, ok := r.Context().Value(middleware.LogEntryCtxKey).(*StructuredLoggerEntry); ok {
+		for k, v := range fields {
+			entry.logger = entry.logger.With(k, v)
+		}
+	}
+}
```
sagikazarmark, line 127: I know it's fairly contested in the community, but a majority of trustable sources suggests this practice is as bad as it gets. It abuses the context as a request scoped service locator.

The primary reason why the log handler receives the context is that it can unpack potential details necessary for logs _from_ the context.

The best practice is to inject the logger (slog.Logger) wherever necessary and use a handler that's capable of extracting relevant information from the context.

sagikazarmark (commented):
Note my strong objection to injecting the logger into the context.

The rest looks good to me.

hekike (approved):

PR #117, opened 2023-07-21
feat(meter): named group by

hekike:
Support named group bys to improve query DevEx.

Example with two types of group bys, simplified list and named:

```yaml
meters:
  - slug: m1
    name: Meter 1
    eventType: api-calls
    valueProperty: $.duration_ms
    aggregation: SUM
    groupBy:
      method: $.method
      path: $.path
  - slug: m2
    name: Open AI Tokens
    eventType: openai
    valueProperty: $.total_tokens
    aggregation: SUM
    groupBy:
      - model
```

Depends on: https://github.com/openmeterio/openmeter/pull/116


hekike:
Quickstart CI fails because query param `window_size` in OpenAPI doesn't match `windowSize` in `verify.sh`. 
This should be addressed in https://github.com/openmeterio/openmeter/pull/116

hekike:
Superseded by: https://github.com/openmeterio/openmeter/pull/116

PR #122, opened 2023-07-28
Add namespace support

hekike:
Continuation of https://github.com/openmeterio/openmeter/pull/121

I feel some disconnect between Kafka connector and Namespace Manager.
I think I would prefer the connector to handle namespaces.


internal/server/router/router.go
```diff
@@ -48,11 +54,26 @@ func NewRouter(config Config) (*Router, error) {
 	}, nil
 }
 
-func (a *Router) IngestEvents(w http.ResponseWriter, r *http.Request) {
-	a.config.IngestHandler.ServeHTTP(w, r)
+// FIXME: openapi generator doesn't generate an input param
```
hekike, line 57: @tothandras openapi generated doesn't add an input param here like for meter. Any idea why?

api/openapi.yaml
```diff
@@ -167,6 +176,28 @@ paths:
...
+            schema:
+              $ref: "#/components/schemas/Namespace"
+      responses:
+        "201":
+          description: Created
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Namespace"
+          $ref: "#/components/responses/NotFoundProblemResponse"
```
tothandras, line 198: Must be left here by accident. Please also include in case no manager is used:
```
        "501":
          $ref: "#/components/responses/NotImplementedProblemResponse"
```

api/openapi.yaml
```diff
@@ -167,6 +176,27 @@ paths:
           $ref: "#/components/responses/BadRequestProblemResponse"
         default:
           $ref: "#/components/responses/UnexpectedProblemResponse"
+  /api/v1alpha2/namespaces:
+    post:
+      description: Create namespace
+      operationId: createNamespace
+      tags:
+        - Meters
```
sagikazarmark, line 184: Should this belong to ingest or it's own category?

api/openapi.yaml
```diff
@@ -245,7 +275,7 @@ components:
           type: string
           minLength: 1
           maxLength: 63
-          pattern: "^[a-z0-9]+(?:[_-][a-z0-9]+)*$"
+          pattern: "^[a-z0-9]+(?:_[a-z0-9]+)*$"
```
sagikazarmark, line 278: What's the purpose of this change?

api/openapi.yaml
```diff
@@ -404,6 +447,15 @@ components:
       required: true
       schema:
         $ref: "#/components/schemas/IdOrSlug"
+    namespace:
+      name: OM-Namespace
+      description: Optional namespace
+      default: default
```
sagikazarmark, line 453: Default should be empty, shouldn't it?

sagikazarmark: BTW, we could eliminate the difference in how we handle namespaces and just use `default` as the default similarly to Kubernetes.

sagikazarmark: Scratch that, looks like you reached the same conclusion. 😄

sagikazarmark (commented):
Left a few comments. Other than that, LGTM.

sagikazarmark:
> I feel some disconnect between Kafka connector and Namespace Manager.

It doesn't really matter who implements the interface. The reason why I've put it in a separate component is to gain some clarity around what the Kafka connector does. IMO it's quite large at the moment.

For example: the meter init part will change as well for sure.

The good thing about Go is interface implementation is implicit, so you could pass anything to the namespace manager as a handler as long as it has the right methods.

I think we should we a bit before we start merging code back to that connector. For one, it's going to change for sure. Two, it's not actually a Kafka connector, but a KSQLDB one.

Architecturally, it's gonna settle down at some point and we will have a better picture of what belongs where.

Originally, I wanted to move the KSQLDB namespace handler to the namespace adapter package as well, but it would have been too much change, so it stayed in the kafka package.

api/openapi.yaml
hekike: Not allowing dash is both a devex decision to make it easier to copy slugs, also compatibility decision as it goes into table name.

api/openapi.yaml
tothandras: Both Ingest and Meters should be added.

PR #124, opened 2023-07-29
Add ClickHouse streaming processor

hekike:
Alternative dev-only streaming processor.
This processor doesn't deduplicate and doesn't ensure exactly once guarantee.

## How to test

1. Run docker compose as: `docker compose -f docker-compose-clickhouse.yaml up`

2. Add to `config.yaml`:

```yaml
processor:
  clickhouse:
    enabled: true
  ksqldb:
    enabled: false
sink:
  kafkaConnect:
    enabled: true
```

3. Run `air`

## BREAKING CHANGES

- Kafka payload `time` ISO string -> unix timestamp integer


docker-compose-clickhouse.yaml
```diff
@@ -0,0 +1,83 @@
...
+      KAFKA_BROKER_ID: 1
+      KAFKA_ZOOKEEPER_CONNECT: 'zookeeper:2181'
+      KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: PLAINTEXT:PLAINTEXT,PLAINTEXT_HOST:PLAINTEXT
+      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://broker:9092,PLAINTEXT_HOST://localhost:29092
+      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
+      KAFKA_GROUP_INITIAL_REBALANCE_DELAY_MS: 0
+      KAFKA_TRANSACTION_STATE_LOG_MIN_ISR: 1
+      KAFKA_TRANSACTION_STATE_LOG_REPLICATION_FACTOR: 1
+
+  clickhouse:
```
sagikazarmark, line 32: I'd move this back to the primary docker compose file and add `profiles: ["clickhouse"]` to this section.

That way you can optionally start it with docker compose up --profile clickhouse

Reference: https://docs.docker.com/compose/profiles/

go.mod
```diff
@@ -27,6 +27,18 @@ require (
...
+	github.com/andybalholm/brotli v1.0.5 // indirect
+	github.com/go-faster/city v1.0.1 // indirect
+	github.com/go-faster/errors v0.6.1 // indirect
+	github.com/klauspost/compress v1.16.3 // indirect
+	github.com/paulmach/orb v0.9.0 // indirect
+	github.com/pierrec/lz4/v4 v4.1.17 // indirect
+	github.com/segmentio/asm v1.2.0 // indirect
+)
+
+require (
```
sagikazarmark, line 40:
```suggestion
```

Merge these sections please.

internal/ingest/kafkaingest/serializer/serializer.go
```diff
@@ -17,12 +16,12 @@ type Serializer interface {
...
-	Type    string    `json:"type"`
-	Source  string    `json:"source"`
-	Subject string    `json:"subject"`
-	Time    time.Time `json:"time"`
-	Data    string    `json:"data"`
+	Id      string `json:"id"`
+	Type    string `json:"type"`
+	Source  string `json:"source"`
+	Subject string `json:"subject"`
+	Time    int64  `json:"time"`
```
sagikazarmark, line 23: There is one important consequence of this change: we lose timezone information. It's not necessarily a problem, unless there is a timezone change during a meter aggregation window in which case events may not belong to the right window.

It's as much as of an edge case as it can get, but it's good to know, maybe we should write it down somewhere.

internal/sink/kafka_connect.go
```diff
@@ -0,0 +1,57 @@
...
+type KafkaConnectConfig struct {
+	Address string
+}
+
+type Connector struct {
+	Name   string            `json:"name"`
+	Config map[string]string `json:"config"`
+}
+
+func NewKafkaConnect(config *KafkaConnectConfig) (*KafkaConnect, error) {
```
sagikazarmark, line 24: There is no need for either the config or the KafkaConnect struct to be pointers. Since we don't actually return any errors, maybe skip it (unless you know something I don't).

Also, since the config struct has a single member, maybe just pass that value to the constructor function.

internal/sink/kafka_connect.go
```diff
@@ -0,0 +1,57 @@
...
+	"fmt"
+	"net/http"
+)
+
+type KafkaConnect struct {
+	config *KafkaConnectConfig
+}
+
+type KafkaConnectConfig struct {
+	Address string
```
sagikazarmark, line 16: If this is a URL, maybe parse it as a URL in the original config which in turn can serve as a validation: if it can't parsed as a URL, it's invalid.

internal/sink/kafka_connect.go
```diff
@@ -0,0 +1,57 @@
...
+	Config map[string]string `json:"config"`
+}
+
+func NewKafkaConnect(config *KafkaConnectConfig) (*KafkaConnect, error) {
+	return &KafkaConnect{
+		config: config,
+	}, nil
+}
+
+func (k *KafkaConnect) CreateConnector(ctx context.Context, connector *Connector) error {
```
sagikazarmark, line 30: Connector shouldn't be a pointer.

internal/sink/kafka_connect.go
```diff
@@ -0,0 +1,57 @@
...
+	endpoint := fmt.Sprintf("%s/connectors", k.config.Address)
+
+	jsonData, err := json.Marshal(connector)
+	if err != nil {
+		return err
+	}
+
+	request, err := http.NewRequest("POST", endpoint, bytes.NewBuffer(jsonData))
```
sagikazarmark, line 38: Use NewRequestWithContext instead.

internal/sink/kafka_connect.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		return err
+	}
+	request = request.WithContext(ctx)
+	request.Header.Set("Content-Type", "application/json; charset=UTF-8")
+
+	client := &http.Client{}
```
sagikazarmark, line 45: If anything, just use the http.DefaultClient, but better yet accept an http client in the constructor and fall back to the default if none provided.

The default client does not specify any timeouts whatsoever, so in the future we certainly want to set a custom HTTP client.

internal/sink/kafka_connect.go
```diff
@@ -0,0 +1,57 @@
...
+	if err != nil {
+		return err
+	}
+	defer response.Body.Close()
+
+	if response.StatusCode == 201 || response.StatusCode == 409 {
+		return nil
+	}
+
+	return fmt.Errorf("unexpected status code at connector create: %d", response.StatusCode)
```
sagikazarmark, line 56: Nit: maybe parse the response so we know what went wrong.

internal/streaming/clickhouse_connector/connector.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+type ClickhouseConnectorConfig struct {
+	Logger       *slog.Logger
+	KafkaConnect *sink.KafkaConnect
+	ClickHouse   clickhouse.Conn
+	Database     string
+}
+
+func NewClickhouseConnector(config *ClickhouseConnectorConfig) (*ClickhouseConnector, error) {
```
sagikazarmark, line 30: Same thing about pointers.

internal/streaming/clickhouse_connector/connector.go
```diff
@@ -0,0 +1,201 @@
...
+	connector := &ClickhouseConnector{
+		config: config,
+	}
+
+	return connector, nil
+}
+
+func (c *ClickhouseConnector) Init(meter *models.Meter, namespace string) error {
+	// TODO: pass context to Init, also consider renaming it to CreateMeter
```
sagikazarmark, line 39: 👍

internal/streaming/clickhouse_connector/connector.go
```diff
@@ -0,0 +1,201 @@
...
+	ctx := context.Background()
```
sagikazarmark, line 40: Use context.TODO() so we know it needs to be changed (that's why it exists actually)

internal/streaming/clickhouse_connector/connector.go
```diff
@@ -0,0 +1,201 @@
...
+	if err != nil {
+		return fmt.Errorf("init: %w", err)
+	}
+
+	return nil
+}
+
+func (c *ClickhouseConnector) GetValues(meter *models.Meter, params *streaming.GetValuesParams, namespace string) ([]*models.MeterValue, error) {
+	// TODO: pass context to GetValues, also consider renaming it to QueryMeter
+	ctx := context.Background()
```
sagikazarmark, line 52: Same as above

main.go
```diff
@@ -164,6 +167,7 @@ func main() {
 	var group run.Group
 	var ingestCollector ingest.Collector
 	var streamingConnector streaming.Connector
+	namespaceHandlers := make([]namespace.Handler, 0)
```
sagikazarmark, line 170:
```suggestion
	var namespaceHandlers []namespace.Handler
```

append will initialize the slice

main.go
```diff
@@ -178,17 +182,33 @@ func main() {
...
 
 	// Initialize ksqlDB Streaming Processor
-	streamingConnector, ksqlDBNamespaceHandler, err := initKsqlDBStreaming(config, logger, eventSerializer, healthChecker)
-	if err != nil {
-		slog.Error("failed to initialize ksqldb streaming processor", "error", err)
-		os.Exit(1)
+	if config.Processor.KSQLDB.Enabled {
+		ksqlDBStreamingConnector, ksqlDBNamespaceHandler, err := initKsqlDBStreaming(config, logger, eventSerializer, healthChecker)
+		if err != nil {
+			slog.Error("failed to initialize ksqldb streaming processor", "error", err)
```
sagikazarmark, line 192:
```suggestion
			logger.Error("failed to initialize ksqldb streaming processor", "error", err)
```

internal/streaming/clickhouse_connector/connector.go
```diff
@@ -0,0 +1,201 @@
...
+			"hostname":                       "clickhouse",
+			"port":                           "8123",
+			"ssl":                            "false",
+			"username":                       "default",
+			"password":                       "",
+			"key.converter":                  "org.apache.kafka.connect.storage.StringConverter",
+			"value.converter":                "org.apache.kafka.connect.json.JsonConverter",
+			"value.converter.schemas.enable": "false",
+			"schemas.enable":                 "false",
+			"topics":                         getEventsTableName(namespace),
```
hekike, line 183: TODO: check if wildcard works to match all events topic instead of connector per namespaces

internal/streaming/clickhouse_connector/connector.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+func (c *ClickhouseConnector) createSinkConnector(ctx context.Context, namespace string) error {
+	connector := &sink.Connector{
+		Name: "clickhouse",
+		Config: map[string]string{
+			"connector.class":                "com.clickhouse.kafka.connect.ClickHouseSinkConnector",
+			"database":                       "default",
+			"errors.retry.timeout":           "30",
+			"hostname":                       "clickhouse",
```
hekike, line 174: TODO: pass in ClickHouse hostname, port, username, password

go.mod
hekike: Go mod tidy cuts it into two least

internal/sink/kafka_connect.go
hekike: How do I parse it as a URL?

internal/streaming/clickhouse_connector/connector.go
hekike: it does

internal/streaming/clickhouse_connector/connector.go
```diff
@@ -0,0 +1,211 @@
...
+			"hostname":                       c.config.SinkConfig.Hostname,
+			"port":                           fmt.Sprint(c.config.SinkConfig.Port),
+			"ssl":                            fmt.Sprint(c.config.SinkConfig.SSL),
+			"username":                       c.config.SinkConfig.Username,
+			"password":                       c.config.SinkConfig.Password,
+			"key.converter":                  "org.apache.kafka.connect.storage.StringConverter",
+			"value.converter":                "org.apache.kafka.connect.json.JsonConverter",
+			"value.converter.schemas.enable": "false",
+			"schemas.enable":                 "false",
+			"topics.regex":                   "om_.+_events",
```
sagikazarmark, line 193: Shouldn't the namespace be added here? It's not used anywhere right now.

sagikazarmark (approved):

sagikazarmark (approved):

PR #126, opened 2023-07-31
feat(dedupe): add redis dedupe

hekike:
Dedupe in API via Redis.

Deduping in the API can't guarantee consistency due to the parallel ingestion of events.
Currently, the only consistent deduplication is provided by the ksqlDB Processor.

### DIFF
--- a/config.go
+++ b/config.go
@@ -36,7 +36,8 @@ type configuration struct {
 
 	// Dedupe configuration
 	Dedupe struct {
-		Redis dedupeRedisConfiguration
+		Redis  dedupeRedisConfiguration
+		Memory dedupeMemoryConfiguration
 	}
 
 	Namespace struct {
@@ -298,6 +299,24 @@ func (c dedupeRedisConfiguration) Validate() error {
 	return nil
 }
 
+// Dedupe memory configuration
+type dedupeMemoryConfiguration struct {
+	Enabled bool
+	Size    int
+}
+
+func (c dedupeMemoryConfiguration) Validate() error {
+	if !c.Enabled {
+		return nil
+	}
+
+	if c.Size == 0 {
+		return errors.New("dedupe memory size is required")
+	}
+
+	return nil
+}
+
 type logConfiguration struct {
 	// Format specifies the output log format.
 	// Accepted values are: json, text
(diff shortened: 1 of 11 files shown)


go.mod
```diff
@@ -29,6 +29,8 @@ require (
 	golang.org/x/exp v0.0.0-20230522175609-2e198f4a06a1
 )
 
+require github.com/dgryski/go-rendezvous v0.0.0-20200823014737-9f7001d12a5f // indirect
+
 require (
```
sagikazarmark, line 34: Merge these sections please then run `go mod tidy`

internal/dedupe/redis_dedupe/redis_dedupe.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		return isUnique, err
+	}
+
+	return isUnique, nil
+}
+
+// CloudEvents are unique based on the source and id
+func getKey(ev event.Event) string {
+	return fmt.Sprintf("%s-%s", ev.Source(), ev.ID())
```
sagikazarmark, line 54: This is problematic in the sense that ID's need to be globally unique, but there is nothing to enforce that at the moment. An ID submitted by one tenant would transparently be deduplicated. We need to add some tenancy information in here (namespace?) to make sure deduplication only happens below the tenant level.

internal/dedupe/dedupe.go
```diff
@@ -0,0 +1,26 @@
+package dedupe
+
+import (
+	"context"
+
+	"github.com/cloudevents/sdk-go/v2/event"
+)
+
+type Connector interface {
```
sagikazarmark, line 9: Maybe call it Deduplicator or something like that.

internal/ingest/httpingest/httpingest.go
```diff
@@ -51,6 +52,24 @@ func (h Handler) ServeHTTP(w http.ResponseWriter, r *http.Request, params api.In
...
+			models.NewStatusProblem(r.Context(), err, http.StatusInternalServerError).Respond(w, r)
+			return
+		}
+		if !isUnique {
+			logger.DebugCtx(r.Context(), "event is not unique")
+			w.WriteHeader(http.StatusOK)
+			return
+		}
+		logger.DebugCtx(r.Context(), "event is unique")
+	}
```
sagikazarmark, line 71: I'm thinking this should be a collector implementation in the ingest package and we should wrap the "real" collector as a sort of delegate or middleware. But the HTTP layer is definitely not where it should be.

config.go
```diff
@@ -33,6 +33,11 @@ type configuration struct {
 		Kafka ingestKafkaConfiguration
 	}
 
+	// Dedupe configuration
+	Dedupe struct {
```
sagikazarmark, line 37: We could provide a simple in-memory implementation for local development and testing. Then the enable flag should be moved the dedupe level.

I'm also wondering if this should be within ingest, given that it's happening during ingest.

hekike: Added

sagikazarmark (approved):
LGTM

PR #128, opened 2023-07-31
feat(namespace): configure default and management

hekike:
- Configurable default namespace
- Configurable namespace management


sagikazarmark (approved):
LGTM

PR #129, opened 2023-07-31
feat(meter): create and delete via api

hekike:
- Create meter via API
- Delete meter via API
- Rename `collector.Init` to `collector.CreateMeter`
- Rename `collector.GetValues` to `collector.QueryMeter`
- Add `collector.DeleteMeter`
- Add context to collector methods and write them up


sagikazarmark (approved):
LGTM

PR #151, opened 2023-08-08
refactor(processor): clickhouse sql builder

hekike:
Use SQL builder and avoid injection.


internal/streaming/clickhouse_connector/query.go
```diff
@@ -2,56 +2,224 @@ package clickhouse_connector
...
+	sb.IfNotExists()
+	sb.Define("id", "String")
+	sb.Define("type", "LowCardinality(String)")
+	sb.Define("subject", "String")
+	sb.Define("source", "String")
+	sb.Define("time", "DateTime")
+	sb.Define("data", "String")
+	sb.SQL("ENGINE = MergeTree")
+	sb.SQL("PARTITION BY toYYYYMM(time)")
+	sb.SQL("ORDER BY (time, type, subject)")
```
tothandras, line 40: Not sure about this, but could it be:
```
ORDER BY (type, time, subject, source, id)
PRIMARY KEY (type, time, subject)
```

hekike: Will change in separate PR

tothandras (approved):

PR #156, opened 2023-08-14
Get rid of unnecessary namespace model

sagikazarmark:
## Overview

This is an example of what I'd like to do with all models eventually. It's going to make third party code generation easier and won't require importing openmeter in order to use it's OpenAPI descriptor.


api/openapi.yaml
```diff
@@ -440,10 +436,6 @@ components:
           maxLength: 63
     Namespace:
       type: object
-      x-go-name: Namespace
```
tothandras, line 443: I'm not sure about this. Go client generators could easily ignore the `x-go-*` extended properties.

tothandras (changes requested):

api/openapi.yaml
sagikazarmark: Well, it's currently a problem in cloud. I don't think we need these mappings. We can easily map the few structs where it matters manually.

PR #173, opened 2023-08-15
Refactor dedupe configuration

sagikazarmark:
## Overview

Refactor deduplication config to avoid invalid scenarios (eg. both redis and memory are enabled).

Polymorphic configuration is easier to understand:

```yaml
dedupe:
  enabled: true
  driver: redis|memory
  config:
    address: ...
    size: ...
```

**This is a breaking change! A new release will require configuration updates as well**

### DIFF
--- a/config.go
+++ b/config.go
@@ -10,12 +10,17 @@ import (
 
 	"github.com/confluentinc/confluent-kafka-go/v2/kafka"
 	"github.com/lmittmann/tint"
+	"github.com/mitchellh/mapstructure"
+	"github.com/redis/go-redis/v9"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 	"github.com/thmeitz/ksqldb-go/net"
 	"golang.org/x/exp/slices"
 	"golang.org/x/exp/slog"
 
+	"github.com/openmeterio/openmeter/internal/dedupe/memorydedupe"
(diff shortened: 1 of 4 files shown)


config.go
```diff
@@ -465,17 +580,18 @@ func configure(v *viper.Viper, flags *pflag.FlagSet) {
...
+	v.SetDefault("dedupe.enabled", false)
+	v.SetDefault("dedupe.driver", "memory")
 
 	// Dedupe Memory configuration
-	v.SetDefault("dedupe.memory", false)
-	v.SetDefault("dedupe.memory.size", 128)
+	v.SetDefault("dedupe.config.size", 128)
+
+	// Dedupe Redis configuration
+	v.SetDefault("dedupe.config.address", "127.0.0.1:6379")
```
hekike, line 590: Any reason why `dedupe.config` namespace instead of `dedupe.redis`?

sagikazarmark: That is to use the same config structure regardless of the driver:

```yaml
dedupe:
  enabled: true
  driver: redis
  config:
    address: ...

dedupe:
  enabled: true
  driver: memory
  config:
    size: ...

# VS

dedupe:
  enabled: true
  driver: redis
  redis: # This feels redundant
    address: ...
```

hekike (approved):

PR #174, opened 2023-08-16
Basic otel trace support

sagikazarmark:
## Overview

This PR adds basic otel trace support for incoming requests. Integration for downstream requests (kafka, clickhouse) will be added in subsequent PRs.

### DIFF
--- a/main.go
+++ b/main.go
@@ -154,7 +154,7 @@ func main() {
 	traceProvider := sdktrace.NewTracerProvider(traceProviderOptions...)
 	defer func() {
 		if err := traceProvider.Shutdown(ctx); err != nil {
-			logger.Error("shutting down trace provider: %v", err)
+			logger.Error("shutting down trace provider", "error", err)
 		}
 	}()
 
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -132,6 +137,31 @@ func main() {
...
+			os.Exit(1)
+		}
+
+		traceProviderOptions = append(traceProviderOptions, sdktrace.WithBatcher(exporter))
+	}
+
+	traceProvider := sdktrace.NewTracerProvider(traceProviderOptions...)
+	defer func() {
+		if err := traceProvider.Shutdown(ctx); err != nil {
+			logger.Error("shutting down trace provider: %v", err)
```
tothandras, line 158:
```suggestion
			logger.Error("shutting down trace provider", "error", err)
```

tothandras (changes requested):
Should we make it optional? Seems like the telemetry address is required.

sagikazarmark:
Telemetry address was required even before this change, but has a default value. We can think about making it optional later, but it doesn't cost anything.

tothandras:
I see, I thought it was a push address.

tothandras (approved):

PR #186, opened 2023-08-20
ci: remove next from build targets

sagikazarmark:
## Overview

We no longer use next


sagikazarmark:
Next branch is removed, so merging.

PR #218, opened 2023-08-30
feat!: drop ksqldb support

sagikazarmark:
## Overview

We decided to go with Clickhouse as a storage layer.


quickstart/config.yaml
```diff
@@ -3,8 +3,6 @@ ingest:
     broker: kafka:29092
 
 processor:
-  ksqldb:
-    enabled: false
   clickhouse:
     enabled: true
```
hekike, line 7: `enabled` can be removed

hekike (approved):

tothandras (approved):

Issue #213, opened 2023-08-29
Create new meter query endpoint


sagikazarmark:
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/openmeterio/openmeter/issues) for an issue that matches the one I want to file, without success.

### Problem Description

The current /values endpoint has a few shortcomings that makes it less useful (at least in the UI):

- Grouping results by default window size when no `windowSize` parameter is present
- `groupBy` parameter does not accept multiple values (it does accept comma separated values)

### Proposed Solution

Implement a new /query endpoint:

- Do not group results by default window size when no `windowSize` is present
- Accept array values for `groupBy` and `subject`
- Return `from` and `to` parameters (and make both of them optional)

The new endpoint will be able to return values aggregated for a given (open ended) interval:

- without any window size, subject or groupBy grouping
- groupped by window size (but not subject or groupBy fields)
- groupped by window size and subject (but not groupBy fields)
- groupped by window size subject and any/all group by fields

The /query endpoint will deprecate the /values endpoint once we have adopted it.

### Additional Information

What's the default window size for if we are not going to use it?

PR #221, opened 2023-08-30
Meter query changes for the new query endpoint

sagikazarmark:
## Overview

This PR introduces changes to the Clickhouse query implementation necessary for the new query endpoint outlined in #213

The changes _SHOULD BE_ backwards compatible, hence they are in a separate PR for easier review.

Related #213


internal/streaming/clickhouse_connector/query.go
```diff
@@ -221,21 +242,36 @@ func (d queryMeterView) toSQL() (string, []interface{}, error) {
...
+		}
+
+		where = append(where, queryView.Or(slicesx.Map(d.Subject, mapFunc)...))
 	}
+
 	if d.From != nil {
 		where = append(where, queryView.GreaterEqualThan("windowstart", d.From.Unix()))
 	}
+
+	windowEndColumn := "windowstart"
```
hekike, line 257: It's not clear to me why we need this. Can you please comment code.

sagikazarmark: Discussed it offline, this was a misunderstanding on my part. We can use windowend here.

PR #223, opened 2023-08-31
Add seeder

sagikazarmark:
## Overview

Add a tool to seed OpenMeter


hekike:
We use https://k6.io for load testing. Why not to use that one for seeding as well?

sagikazarmark:
This looked easier. Also, you can switch between continuously seeding and just sending a number of samples to the API. k6 would be more limited in this regard IMO.

PR #238, opened 2023-09-07
Kafka connect refactor

### LINKED ISSUE #230: Refactor Kafka Connect configuration
### Preflight Checklist

- [X] I have searched the [issue tracker](https://www.github.com/openmeterio/openmeter/issues) for an issue that matches the one I want to file, without success.

### Problem Description

#229 refactors Kafka Connect, but I believe there is more room for improvement.

The configuration feels a bit "disorganized" at the moment.

I'd also note that we probably do not want to use this feature in production, but want to manually configure Kafka Connect (I may be wrong about that).

### Proposed Solution

I can imagine two potential routes, both involve supporting multiple connectors.

The advantage of supporting multiple connectors is that we can experiment with different configurations without breaking existing ones.

Option one is to allow passing raw connector configuration:

```yaml
sink:
  kafkaConnect:
    enabled: true
    connectors:
      - name: clickhouse
        config:
          connector.class: "com.clickhouse.kafka.connect.ClickHouseSinkConnector"
          database: "openmeter"

          # ...
```

The downside of this solution is that we have no way to validate the configuration.

An alternative solution would be adding config schema for supported connectors (currently only clickhouse anyway):

```yaml
sink:
  kafkaConnect:
    enabled: true
    connectors:
      - name: clickhouse
        type: clickhouse
        config:
          database: "openmeter"

          # ...
```

Although option two is a bit more work, I probably prefer that for the moment because it makes misconfiguration harder.

### Alternatives Considered

_No response_

### Additional Information

_No response_

sagikazarmark:
## Overview

Fixes #230

## Notes for reviewer

There is one major downside to this PR: we lose connector defaults, but I believe it's something we can live with.

I will improve the dev configuration in a subsequent PR.

### DIFF
--- a/quickstart/config.yaml
+++ b/quickstart/config.yaml
@@ -10,8 +10,20 @@ sink:
   kafkaConnect:
     enabled: true
     url: http://kafka-connect:8083
-    clickhouse:
-      hostname: clickhouse
+    connectors:
+      - name: clickhouse
+        type: clickhouse
+        config:
+          hostname: clickhouse
+          port: 8123
+          ssl: false
+          database: openmeter
+          username: default
+          password: default
+          deadLetterQueue:
+            topicName: om_deadletterqueue
+            replicationFactor: 1
+            contextHeaders: true
 
 meters:
   - slug: m1
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -81,18 +81,23 @@ func TestComplete(t *testing.T) {
 			KafkaConnect: KafkaConnectSinkConfiguration{
 				Enabled: true,
 				URL:     "http://127.0.0.1:8083",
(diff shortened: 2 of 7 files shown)


quickstart/config.yaml
```diff
@@ -10,8 +10,20 @@ sink:
   kafkaConnect:
     enabled: true
     url: http://kafka-connect:8083
-    clickhouse:
-      hostname: clickhouse
+    connectors:
```
hekike, line 13: I think we should add it to `config.example.yaml` too

sagikazarmark: I wanna change that config a bit anyway, so I'll do that in a followup PR.

hekike (approved):

PR #243, opened 2023-09-08
chore(deploy): update helm chart

tothandras:
Update helm chart to use ClickHouse and latest image.

TODO:
- [ ] ability to pass in credentials for existing clickhouse and kafka
- [ ] expose more configurations via values


hekike:
Tested with quickstart commands, works great.

hekike (approved):

PR #244, opened 2023-09-08
feat(api): always include windowStart and windowEnd in query values

tothandras:
## Overview

When window size is not specified in the query params, get the windowstart and windowend as the min and max of the timestamps.

`GET /api/v1/meters/m1/query`

before:

```json
{
  "data": [
    {
      "value": 3,
      "windowStart": null,
      "windowEnd": null,
      "subject": null
    }
  ]
}
```

after:

```json
{
  "data": [
    {
      "value": 3,
      "windowStart": "2023-01-01T00:00:00Z",
      "windowEnd": "2023-01-02T00:01:00Z"
    }
  ]
}
```


pkg/models/meter.go
```diff
@@ -359,9 +359,9 @@ func AggregateMeterValues(values []*MeterValue, aggregation MeterAggregation, wi
 // MeterQueryRow returns a single row from the meter dataset.
 type MeterQueryRow struct {
 	Value       float64           `json:"value"`
-	WindowStart *time.Time        `json:"windowStart"`
-	WindowEnd   *time.Time        `json:"windowEnd"`
-	Subject     *string           `json:"subject"`
+	WindowStart time.Time         `json:"windowStart"`
+	WindowEnd   time.Time         `json:"windowEnd"`
+	Subject     string            `json:"subject,omitempty"`
```
sagikazarmark, line 364: Shouldn't this stay nil?

sagikazarmark (changes requested):

internal/server/router/router.go
```diff
@@ -323,31 +323,18 @@ func (a *Router) QueryMeter(w http.ResponseWriter, r *http.Request, meterIDOrSlu
...
 		To:         params.To,
 		Data: slicesx.Map(values, func(val *models.MeterValue) models.MeterQueryRow {
-			row := models.MeterQueryRow{
-				Value:   val.Value,
-				GroupBy: val.GroupBy,
-			}
-
-			if val.Subject != "" {
-				row.Subject = &val.Subject
-			}
```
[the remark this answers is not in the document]
tothandras: nil vs empty string with omitempty

internal/streaming/clickhouse_connector/query.go
```diff
@@ -199,10 +199,10 @@ func (d queryMeterView) toSQL() (string, []interface{}, error) {
 		default:
 			return "", nil, fmt.Errorf("invalid window size type: %s", *d.WindowSize)
 		}
-	}
 
-	if groupByWindowSize {
 		groupByColumns = append(groupByColumns, "windowstart", "windowend")
+	} else {
+		selectColumns = append(selectColumns, "min(windowstart)", "max(windowend)")
```
hekike, line 205: Shouldn't we alias them? I know we reverse-map them by position but we alias other fields

hekike: Also, do we need to do this in SQL? This `min` and `max` are only used to return the smallest window between the already returned rows. As the returned data is ordered it's basically `rows[0].windowStart` and `rows[len(rows) - 1).windowEnd`, no? Also in many cases when we don't aggregate it's equal with raw windows.

api/openapi.yaml
```diff
@@ -280,6 +280,8 @@ paths:
                     items:
                       $ref: "#/components/schemas/MeterQueryRow"
                 required:
+                  - from
+                  - to
```
hekike, line 284: I think it's worth considering `to` to be nullable. For example when someone just wants the current month. Or would we expect users to use future dates in this case?

internal/streaming/clickhouse_connector/query.go
hekike: Maybe I misunderstood this. This is not the top level windowStart, it's the per value.

api/openapi.yaml
sagikazarmark: I think that's fair, yes. I believe grafana also translates special date ranges to actual dates on the ui:

Screenshot 2023-09-10 at 13 40 22

internal/streaming/clickhouse_connector/query.go
sagikazarmark: This piece of code could probably use some refactoring with all the different conditional logic and changes lately.

But the min, max aggregations are only added when there is no grouping by window size which means we get one row per any other group by (or a full total when no grouping is applied).

sagikazarmark (approved):
LGTM

One comment though: I'm not sure if this change breaks the values endpoint.

We should deprecate and drop that endpoint as soon as we can.

internal/streaming/clickhouse_connector/query.go
tothandras: > Shouldn't we alias them? I know we reverse-map them by position in Go but we alias other fields

Unfortunately, not. For that to work, we'd need to give an alias to the table. I'm not sure which one is nicer at the end.

```
SELECT min(windowstart) AS windowstart, max(windowend), sumMerge(value) FROM "om_default_m1"
WHERE windowstart >= '2023-07-31 22:02:00';

Code: 184. DB::Exception: Aggregate function min(windowstart) is found in WHERE in query: While processing min(windowstart) AS windowstart. (ILLEGAL_AGGREGATION) (version 23.8.1.41512 (official build))
```

tothandras:
@sagikazarmark review pls, rebased on main

sagikazarmark (approved):
LGTM

PR #245, opened 2023-09-10
chore: update example config

sagikazarmark:
## Overview

Add kafka connect config to dev example

### DIFF
--- a/config.example.yaml
+++ b/config.example.yaml
@@ -1,11 +1,35 @@
 address: 127.0.0.1:8888
 
-log:
-  format: json
-  level: info
-
 telemetry:
-  address: 127.0.0.1:10000
+  log:
+    level: debug
+
+sink:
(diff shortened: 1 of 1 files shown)


config.example.yaml
```diff
@@ -1,11 +1,35 @@
...
+          ssl: false
+          database: openmeter
+          username: default
+          password: default
+          deadLetterQueue:
+            topicName: om_deadletterqueue
+            replicationFactor: 1
+            contextHeaders: true
+
+# dedupe:
```
hekike, line 26: why commented out?

tothandras: use the memory dedup as default?

sagikazarmark: yep

It's there so we can easily enable it locally. But locally, memory dedupe is the default.

PR #260, opened 2023-09-12
feat(meter): query with accept csv support

hekike:
Accept CSV:

`curl --header "Accept: text/csv" http://localhost:8888/api/v1/meters/m1/query\?windowSize\=HOUR\&subject\=customer-1\&groupBy\=method\&groupBy\=path`

```
window_start,window_end,subject,method,path,value
2023-01-01T00:00:00Z,2023-01-01T01:00:00Z,customer-1,GET,/hello,2.000000
2023-01-02T00:00:00Z,2023-01-02T01:00:00Z,customer-1,GET,/hello,1.000000
```


internal/server/router/router.go
```diff
@@ -408,6 +412,35 @@ func (resp QueryMeterResponse) Render(_ http.ResponseWriter, _ *http.Request) er
...
+		}
+		data = append(data, fmt.Sprintf("%f", row.Value))
+		out = append(out, strings.Join(data, ","))
+	}
+
+	csvContent := strings.Join(out, "\n")
+
+	w.Header().Set("Content-Type", "text/csv")
+	w.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=%s.csv", meterIDOrSlug))
+	w.Write([]byte(csvContent))
```
sagikazarmark, line 441: How about using encoding/csv instead?

https://pkg.go.dev/encoding/csv

internal/server/router/router.go
```diff
@@ -374,7 +374,11 @@ func (a *Router) QueryMeter(w http.ResponseWriter, r *http.Request, meterIDOrSlu
 		}),
 	}
 
-	_ = render.Render(w, r, resp)
+	if r.Header.Get("Accept") == "text/csv" {
```
sagikazarmark, line 377: Parsing media type would be safer: https://pkg.go.dev/mime#ParseMediaType

sagikazarmark: Or use a snippet from the render library: https://github.com/go-chi/render/blob/v1.0.3/content_type.go#L67

(We should probably contribute CSV support to that library)

sagikazarmark (changes requested):
I would probably add a test to see how everything behaves when there is no subject in the response (subject is null).

tothandras (approved):

internal/server/router/router.go
hekike: Done

PR #262, opened 2023-09-13
fix(processor): clickhouse materialized view type shadow

hekike:
Use absolute path to filter on events table's `type` at materialized view creation to avoid shadowing if the MV has a `type` column due to a group by.


sagikazarmark:
This is not going to change existing tables. Do we have a plan for supporting these kinds of changes with migrations?

Or until 1.0.0 stable this is considered fair play?

hekike:
It's fine if old tables have the relative WHERE. It only impacts tables where `type` shadows.

sagikazarmark (approved):

PR #283, opened 2023-09-29
feat(api): add vercel ai helper to node sdk

hekike:
I'm still debating if this should be in SDK or just an example.
I can see an argument for both. What do you think?

### DIFF
--- a/api/client/node/next.ts
+++ b/api/client/node/next.ts
@@ -1,0 +1,77 @@
+import type { OpenAIStreamCallbacks } from 'ai'
+import type { TiktokenModel } from 'js-tiktoken'
+
+let encodingForModel: (model: TiktokenModel) => any | undefined
+
+type OpenAIUsage = {
+  total_tokens: number
+  prompt_tokens: number
+  completion_tokens: number
+}
+
+type OpenAIStreamCallbacksWithUsage = OpenAIStreamCallbacks & {
+  onUsage?: (usage: OpenAIUsage) => void
+}
+
+export function createOpenAIStreamCallback(
+  {
+    model,
+    prompts,
+  }: {
+    model: TiktokenModel
+    prompts: string[]
+  },
+  openAIStreamCallbacks: OpenAIStreamCallbacksWithUsage
+) {
+  // Tiktoken is an optional dependency, so we require it conditionally
+  if (!encodingForModel) {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const { encodingForModel: encodingForModel_ } = require('js-tiktoken')
+    encodingForModel = encodingForModel_
+  }
+
+  const enc = encodingForModel(model)
+  let promptTokens = 0
+  let completionTokens = 0
+
+  const streamCallbacks: OpenAIStreamCallbacks = {
+    ...openAIStreamCallbacks,
+
+    onStart() {
+      for (const content of prompts) {
+        const tokens = enc.encode(content)
+        promptTokens += tokens.length
+      }
+
+      if (typeof openAIStreamCallbacks?.onStart === 'function') {
+        return openAIStreamCallbacks.onStart()
+      }
+    },
+    onToken: (content) => {
+      // To test tokenizaton see: https://platform.openai.com/tokenizer
+      const tokens = enc.encode(content)
+      completionTokens += tokens.length
+
+      if (typeof openAIStreamCallbacks?.onToken === 'function') {
+        return openAIStreamCallbacks.onToken(content)
+      }
+    },
+    onFinal(completion: string) {
+      // Mimicking OpenAI usage metadata API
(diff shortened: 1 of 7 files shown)


api/client/node/next.ts
```diff
@@ -0,0 +1,77 @@
...
+  }: {
+    model: TiktokenModel
+    prompts: string[]
+  },
+  openAIStreamCallbacks: OpenAIStreamCallbacksWithUsage
+) {
+  // Tiktoken is an optional dependency, so we require it conditionally
+  if (!encodingForModel) {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const { encodingForModel: encodingForModel_ } = require('js-tiktoken')
```
tothandras, line 29: `import` and `require` shouldn't be mixed

api/client/node/next.ts
```diff
@@ -0,0 +1,77 @@
...
+      for (const content of prompts) {
+        const tokens = enc.encode(content)
+        promptTokens += tokens.length
+      }
+
+      if (typeof openAIStreamCallbacks?.onStart === 'function') {
+        return openAIStreamCallbacks.onStart()
+      }
+    },
+    onToken: (content) => {
```
tothandras, line 50: To keep it consistent
```suggestion
    onToken (content) {
```

api/client/node/test/next.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+              done()
+            },
+          }
+        )
+
+        streamCallbacks.onStart && streamCallbacks.onStart()
+        streamCallbacks.onToken && streamCallbacks.onToken('Hello ')
+        streamCallbacks.onToken && streamCallbacks.onToken(' World!')
+        streamCallbacks.onFinal && streamCallbacks.onFinal('Hello World!')
```
tothandras, line 33:
```suggestion
        streamCallbacks.onStart?.()
        streamCallbacks.onToken?.('Hello ')
        streamCallbacks.onToken?.(' World!')
        streamCallbacks.onFinal?.('Hello World!')
```

api/client/node/test/next.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    it('should return usage', () =>
+      new Promise<void>((done) => {
+        const streamCallbacks = createOpenAIStreamCallback(
+          {
+            model,
+            prompts,
+          },
+          {
+            onUsage: (usage) => {
+              expect(usage).toEqual({
```
tothandras, line 19: You might need to wrap this into a try/catch and reject the promise?

api/client/node/test/next.test.ts
```diff
@@ -0,0 +1,75 @@
...
+            },
+          }
+        )
+
+        streamCallbacks.onStart && streamCallbacks.onStart()
+        streamCallbacks.onToken && streamCallbacks.onToken('Hello ')
+        streamCallbacks.onToken && streamCallbacks.onToken(' World!')
+        streamCallbacks.onFinal && streamCallbacks.onFinal('Hello World!')
+
+        expect(true).toBe(true)
```
tothandras, line 35: Why do you need this? :)

api/client/node/test/next.test.ts
```diff
@@ -0,0 +1,75 @@
...
+          },
+          callbacks
+        )
+
+        streamCallbacks.onStart && streamCallbacks.onStart()
+        streamCallbacks.onToken && streamCallbacks.onToken('Hello ')
+        streamCallbacks.onToken && streamCallbacks.onToken(' World!')
+        streamCallbacks.onCompletion &&
+          streamCallbacks.onCompletion('Hello World!')
+        streamCallbacks.onFinal && streamCallbacks.onFinal('Hello World!')
```
tothandras, line 70:
```suggestion
        streamCallbacks.onStart?.()
        streamCallbacks.onToken?.('Hello ')
        streamCallbacks.onToken?.(' World!')
        streamCallbacks.onCompletion?.('Hello World!')
        streamCallbacks.onFinal?.('Hello World!')
```

api/client/node/test/next.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+        expect(true).toBe(true)
```
tothandras, line 72: Same

tothandras (changes requested):

api/client/node/test/next.test.ts
hekike: It exists this way too

api/client/node/next.ts
hekike: ES6 imports are async by default. You can't do synchronous `import`, only with `require`. As this factory is synchronous, I don't see a way to use `import`. Btw I don't know any reason why mixing would impact a Node.js application.

PR #320, opened 2023-10-16
Kafka Consumer Worker

hekike:
## What?

- Kafka Consumer as Go Worker instead of Kafka Connect
- Separate server and worker commands
- Horizontally scalable consumers via Kafka's consumer rebalancing
- Validating events against meters for better deadlettering

## Why?

Motivation in order of importance:

- Increase deduplication consistency by deduping in consumer
- Reduce error blast radius when Redis is down by deduping after ingestion
- Minimize the number of correct events sent to dead-letter when a single event is malformed in the batch through validating against meters
- Enable autoscaling via Kafka consumer rebalancing
- Being able to implement more advanced dead-letter logic, like sinking it to a separate clickhouse table per namespace

Tradeoff:

- We are reimplementing some of Kafka Connect

## Breaking Change

Operational breaking change. Replacing Kafka Connect with the sink-worker process.
There is no breaking change for the APIs.


config/ingest.go
```diff
@@ -40,6 +41,12 @@ func (c KafkaIngestConfiguration) CreateKafkaConfig() kafka.ConfigMap {
 		"go.logs.channel.enable": true,
 	}
 
+	// This is needed when using localhost brokers on OSX,
+	// since the OSX resolver will return the IPv6 addresses first.
+	if strings.Contains(c.Broker, "localhost") || strings.Contains(c.Broker, "127.0.0.1") {
+		config["broker.address.family"] = "v4"
+	}
```
sagikazarmark, line 48: This is okay temporarily, but I can imagine a scenario where users want ipv6 resolution for localhost.

I'd suggest making this setting available in config for the long-term.

Removing localhost and using the loopback IP address would also be okay.

Dockerfile
```diff
@@ -27,7 +27,7 @@ ARG VERSION
 
 # See https://github.com/confluentinc/confluent-kafka-go#librdkafka
 # See https://github.com/confluentinc/confluent-kafka-go#static-builds-on-linux
-RUN go build -ldflags "-linkmode external -extldflags \"-static\" -X main.version=${VERSION}" -tags musl -o /usr/local/bin/openmeter .
+RUN go build -ldflags "-linkmode external -extldflags \"-static\" -X main.version=${VERSION}" -tags musl -o /usr/local/bin/openmeter ./cmd/server
```
sagikazarmark, line 30: I don't think we need two dockerfiles. One should be enough. It make distribution easier.

cmd/server/main.go
```diff
@@ -204,20 +193,6 @@ func main() {
...
-		if err != nil {
-			logger.Error("failed to initialize deduplicator", "error", err)
-			os.Exit(1)
-		}
-
-		ingestCollector = ingest.DeduplicatingCollector{
-			Collector:    ingestCollector,
-			Deduplicator: deduplicator,
-		}
-	}
```
sagikazarmark, line 219: Can you please explain the changes around deduplication? (Keep in mind that we may have to hold off on merging this or tagging a new openmeter, until we roll out the proxy removal in cloud to prod AND also need to make these changes in cloud)

cmd/sink-worker/main.go
```diff
@@ -0,0 +1,258 @@
...
+	err = conf.Validate()
+	if err != nil {
+		panic(err)
+	}
+
+	extraResources, _ := resource.New(
+		context.Background(),
+		resource.WithContainer(),
+		resource.WithAttributes(
+			semconv.ServiceName("openmeter"),
```
sagikazarmark, line 77: Service name should be unique, so it should be something like `openmeter-sink-worker`.

cmd/sink-worker/main.go
```diff
@@ -0,0 +1,258 @@
...
+		Storage:             storage,
+		Deduplicator:        deduplicator,
+		ConsumerKafkaConfig: consumerKafkaConfig,
+		ProducerKafkaConfig: producerKafkaConfig,
+		MinCommitCount:      config.Sink.MinCommitCount,
+		MaxCommitWait:       config.Sink.MaxCommitWait,
+		NamespaceRefetch:    config.Sink.NamespaceRefetch,
+	}
+
+	return sink.NewSink(&sinkConfig)
```
sagikazarmark, line 257: No need for a pointer unless you modify it...which you don't.

cmd/sink-worker/version.go
```diff
@@ -0,0 +1,36 @@
+package main
+
+// TODO: consolidate between ../version.go
```
sagikazarmark, line 3: Copying this 30 LOC is probably easier, so I would leave it that way.

config/sink.go
```diff
@@ -2,222 +2,36 @@ package config
...
 
-	"github.com/mitchellh/mapstructure"
 	"github.com/spf13/viper"
 )
 
 type SinkConfiguration struct {
-	KafkaConnect KafkaConnectSinkConfiguration
+	MinCommitCount   int
+	MaxCommitWait    time.Duration
+	NamespaceRefetch time.Duration
```
sagikazarmark, line 13: Can you write some comments explaining what these options do?

internal/dedupe/dedupe.go
```diff
@@ -2,12 +2,22 @@
...
-	"github.com/cloudevents/sdk-go/v2/event"
 )
 
-// GetEventKey creates a unique key from an event.
-func GetEventKey(namespace string, ev event.Event) string {
-	return fmt.Sprintf("%s-%s-%s", namespace, ev.Source(), ev.ID())
+// Deduplicator checks if an event is unique.
+type Deduplicator interface {
+	IsUnique(ctx context.Context, item Item) (bool, error)
+	Set(ctx context.Context, events ...Item) error
```
sagikazarmark, line 12: Doesn't this change introduce a race condition? (Setting and checking in different steps)

internal/sink/buffer.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"sync"
+)
+
+type SinkBuffer struct {
+	mu   sync.Mutex
+	data map[string]SinkMessage
+}
+
+func NewSinkBuffer() SinkBuffer {
```
sagikazarmark, line 12: In this case, you have to return a pointer to avoid copying the mutex.

internal/sink/buffer.go
```diff
@@ -0,0 +1,41 @@
...
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	return len(b.data)
+}
+
+func (b *SinkBuffer) Add(message SinkMessage) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	// Unique identifier for each message (topic + partition + offset)
+	key := message.KafkaMessage.String()
```
sagikazarmark, line 28: Can we introduce a method for this on `SinkMessage`? Kafka appearing here doesn't read well.

internal/sink/namespaces.go
```diff
@@ -0,0 +1,116 @@
...
+// validateEvent validates a single event by matching it with the corresponding meter if any
+func (a *NamespaceStore) validateEvent(ctx context.Context, event serializer.CloudEventsKafkaPayload, namespace string) error {
+	namespaceStore := a.namespaces[namespace]
+	if namespaceStore == nil {
+		// We drop events from unknown org
+		return NewProcessingError(fmt.Sprintf("namespace not found: %s", namespace), DROP)
+	}
+
+	// Validate a single event against multiple meters
+	found := 0
```
sagikazarmark, line 48: This could be a boolean from what I can tell.

internal/sink/namespaces.go
```diff
@@ -0,0 +1,116 @@
...
+
+	return nil
+}
+
+// validateEventWithMeter validates a single event against a single meter
+func validateEventWithMeter(meter *models.Meter, ev serializer.CloudEventsKafkaPayload) *ProcessingError {
+	// We can skip count events with no group bys
+	if meter.Aggregation != models.MeterAggregationCount && len(meter.GroupBy) == 0 {
+		return nil
+	}
```
sagikazarmark, line 73: Here is an idea to make this more readable:

- Validate group bys first (if there are no group bys, the loop will do nothing)
- if the meter is a count meter, exit with nil
- continue running the rest of the function without any further checks for aggregation

From what I can tell even the json unmarshal is unnecessary for the count aggregation, so reorganizing the function like this wouldn't just optimize it, but would also improve readability (unless there is a reason why this wouldn't work)

internal/sink/sink.go
```diff
@@ -0,0 +1,544 @@
...
+	config           *SinkConfig
+	running          bool
+	buffer           SinkBuffer
+	flushTimer       *time.Timer
+	namespaceStore   *NamespaceStore
+	namespaceRefetch *time.Timer
+}
+
+type SinkConfig struct {
+	Context                 context.Context
```
sagikazarmark, line 42: This is generally an anti-pattern. Unless there is a very good reason, a context should be created for each request.

internal/sink/sink.go
```diff
@@ -0,0 +1,544 @@
...
+
+	consumer, err := kafka.NewConsumer(&config.ConsumerKafkaConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create consumer: %s", err)
+	}
+
+	producer, err := kafka.NewProducer(&config.ProducerKafkaConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create producer: %w", err)
+	}
```
sagikazarmark, line 73: It would be nicer to inject these.

internal/sink/sink.go
```diff
@@ -0,0 +1,544 @@
...
+		if err != nil {
+			logger.Error("failed to commit offset to kafka", "err", err)
+			// do not return error here as we want to write Redis even if commit fails as we already saved them in ClickHouse
+		}
+		logger.Debug("succeeded to commit offset to kafka", "offsets", commitedOffsets)
+	}
+
+	// 3. Sink to Redis
+	// Least once guarantee, if Redis write fails we will accept messages with same idempotency key in future and
+	// we have to relay on ClickHouse's deduplication.
```
sagikazarmark, line 191:
```suggestion
	// we have to rely on ClickHouse's deduplication.
```

Typo

internal/sink/sink.go
```diff
@@ -0,0 +1,544 @@
...
+				Namespace: message.Namespace,
+				ID:        message.Serialized.Id,
+				Source:    message.Serialized.Source,
+			})
+		}
+
+		err := s.config.Deduplicator.Set(s.config.Context, dedupeItems...)
+		if err != nil {
+			logger.Error("failed to sink to redis", "err", err)
+			return fmt.Errorf("failed to sink to redis: %s", err)
```
sagikazarmark, line 206: Thumb rule: if an error is returned, it should not be logged there. The parent should take care of that. If that doesn't work here, please leave a comment.

internal/sink/sink.go
```diff
@@ -0,0 +1,544 @@
...
+			if err != nil {
+				return fmt.Errorf("failed to deadletter messages: %s", err)
+			}
+		}
+	}
+
+	// 2. Commit Offset to Kafka
+	// Least once guarantee, if offset commit fails we will potentially process the same messages again as they are not committed yet
+	// If Redis write succeeds but Kafka commit fails we will reprocess messages without side effect as they will be dropped at duplicate check
+	// If Redis write fails we will double write them to ClickHouse and we have to relay on ClickHouse's deduplication
```
sagikazarmark, line 174:
```suggestion
	// If Redis write fails we will double write them to ClickHouse and we have to rely on ClickHouse's deduplication
```

Typo

internal/sink/sink.go
```diff
@@ -0,0 +1,544 @@
...
+
+		if len(deadletterMessages) > 0 {
+			err := s.deadLetter(deadletterMessages...)
+			if err != nil {
+				return fmt.Errorf("failed to deadletter messages: %s", err)
+			}
+		}
+	}
+
+	// 2. Commit Offset to Kafka
```
sagikazarmark, line 171: I wonder if it would make sense to introduce some sort of retry logic here.

internal/sink/sink.go
sagikazarmark: Also, do we use the producer for anything, but deadletter? Would it make sense to introduce an interface for invalid messages and make Kafka an implementation detail there? We could add other reporting options in the future?

internal/sink/sink.go
```diff
@@ -0,0 +1,544 @@
...
+	namespaceStore := NewNamespaceStore()
+	for _, meter := range meters {
+		namespaceStore.AddMeter(meter.Namespace, meter)
+	}
+
+	return namespaceStore, nil
+}
+
+func (s *Sink) subscribeToNamespaces() error {
+	logger := s.config.Logger.With("sink", "subscribeToNamespaces")
```
sagikazarmark, line 260:
```suggestion
	logger := s.config.Logger.With("operation", "subscribeToNamespaces")
```

We could add a "component => sink" in a higher layer.

internal/sink/sink.go
```diff
@@ -0,0 +1,544 @@
...
+	ns, err := s.getNamespaces()
+	if err != nil {
+		return fmt.Errorf("failed to get namespaces: %s", err)
+	}
+
+	// Identify if we have a new namespace
+	isNewNamespace := len(ns.namespaces) != len(s.namespaceStore.namespaces)
```
sagikazarmark, line 267: Probably an edge case, but: what if a namespace is deleted and a new one is created?

internal/sink/sink.go
```diff
@@ -0,0 +1,544 @@
...
+	// We always replace store to ensure we have latest meter changes
+	s.namespaceStore = ns
+
+	// We only subscribe to topics if we have a new namespace
+	if isNewNamespace {
+		// We always subscribe to all namespaces as consumer.SubscribeTopics replaces the current subscription.
+		topics := getTopics(*s.namespaceStore)
+		logger.Info("new namespaces detected, subscribing to topics", "topics", topics)
+
+		err = s.consumer.SubscribeTopics(topics, s.rebalanceCallback)
```
sagikazarmark, line 286: Does this cleanup old subscriptions (ie. deleted namespaces?)

internal/sink/sink.go
```diff
@@ -0,0 +1,544 @@
...
+				// Errors should generally be considered
+				// informational, the client will try to
+				// automatically recover.
+				logger.Error("kafka error", "code", e.Code(), "event", e)
+			case kafka.OffsetsCommitted:
+				// do nothing, this is an ack of the periodic offset commit
+				logger.Debug("kafka offset committed", "offset", e.Offsets)
+			default:
+				logger.Debug("kafka ignored event", "event", e)
+			}
```
sagikazarmark, line 405: Might make sense to extract this into a separate function

internal/sink/sink.go
```diff
@@ -0,0 +1,544 @@
...
+
+	// Schedule flush
+	s.flushTimer = time.AfterFunc(s.config.MaxCommitWait, flush)
+}
+
+func (s *Sink) clearFlushTimer() {
+	if s.flushTimer != nil {
+		s.flushTimer.Stop()
+	}
+}
```
sagikazarmark, line 318: This is technically not thread safe at the moment.

internal/sink/sink.go
```diff
@@ -0,0 +1,544 @@
...
+			// TODO: should we panic?
+			logger.Error("failed to subscribe to namespaces", "err", err)
+		}
+		s.namespaceRefetch = time.AfterFunc(s.config.NamespaceRefetch, refetch)
+	}
+	s.namespaceRefetch = time.AfterFunc(s.config.NamespaceRefetch, refetch)
+
+	// Reset state
+	s.running = true
+	s.setFlushTimer()
```
sagikazarmark, line 348: Where is this timer used at the moment?

internal/sink/storage.go
```diff
@@ -0,0 +1,246 @@
...
+)
+
+var codeRegexp = regexp.MustCompile(`code: (0-9]+)`)
+var jsonValueRegexp = regexp.MustCompile(`JSON_VALUE\(data, '(?P<path>[$._a-z]+)'\)(?:, 'Float64'\)\))? AS (?P<name>[a-z_]+)`)
+var eventTypeRegexp = regexp.MustCompile(`(?m:^WHERE [a-zA-z0-z_.]+_events\.type = '(?P<type>[a-zA-z0-9_-]+)'$)`)
+var namespaceRegexp = regexp.MustCompile(`(?m:^FROM (?P<namespace>[a-zA-z0-9_.]+)_events$)`)
+
+type Storage interface {
+	BatchInsert(ctx context.Context, namespace string, events []*serializer.CloudEventsKafkaPayload) error
+	GetMeters(ctx context.Context) ([]*models.Meter, error)
```
sagikazarmark, line 25: Should we use the meter repository instead?

pkg/models/meter.go
```diff
@@ -135,6 +135,7 @@ func WindowSizeFromDuration(duration time.Duration) (WindowSize, error) {
 }
 
 type Meter struct {
+	Namespace     string
```
sagikazarmark, line 138: Should this be returned in the API? If so, it should be added to openapi. If not, `json:"-"` should be added, but I'm not sure that's a good idea.

Also, if this information is necessary, we should probably introduce a static namespace store.

quickstart/docker-compose.ci.yaml
```diff
@@ -2,5 +2,11 @@ version: "3.9"
 
 services:
   openmeter:
-    image: "local"
+    image: "local-server"
```
sagikazarmark, line 5: Doesn't matter, this is a bogus image to make sure it's never found

sagikazarmark (changes requested):

internal/dedupe/dedupe.go
hekike: Depends on the usage. In the sink worker, it is okay because we parallelize processing per partition, and processing per partition is sequential while we also dedupe in the same batch in-memory.  

Inconsistency can happen in some edge cases still tough as we partition based on `subject`, so if someone sends in the same ID for a different subject at the exact same time, it can go to a parallel worker. It is an edge case as you shouldn't change the subject for the same ID. Optionally we can change the Kafka key (used by partitioning) to ID+Subject to protect against it. Originally we chose the subject as we needed data co-location with KSQLDB.

config/ingest.go
hekike: Opened an issue and added to comment: https://github.com/openmeterio/openmeter/issues/321

internal/sink/buffer.go
hekike: This is actually the Kafka Message's key, not the event/payload.

internal/sink/sink.go
hekike: It's fine, retry is basically that the batch will be processed again as offset doesn't update.

internal/sink/sink.go
hekike: We always overwrite local in-memory cache of namespaces, and Kafka subscribe is also an override.

hekike: See comment below:
> // We always subscribe to all namespaces as consumer.SubscribeTopics replaces the current subscription.

hekike: We always override local namespace cache. And although I haven't tested yet we remove kafka topics when we remove namespaces and kafka consumer client should handle that already as it handles rebalancing.

internal/sink/sink.go
hekike: Good idea, created an issue for future improvements: https://github.com/openmeterio/openmeter/issues/322

internal/sink/sink.go
hekike: We shouldn't have parallel flushes. Should I add a mutex?

pkg/models/meter.go
hekike: Let's discuss it here: https://github.com/openmeterio/openmeter/pull/323

internal/sink/sink.go
sagikazarmark: If there are absolutely no parallel flushes, a comment should be enough detailing that this is not concurrent safe and it's okay. Just in case we want to change things in the future.

sagikazarmark (approved):

PR #335, opened 2023-10-20
feat(sink): otel integration for worker

hekike:
OTel metrics and tracing integration for sink worker.

- Adding metrics
- Adding tracing
- Reorganizing code

Screenshot 2023-10-20 at 2 39 42 PM


internal/sink/sink.go
```diff
@@ -22,6 +27,12 @@ import (
 var namespaceTopicRegexp = regexp.MustCompile(`^om_([A-Za-z0-9]+(?:_[A-Za-z0-9]+)*)_events$`)
 var defaultDeadletterTopicTemplate = "om_%s_events_deadletter"
 
+var (
+	tracer            = otel.Tracer("sink")
+	metric            = otel.Meter("sink")
```
sagikazarmark, line 32: This may be problematic because Trace and Meter providers are (and should be) initialized in func main.

Also, the name should be something like `github.com/openmeterio/openmeter/...` or `openmeter.io/...`

sagikazarmark (changes requested):

PR #367, opened 2023-10-28
Lint openapi

sagikazarmark:
## Overview

Add a linter to make sure we catch openapi errors

Usage: `dagger call lint openapi`

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -324,14 +324,12 @@ components:
         - title
         - status
         - detail
-      example: |
-        {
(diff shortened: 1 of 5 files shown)


api/openapi.yaml
```diff
@@ -324,14 +324,12 @@ components:
         - title
         - status
         - detail
-      example: |
```
sagikazarmark, line 327: @tothandras please take a look at these fixes. According to the tool, this needs to be an object.

PR #371, opened 2023-10-29
feat(api)!: remove stateful management APIs

hekike:
Remove non-persistent HTTP APIs. 
OpenMeter OSS was intended to be config-driven and these APIs never got implemented in a stateful way.
They create confusion and are a source of bugs.

BREAKING CHANGE:

- return unimplemented for meter create HTTP API
- return unimplemented for meter delete HTTP API
- remove namespace create HTTP API
- remove namespace headers


sagikazarmark (approved):
Needs rebase. Other than that LGTM

PR #380, opened 2023-11-02
feat(sink): call hook on successful flush

hekike:


internal/sink/sink.go
```diff
@@ -67,6 +67,8 @@ type SinkConfig struct {
 	// DeadletterTopicTemplate is the template used to create the deadletter topic name per namespace.
 	// It is a sprintf template with the namespace as the only argument.
 	DeadletterTopicTemplate string
+	// Optional, Lifecycle Events Hooks
```
tothandras, line 70:
```suggestion
	// OnFlushSuccess is an optional lifecycle hook
```

tothandras (changes requested):

PR #395, opened 2023-11-08
fix(streaming): query params validator and timezones

hekike:
- Convert non UTC API inputs to UTC
- Fix query params validator to allow querying fine-grained window size meters on coarser window sizes.

### DIFF
--- a/internal/streaming/query_params.go
+++ b/internal/streaming/query_params.go
@@ -32,14 +32,26 @@ func (p *QueryParams) Validate(meterWindowSize models.WindowSize) error {
 	if p.WindowSize != nil {
 		err := isRoundedToWindowSize(*p.WindowSize, p.From, p.To)
 		if err != nil {
-			return fmt.Errorf("cannot query with %s: %w", *p.WindowSize, err)
+			return fmt.Errorf("cannot query with %s window size: %w", *p.WindowSize, err)
+		}
+
+		// Ensure query param window size is not smaller than meter window size
+		switch meterWindowSize {
+		case models.WindowSizeHour:
+			if p.WindowSize != nil && *p.WindowSize == models.WindowSizeMinute {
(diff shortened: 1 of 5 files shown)


internal/streaming/query_params.go
```diff
@@ -0,0 +1,88 @@
...
+
+	return nil
+}
+
+func isMinuteRounded(t time.Time) bool {
+	return t.Second() == 0
+}
+
+func isHourRounded(t time.Time) bool {
+	return isMinuteRounded(t) && t.Minute() == 0
```
tothandras, line 83: I'd rather duplicate the check logic for readability

```suggestion
	return t.Second() == 0 && t.Minute() == 0
```

internal/streaming/query_params.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func isDayRounded(t time.Time) bool {
+	return isMinuteRounded(t) && isHourRounded(t) && t.Hour() == 0
```
tothandras, line 87:
```suggestion
	return t.Second() == 0 && t.Minute() == 0 && t.Hour() == 0
```

internal/streaming/query_params.go
```diff
@@ -0,0 +1,88 @@
...
+	Aggregation    models.MeterAggregation
+	WindowSize     *models.WindowSize
+}
+
+func (p *QueryParams) Validate(meterWindowSize models.WindowSize) error {
+	if p.From != nil && p.To != nil {
+		if p.From.After(*p.To) {
+			return errors.New("from must be before to")
+		}
+		if p.From.Equal(*p.To) {
```
tothandras, line 26: Is this necessary?

tothandras (changes requested):

tothandras (approved):

PR #407, opened 2023-11-12
fix(streaming): escape sql

hekike:
Escape timezone input.


hekike:
Not necessary: `WindowTimeZone *time.Location`

PR #408, opened 2023-11-12
feat(sink): create deadletter topic with admin client

hekike:
Explicitly create dead-letter topics with a Kafka admin client.
Useful when Kafka broker setting is `auto.create.topics.enable: false`


hekike:
Closing in favor of: https://github.com/openmeterio/openmeter/pull/409

PR #409, opened 2023-11-13
feat(sink): save invalid events to storage

hekike:
Deadletter invalid events to storage instead of Kafka.


hekike:
Included in https://github.com/openmeterio/openmeter/pull/410

PR #444, opened 2023-11-22
Implement consumer portal feature and Web client

tothandras:
- Create web client
- Update quickstart
- Add Next.js dashboard example
image


api/client/web/README.md
```diff
@@ -0,0 +1,12 @@
+# OpenMeter Web Client
+
```
hekike, line 2: Please add a blurb about when users should use this library. Maybe add a similar to Node SDK too.
Something like:
> Client library for integrating web applications with OpenMeter. Useful to build customer dashboards. The library comes with integrations for React hooks.

api/client/web/src/react/hooks.ts
```diff
@@ -0,0 +1,23 @@
...
+
+import { useContext, useMemo } from 'react'
+import { OpenMeterClient } from '../client'
+import { OpenMeterContext } from './context'
+
+export function useOpenMeter() {
+  const context = useContext(OpenMeterContext)
+
+  if (!context) {
+    throw new Error('useOpenMeter must be used within an OpenMeterProvider')
```
hekike, line 11: Consider linking to docs [api/client](https://openmeter.io/docs/portal)

api/openapi.yaml
```diff
@@ -481,8 +537,8 @@ components:
       type: object
       required:
         - subject
-        - expiresAt
-        - token
+        # - expiresAt
```
hekike, line 540: WHy is this commented out?

examples/customer-portal-nextjs/.env.example
hekike, line 1: Wondering if example should be called: `customer-dashboard-nextjs`?

api/openapi.yaml
tothandras: The OpenAPI generator doesn't support readOnly, writeOnly: https://github.com/drwpow/openapi-typescript/issues/604

PR #487, opened 2023-12-15
feat(namespace): delete

hekike:
Delete namespace:

- Delete Kafka topic belonging to namespace
- Delete ClickHouse materialized views belonging to namespace meters


sagikazarmark (commented):
Wouldn't it be nicer if the individual handler would list meters instead of adding them to the interface?

The only place meters are used right now is in the clickhouse namespace handler. I see no reason why it needs to be added to the interface instead of that one implementation.

hekike:
@sagikazarmark good point, I refactored it.

sagikazarmark (approved):

PR #524, opened 2024-01-14
refactor(models/meter): remove redundant len check in `Validate`

Juneezee:
## Overview

From the Go specification [^1]:

> "3. If the map is nil, the number of iterations is 0."

`len` returns 0 if the map is nil [^2]. Therefore, checking `len(v) != 0` around a loop is unnecessary.

[^1]: https://go.dev/ref/spec#For_range
[^2]: https://pkg.go.dev/builtin#len

## Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/pkg/models/meter.go
+++ b/pkg/models/meter.go
@@ -222,11 +222,9 @@ func (m *Meter) Validate() error {
 		}
 	}
 
-	if len(m.GroupBy) != 0 {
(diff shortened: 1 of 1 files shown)


sagikazarmark (approved):
Thanks!

PR #534, opened 2024-01-19
feat(api): add subjects

hekike:
- Add subjects stub

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -278,7 +278,7 @@ paths:
     get:
       operationId: listSubjects
       tags:
-        - Cloud
+        - Subject
       responses:
         "200":
           description: OK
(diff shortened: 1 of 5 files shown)


api/openapi.yaml
```diff
@@ -273,6 +273,48 @@ paths:
...
+              schema:
+                type: array
+                items:
+                  $ref: "#/components/schemas/Subject"
+        default:
+          $ref: "#/components/responses/UnexpectedProblemResponse"
+    post:
+      operationId: upsertSubject
+      tags:
+        - Cloud
```
sagikazarmark, line 296: Do we want this in OSS?

hekike: Good catch, should be `Subject`

sagikazarmark (approved):

PR #568, opened 2024-01-24
feat(sink): kafka cooperative consumer

hekike:
Switching to cooperative sticky consumers to better distribute load and enable incremental rebalancing.


cmd/sink-worker/main.go
```diff
@@ -258,6 +258,10 @@ func initSink(config config.Configuration, logger *slog.Logger, metricMeter metr
 	_ = consumerKafkaConfig.SetKey("enable.auto.commit", false)
 	_ = consumerKafkaConfig.SetKey("enable.auto.offset.store", false)
 	_ = consumerKafkaConfig.SetKey("go.application.rebalance.enable", true)
+	// Used when offset retention resets the offset. In this case we want to consume from the latest offset as everything before should be already processed.
+	_ = consumerKafkaConfig.SetKey("auto.offset.reset", "latest")
+	// Guarantees an assignment that is maximally balanced while preserving as many existing partition assignments as possible.
+	_ = consumerKafkaConfig.SetKey("partition.assignment.strategy", "cooperative-sticky")
```
jeffchao, line 264: LGTM, just measure it.

There's an older blog post around cooperative (incremental) rebalancing. tl;dr is to measure throughput in records/s.

The idea is to improve rebalances by not doing a stop-the-world approach -- it's the difference between doing things eagerly vs not (aka cooperatively).

[0] Older blog - https://www.confluent.io/blog/cooperative-rebalancing-in-kafka-streams-consumer-ksqldb/
[1] Older blog 2 - https://www.confluent.io/blog/incremental-cooperative-rebalancing-in-kafka/
[3] Original JIRA of the KIP - https://cwiki.apache.org/confluence/display/KAFKA/KIP-429%3A+Kafka+Consumer+Incremental+Rebalance+Protocol

jeffchao: Also it seems like you're subscribing to multiple topics from a consumer, via a list of topics, so this config makes sense.

internal/sink/partition.go
```diff
@@ -0,0 +1,23 @@
...
+	for i := range partitions {
+		topicName := ""
+		if partitions[i].Topic != nil {
+			topicName = *partitions[i].Topic
+		}
+		out = append(out, fmt.Sprintf("%s-%d", topicName, partitions[i].Partition))
+	}
+
+	sort.Strings(out)
+	return out
```
sagikazarmark, line 22:
```suggestion
	out := make([]string, 0, len(partitions))

	for _, partition := range partitions {
		var topicName string
		
		if partition.Topic != nil {
			topicName = *partition.Topic
		}
		
		out = append(out, fmt.Sprintf("%s-%d", topicName, partition.Partition))
	}

	sort.Strings(out)
	
	return out
```

sagikazarmark: What if the topic is empty? Is (missing topic) "-1" valid?

hekike: It's unclear why the topic can be nil: https://github.com/confluentinc/confluent-kafka-go/blob/27274c264f39cdf429a88a762f9796f36837f7b0/kafka/kafka.go#L279

But as this method is only used to pretty logs it's not a problem.

tothandras (approved):

PR #611, opened 2024-02-13
Add better error handling

sagikazarmark:
## Overview
Add better error handling

4xx error logging is removed: we can enable access logs for that
Error handler logs 5xx error (when necessary)

It's going to become significantly easier, once we move away from HTTP handlers for core logic (most error mapping can happen outside of controllers)

### DIFF
--- a/internal/server/router/meter.go
+++ b/internal/server/router/meter.go
@@ -40,8 +40,6 @@ func (a *Router) CreateMeter(w http.ResponseWriter, r *http.Request) {
 
 	err := fmt.Errorf("not implemented: manage meters via config or checkout OpenMeter Cloud")
 
-	// TODO: caller error, no need to pass to error handler
-	a.config.ErrorHandler.HandleContext(ctx, err)
 	models.NewStatusProblem(ctx, err, http.StatusNotImplemented).Respond(w, r)
 }
 
@@ -65,8 +63,6 @@ func (a *Router) GetMeter(w http.ResponseWriter, r *http.Request, meterIdOrSlug 
 
 	// TODO: remove once meter model pointer is removed
 	if e := (&models.MeterNotFoundError{}); errors.As(err, &e) {
-		// TODO: caller error, no need to pass to error handler
-		a.config.ErrorHandler.HandleContext(ctx, err)
 		models.NewStatusProblem(ctx, err, http.StatusNotFound).Respond(w, r)
 
 		return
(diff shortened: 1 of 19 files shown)


internal/server/router/meter.go
```diff
@@ -3,23 +3,27 @@ package router
...
+	ctx := contextx.WithAttr(r.Context(), "operation", "listMeters")
+
 	namespace := a.config.NamespaceManager.GetDefaultNamespace()
 
 	meters, err := a.config.Meters.ListMeters(r.Context(), namespace)
 	if err != nil {
 		err := fmt.Errorf("list meters: %w", err)
-		models.NewStatusProblem(r.Context(), err, http.StatusInternalServerError).Respond(logger, w, r)
+
+		a.config.ErrorHandler.HandleContext(ctx, err)
```
hekike, line 24: this is such a commonly used function, wondering if we could add it to router and just be `a.HandleError(ctx, err))` or bake it it into NewStatusProblem.Respond

internal/server/router/meter_query.go
```diff
@@ -12,33 +13,42 @@ import (
...
+	"github.com/openmeterio/openmeter/pkg/contextx"
 	"github.com/openmeterio/openmeter/pkg/models"
 )
 
 // QueryMeter queries the values stored for a meter.
 func (a *Router) QueryMeter(w http.ResponseWriter, r *http.Request, meterIDOrSlug string, params api.QueryMeterParams) {
-	logger := slog.With("operation", "queryMeter", "id", meterIDOrSlug, "params", params)
+	ctx := contextx.WithAttr(r.Context(), "operation", "queryMeter")
+	ctx = contextx.WithAttr(ctx, "id", meterIDOrSlug)
+	ctx = contextx.WithAttr(ctx, "params", params) // TODO: HOW ABOUT NO????
```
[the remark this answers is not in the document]
sagikazarmark: Yeah, yeah, that was born out of my frustration.

internal/server/router/meter.go
sagikazarmark: If we add anything back to the status problem, I'm gonna have to cry myself to sleep. 😄 

The right solution is not baking business logic into HTTP handlers. Then we don't have to handle and map every error manually every single time, but we can rely on middleware instead.

Adding logging to the status problem was always a half measure (if you can call it that). If it doesn't hurt enough though, we are never going to fix it.

tl;dr The explicitness will lead us to a better solution eventually.

hekike (approved):

tothandras (approved):

PR #631, opened 2024-02-21
fix(api): make windowSize optional

tothandras:
Set to MINUTE by default in Cloud.


hekike (approved):

hekike (commented):
This will change read type as well including SDKs

GAlexIHU:
Closing due to inactivity

PR #634, opened 2024-02-22
test: add failing test for empty group by

sagikazarmark:
## Overview

Empty group bys lead to failing queries.


tothandras (approved):

internal/streaming/clickhouse_connector/query.go
```diff
@@ -123,6 +123,11 @@ func (d createMeterView) toSQL() (string, []interface{}, error) {
 	orderBy := []string{"windowstart", "windowend", "subject"}
 	sortedGroupBy := sortedKeys(d.GroupBy)
 	for _, k := range sortedGroupBy {
+		// Skip empty keys
+		if k == "" {
+			continue
```
hekike, line 128: I would return error. I think from operational perspective it's better to be strict and surface issues.

GAlexIHU:
Closing due to inactivity

PR #713, opened 2024-03-20
fix(python): client generate

hekike:
Build our own image as the [official autorest Dockerfile](https://github.com/Azure/autorest/blob/63ffe68961e24ed8aa59a2ca4c16a8019c271e45/docker/base/ubuntu/Dockerfile), which is outdated and not compatible with the latest autorest.
More specifically, the latest autorest npm package depends on other Azure packages that require a higher node version.

Changes:

- Use custom image for `autorest` with newer node version to fix python client generation
- Free up disk for `test`, `lint`, and `e2e` CI jobs to prevent out-of-space errors
- Only run `rm` commands for Dagger CI disk freeing step to speed up the build


.github/workflows/ci.yaml
```diff
@@ -231,8 +240,20 @@ jobs:
...
+          tool-cache: true
+
+          # These run slower as use apt to uninstall packages
+          # Turned on by default, so we disable them
+          android: false
+          dotnet: false
+          haskell: false
+          large-packages: false
+          docker-images: false
+          swap-storage: false
```
tothandras, line 256:
```suggestion
      - name: Free disk space
        uses: ./.github/actions/free-disk-space
```

tothandras (changes requested):

tothandras (approved):

.github/workflows/ci.yaml
hekike: wouldn't work as we don't checkout repo in this step

PR #805, opened 2024-04-22
feat(credits): implement credits

hekike:
OpenMeter can handle complex scenarios related to credit management, usage, and rate limits.
You can grant credits via API for specific subjects with priorities, effective dates, and expiration dates.
OpenMeter then burns down credits in order based on these parameters and the ingested usage, ensuring optimal performance.

OpenMeter Credits can help to implement:

- Monthly usage-limit enforcement
- Implementing prepaid credits
- Combining recurring credits with top-ups
- Rolling over remaining credits between billing periods

## Quickstart

Run docker compose with `--profile postgres` and enable credits in `config.yaml` via `credits: true`

## Credit Burndown

credit-burndown

## Action Items

- [x] Mark Feature and API experimental
- [x] Dagger Postgres integration (to run unit and e2e)
- [x] Rename Product to Feature
- [x] `Credits (Experimental)` OpenAPI Tag

## Open Questions

- Should we move watermark checks from the router into the model?
- Should we move the credit consumed check in grant void into the model from the router?
- Should we rename GET `/credit-ledger` to `/ledger` or `/credit-history`
- Should we use decimals for amounts


internal/credit/postgres_connector/grant.go
```diff
@@ -0,0 +1,401 @@
...
+}
+
+// Reset resets the ledger for the subject.
+// Rolls over grants with rollover configuration.
+func (c *PostgresConnector) Reset(ctx context.Context, namespace string, reset credit_model.Reset) (credit_model.Reset, []credit_model.Grant, error) {
+	var rollovedGrants []credit_model.Grant
+
+	// Lock ledger for the subject
+	logger := c.logger.With("operation", "reset", "namespace", namespace, "subject", reset.Subject, "id", reset.ID)
+	lock, err := c.lockManager.Obtain(ctx, namespace, reset.Subject)
```
turip, line 246: Having an in memory lock will not scale if we are using multiple instances of the ledger for HA (or edge use-case).

Given that most of the code (besides Reset) manipulates a few rows, and given that this is using postgres I think it would be better to rely on postgres' built-in transaction support and [SELECT FOR UPDATE schemantics](https://www.postgresql.org/docs/current/sql-select.html#SQL-FOR-UPDATE-SHARE).

What you can do is to have a table for each ledger, e.g. 
```
CREATE TABLE ledgers (
   namespace VARCHAR(X)
   subject VARCHAR(X)
   -- cache fields such as credits, last reset record reference etc
)
-- + indexes
```

Then in case you want to forbid any updates to the specific ledger you can:
- start a transaction for each query
- do 
```
SELECT * FROM ledgers WHERE namespace = 'XXX'   and subject = 'XXX' FOR UPDATE;
```

This will make postgres sequence the calls instead of the in memory lock.

Generally speaking however this is still problematic as this will still be a global lock per namespace per subject.

pkg/credit/ledger.go
```diff
@@ -0,0 +1,139 @@
...
+func (c LedgerEntryList) Render(w http.ResponseWriter, r *http.Request) error {
+	return nil
+}
+
+func (f LedgerEntryList) GetEntries() []LedgerEntry {
+	list := make([]LedgerEntry, len(f.list))
+	_ = copy(list, f.list)
+
+	// Sort ledger entries by type first
+	sort.Slice(list, func(i, j int) bool {
```
turip, line 75: Why not merge these into:

```
sort.Slice(list, func(i, j int) bool {
   if  list[i].Time.Equal(list[j].Time) {
             return  entryTypeWeight[list[i].Type] < entryTypeWeight[list[j].Type]
   }
   
   return list[i].Time.Before(list[j].Time)
}
   
```

pkg/credit/ledger.go
```diff
@@ -0,0 +1,139 @@
...
+		return entryTypeWeight[list[i].Type] < entryTypeWeight[list[j].Type]
+	})
+
+	// Sort ledger entries by time second
```
turip, line 79: Right now we are sorting by exact timestamp not by second. I don't know if it's just the comment or if we are missing a Truncate(time.Second) at creation.

This also means that the previous sort would only put the grant uses first if they happened at the exact same time.

But we are generating the timestamp on the service's side meaning that the timestamps would only match in case we are running multiple pods from this if the pods are on the same timestamp https://github.com/openmeterio/openmeter/blob/6b32fc5927f663328cf7a0c471b694571ee2086a/internal/credit/postgres_connector/ent/schema/schema.go#L42

(Using the database's NOW() function can be common timesource)

e2e/e2e_test.go
```diff
@@ -459,3 +460,217 @@ func TestQuery(t *testing.T) {
...
+
+	// Ingore events
+	{
+		resp, err := client.IngestEventBatchWithResponse(context.Background(), events)
+		require.NoError(t, err)
+		require.Equal(t, http.StatusNoContent, resp.StatusCode())
+	}
+
+	// Wait for events to be processed
+	time.Sleep(10 * time.Second)
```
turip, line 510: I think youshould implement a call that checks if the events are processed in the specific amount of time.

It not just makes integration tests faster, but also checks if the prerequisites for the next test cases have been fulfilled, thus helps with troubleshooting (e.g. it clearly shows if we have an issue with event processing or ledger).

internal/credit/postgres_connector/grant_test.go
```diff
@@ -0,0 +1,238 @@
...
+			test: func(t *testing.T, connector credit.Connector, db_client *db.Client) {
+				ctx := context.Background()
+				p := createProduct(t, connector, namespace, products[0])
+				grant := credit_model.Grant{
+					Subject:     "subject-1",
+					ProductID:   p.ID,
+					Type:        credit_model.GrantTypeUsage,
+					Amount:      100,
+					Priority:    1,
+					EffectiveAt: time.Now().Truncate(0),
```
turip, line 86: This causes test failures at my side.

```
Error:      	Not equal: 
            	            	expected: credit.Grant{ID:(*string)(nil), ParentID:(*string)(nil), Subject:"subject-1", Type:"USAGE", ProductID:(*string)(0xc00049b470), Amount:100, Priority:0x1, EffectiveAt:time.Date(2024, time.April, 22, 8, 27, 26, 83721000, time.Local), Expiration:credit.ExpirationPeriod{Count:0x1, Duration:"DAY"}, Metadata:map[string]string(nil), Rollover:(*credit.GrantRollover)(nil), Void:false}
            	            	actual  : credit.Grant{ID:(*string)(nil), ParentID:(*string)(nil), Subject:"subject-1", Type:"USAGE", ProductID:(*string)(0xc0000da2d0), Amount:100, Priority:0x1, EffectiveAt:time.Date(2024, time.April, 22, 8, 27, 26, 83721902, time.Local), Expiration:credit.ExpirationPeriod{Count:0x1, Duration:"DAY"}, Metadata:map[string]string(nil), Rollover:(*credit.GrantRollover)(nil), Void:false}
            	            	
            	            	Diff:
            	            	--- Expected
            	            	+++ Actual
            	            	@@ -9,3 +9,3 @@
            	            	  EffectiveAt: (time.Time) {
            	            	-  wall: (uint64) 83721000,
            	            	+  wall: (uint64) 83721902,
            	            	   ext: (int64) 63849371246,
            	Test:       	TestPostgresConnectorGrants/VoidGrant
```

I would rather create  a test helper function called MustParseTime somewhere: 

```

func MustParseTime(tStr string) time.Time {
	parsedTime, err := time.Parse(time.RFC3339, tStr)
	if err != nil {
		panic(err)
	}

	return parsedTime.UTC()
}
```

And then rely on hardcoded times such as:
```
 EffectiveAt: timeutil.MustParseTime("2024-04-22T12:33:00")
```

internal/credit/postgres_connector/watermark.go
```diff
@@ -0,0 +1,44 @@
...
+	"entgo.io/ent/dialect/sql"
+
+	"github.com/openmeterio/openmeter/internal/credit/postgres_connector/ent/db"
+	db_credit "github.com/openmeterio/openmeter/internal/credit/postgres_connector/ent/db/creditentry"
+	credit_model "github.com/openmeterio/openmeter/pkg/credit"
+)
+
+// GetHighWatermark returns the high watermark for the given credit and subject pair.
+func (c *PostgresConnector) GetHighWatermark(ctx context.Context, namespace string, subject string) (credit_model.HighWatermark, error) {
+	lastReset, err := c.db.CreditEntry.Query().
```
turip, line 17: I would either add a limit 1 or use a `SELECT max() ...` query to help the query planner with making it explicit that we only need a single item.

internal/credit/postgres_connector/watermark.go
```diff
@@ -0,0 +1,44 @@
...
+		Where(
+			db_credit.Namespace(namespace),
+			db_credit.Subject(subject),
+			db_credit.EntryTypeEQ(credit_model.EntryTypeReset),
```
turip, line 21: You can consider storing the high watermark so that we don't have to do a lookup in a possibly huge table (and we could put that into the table I proposed for locking).

internal/server/router/credit_balance.go
```diff
@@ -0,0 +1,51 @@
...
+	// Get high watermark for credit
+	hw, err := a.config.CreditConnector.GetHighWatermark(ctx, namespace, subject)
+	if err != nil {
+		a.config.ErrorHandler.HandleContext(ctx, err)
+		models.NewStatusProblem(ctx, err, http.StatusInternalServerError).Respond(w, r)
+		return
+	}
+
+	// Check if query time is after high watermark
+	if cutline.Before(hw.Time) {
```
turip, line 34: This can also be problematic if the reset call was received by by a machine whose clock is ahead of this machine and a GetCreditBalance is issued right after.

internal/credit/postgres_connector/grant.go
```diff
@@ -0,0 +1,401 @@
...
+
+	grant, err = mapGrantEntity(entity)
+	if err != nil {
+		return grant, fmt.Errorf("failed to map grant entity: %w", err)
+	}
+	return grant, nil
+}
+
+// TODO: use grant ID as an argument to void grant
+func (c *PostgresConnector) VoidGrant(ctx context.Context, namespace string, grant credit_model.Grant) (credit_model.Grant, error) {
```
turip, line 61: User Story: I think it is a question how we want to handle negative balances.

If a subject has an X amounts of grant from a single source and Y amounts of usage (Y>0) and the operator voids a grant that means that the subject will go into a negative balance.

Given that these are prepaid accounts making a customer pay for the negative balance can be troublesome. We can say that we allow for negative balances, but is this something our users would want or do they want an error in such cases?

api/openapi.yaml
```diff
@@ -621,6 +625,314 @@ paths:
...
+        "204":
+          description: Product deleted.
+        "401":
+          $ref: "#/components/responses/UnauthorizedProblemResponse"
+        "404":
+          $ref: "#/components/responses/NotFoundProblemResponse"
+        default:
+          $ref: "#/components/responses/UnexpectedProblemResponse"
+  # Credit Balances
+  /api/v1/credit-balance/{subject}:
```
sagikazarmark, line 721: Why not `/api/v1/credit/balance`?

cmd/server/main.go
```diff
@@ -217,17 +229,6 @@ func main() {
...
-	// 	Collector:        ingestCollector,
-	// 	NamespaceManager: namespaceManager,
-	// 	Logger:           logger,
-	// 	ErrorHandler:     errorsx.NewAppHandler(errorsx.NewSlogHandler(logger)),
-	// })
-	// if err != nil {
-	// 	logger.Error("failed to initialize http ingest handler", "error", err)
-	// 	os.Exit(1)
-	// }
-
```
sagikazarmark, line 230: I left this here for a reason, so we can easily revert to it if necessary.

config/credits.go
```diff
@@ -0,0 +1,10 @@
+package config
+
+type CreditsConfig struct {
```
sagikazarmark, line 3:
```suggestion
type CreditsConfiguration struct {
```

For consistency

config/postgres.go
```diff
@@ -0,0 +1,17 @@
+package config
+
+import "errors"
+
+type PostgresConfig struct {
```
sagikazarmark, line 5:
```suggestion
type PostgresConfiguration struct {
```

pkg/credit/ledger.go
hekike: I did it, thanks

pkg/credit/ledger.go
hekike: Yes, it's just a comment. I removed `second`.

e2e/e2e_test.go
hekike: This pattern was already established. I opened a an issue to improve it: https://github.com/openmeterio/openmeter/issues/828

sagikazarmark (commented):
Left a couple comments.

Also, some earlier comments that I believe were not addressed:

I'm not sure I like the current API paths. For example: `/api/v1/credit-balance/{subject}`

Shouldn't it rather be `/api/v1/credit/balance/{subject}`?

Or something like `/api/v1/credit/ledgers/{subject}/balance`

Seems like we are mixing a few different API styles and API is the hardest to change, so we should get it as close to right as possible.

I also suggested that the `credit` package follows the code organization of some other existing packages: instead of `nope_connector` and `postgres_connector` packages, the content of those should go under `credit/creditadapter`.

Last, but not least: now that the implementation is not tightly coupled with ClickHouse, maybe we should reconsider making this a separate component from OpenMeter itself.

api/openapi.yaml
```diff
@@ -773,6 +1085,333 @@ components:
...
+            The time the ledger entry was created.
+          type: string
+          format: date-time
+          example: "2023-01-01T00:00:00Z"
+        amount:
+          description: |
+            The amount to apply. Can be positive or negative number. If applicable.
+          type: number
+          example: 100
+        from:
```
turip, line 1162: Maybe, instead of from/to can we use `usage_period.start` and `usage_period.end` or something that reflects that this is a metadata for the grant usage entry?

internal/server/router/feature.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+	list := slicesx.Map[credit_model.Feature, render.Renderer](features, func(feature credit_model.Feature) render.Renderer {
+		return &feature
+	})
+
+	_ = render.RenderList(w, r, list)
+}
+
+// Create feature, POST: /api/v1/features
+func (a *Router) CreateFeature(w http.ResponseWriter, r *http.Request) {
```
GAlexIHU, line 56: Do we validate somewhere to only allow meters with `MeterAggregationSum`?

hekike: That is a very good point!

api/api.go
```diff
@@ -1,2 +1,7 @@
 //go:generate go run github.com/deepmap/oapi-codegen/v2/cmd/oapi-codegen --config=codegen.yaml ./openapi.yaml
 package api
+
+const (
+	// DefaultCreditQueryLimit specifies how many entries to return by default for credit related queries
+	DefaultCreditsQueryLimit = 1000
```
sagikazarmark, line 6: Where is this constant used? If not in this package, it should be moved where it’s used

turip: It's used in the router layer, moved it there.

chrisgacsal (approved):
LGTM!

GAlexIHU (approved):

internal/credit/postgres_connector/transaction.go
```diff
@@ -0,0 +1,101 @@
...
+// lockLedger locks the ledger for the given namespace and subject to avoid concurrent updates
+func lockLedger(tx *db.Tx, ctx context.Context, namespace string, ledgerID ulid.ULID) (*db.Ledger, error) {
+	// Lock ledger for the subject with pessimistic update
+	ledgerEntity, err := tx.Ledger.Query().
+		Where(db_ledger.Namespace(namespace)).
+		Where(db_ledger.ID(pgulid.Wrap(ledgerID))).
+
+		// We use the ForUpdate method to tell ent to ask our DB to lock
+		// the returned records for update.
+		ForUpdate().
```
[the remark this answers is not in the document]
turip: This could not happen as we have a unique constraint on these two fields.

turip:
TODO: split reset request/response body

PR #816, opened 2024-04-23
feat(credits): pg lock

hekike:
Use Postgres row-level lock.


internal/credit/postgres_connector/transaction.go
```diff
@@ -0,0 +1,121 @@
...
+	ledgerEntity, err := tx.Ledger.Query().
+		Where(db_ledger.Namespace(namespace)).
+		Where(db_ledger.Subject(subject)).
+
+		// We use the ForUpdate method to tell ent to ask our DB to lock
+		// the returned records for update.
+		ForUpdate(
+			// We specify that the query should not wait for the lock to be
+			// released and instead fail immediately if the record is locked.
+			sql.WithLockAction(sql.NoWait),
```
turip, line 82: I think if there are a lot of parallel requests then this will cause random requests to fail due to lock acquisition.

We should not use the NoWait IMHO.

PR #830, opened 2024-04-27
test(e2e): no wait

hekike:
Implement a call that checks if the events are processed in a specific amount of time instead of static waits.

It speeds up integration tests and checks whether the prerequisites for the next test cases have been fulfilled, helping with troubleshooting (e.g., it clearly shows if we have an issue with event processing or ledger).

Before:
`┃ ok      github.com/openmeterio/openmeter/e2e    40.696s`

After
`┃ ok      github.com/openmeterio/openmeter/e2e    21.688s`


sagikazarmark:
Nice!

PR #844, opened 2024-05-03
refactor: clean up credits api spec

turip:
## Overview

This patch adjusts the path schema to the agreed on one.

It also fixes:
- Adds limit to the queries, so that we don't return infinite number of records
- Adds check on the existance of the meter
- Adds logic to calculate the expiresAt field from the API spec
- Refactor From/To ledger entries into a seperate type

## Checklist

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my code
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] My changes generate no new warnings
- [X] New and existing unit tests pass locally with my changes


.vscode/launch.json
```diff
@@ -0,0 +1,18 @@
+{
+    // Use IntelliSense to learn about possible attributes.
```
hekike, line 2: comment in JSON, magic

go.sum
```diff
@@ -1305,8 +1305,8 @@ github.com/garyburd/redigo v0.0.0-20150301180006-535138d7bcd7/go.mod h1:NR3MbYis
 github.com/gdamore/optopia v0.2.0/go.mod h1:YKYEwo5C1Pa617H7NlPcmQXl+vG6YnSSNB44n8dNL0Q=
 github.com/generikvault/gvalstrings v0.0.0-20180926130504-471f38f0112a h1:J8FuFJ7K+Hiwkla2kT9fVIVix+EZhAlDsZwRlfFI3MA=
 github.com/generikvault/gvalstrings v0.0.0-20180926130504-471f38f0112a/go.mod h1:ms6iGk40n2YQrbM9Sr6onzwYBD1q5D0T5DQmcaye6uU=
-github.com/getkin/kin-openapi v0.124.0 h1:VSFNMB9C9rTKBnQ/fpyDU8ytMTr4dWI9QovSKj9kz/M=
-github.com/getkin/kin-openapi v0.124.0/go.mod h1:wb1aSZA/iWmorQP9KTAS/phLj/t17B5jT7+fS8ed9NM=
+github.com/getkin/kin-openapi v0.123.0 h1:zIik0mRwFNLyvtXK274Q6ut+dPh6nlxBp0x7mNrPhs8=
```
hekike, line 1308: these will disappear I think if you run `go mod tidy` after `go generate`

internal/server/router/credit_grant.go
```diff
@@ -11,25 +11,47 @@ import (
...
 
-	var subjects []string
-	if params.Subject != nil {
-		subjects = *params.Subject
+	// Get grants
+	grants, err := a.config.CreditConnector.ListGrants(ctx, namespace, credit_connector.ListGrantsParams{
+		Subjects:          defaultx.WithDefault(params.Subject, []string{}),
+		FromHighWatermark: true,
+		IncludeVoid:       true,
+		Limit:             defaultx.WithDefault(params.Limit, api.DefaultCreditsQueryLimit),
```
hekike, line 29: that's a cool party trick

hekike (approved):

go.sum
turip: It won't. The 0.124.0 is incompatble with the openapi generator, so this is intentional

hekike: Shouldn't this be an explicit override in go.mod to avoid merge bot updating it?

PR #856, opened 2024-05-06
refactor(sink-worker): context and error handling

chrisgacsal:
## Overview

* endure there is a single _root_ `context.Context` which handles application lifecycle and all children context(s) are derived from it
* use dedicated context for shutdown operations as root/parent context might be canceled which would prevent the shutdown to run/succeed

Fixes #(issue)

Exit code

* use dedicated exit codes when process is exited normally, due to OS signal or permanent error occured

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] New and existing unit tests pass locally with my changes


cmd/sink-worker/main.go
```diff
@@ -163,42 +181,51 @@ func main() {
...
 
 	// Setup signal handler
-	group.Add(run.SignalHandler(ctx, syscall.SIGINT, syscall.SIGTERM))
-
-	err = group.Run()
-	if e := (run.SignalError{}); errors.As(err, &e) {
-		slog.Info("received signal; shutting down", slog.String("signal", e.Signal.String()))
-	} else if !errors.Is(err, http.ErrServerClosed) {
-		logger.Error("application stopped due to error", slog.String("error", err.Error()))
+	group.Add(run.SignalHandler(ctx, syscall.SIGINT, syscall.SIGTERM, syscall.SIGKILL))
```
sagikazarmark, line 210: You can't register a handler for `SIGKILL`. I guess it's just noop in this case.

cmd/sink-worker/main.go
```diff
@@ -163,42 +181,51 @@ func main() {
...
+
+	// Run actors
+	var retCode int
```
sagikazarmark, line 213:
```suggestion
	var exitCode int
```

nit

cmd/sink-worker/main.go
```diff
@@ -163,42 +181,51 @@ func main() {
...
+	if err = group.Run(); err != nil {
+		signalErr := &run.SignalError{}
+		switch {
+		case errors.As(err, signalErr):
```
sagikazarmark, line 217: This is why I like if/else better in this case. There is no init statement for `case`.

internal/sink/sink.go
```diff
@@ -431,16 +429,18 @@ func (s *Sink) Run() error {
 	s.namespaceRefetch = time.AfterFunc(s.config.NamespaceRefetch, refetch)
 
 	// Reset state
-	s.running = true
+	s.isRunning.Store(true)
+	defer s.Close()
```
sagikazarmark, line 433: `sink.Close()` is called in main by the group function as well (which makes sense IMO). Why is it called here?

internal/sink/sink.go
```diff
@@ -593,15 +592,13 @@ func (s *Sink) rebalance(c *kafka.Consumer, event kafka.Event) error {
...
+		logger.Error("unexpected event type", "event", e)
 	}
 
 	return nil
 }
 
-func (s *Sink) ParseMessage(e *kafka.Message) (string, *serializer.CloudEventsKafkaPayload, error) {
-	ctx := context.TODO()
-
+func (s *Sink) ParseMessage(ctx context.Context, e *kafka.Message) (string, *serializer.CloudEventsKafkaPayload, error) {
```
sagikazarmark, line 601: If this is only called from run, it should be unexported.

internal/sink/sink.go
chrisgacsal: My thinking was to make sure that we properly "cleanup" if `sink.Run()` returns due to ctx cancellation, but I agree that this is not the best of handling this and it is better tor have the grpup function to call the `sinkClose()` when `sink.Run()` returns. I'll push an update ti fix this.

hekike (approved):

hekike (dismissed):

cmd/sink-worker/main.go
```diff
@@ -163,42 +181,54 @@ func main() {
...
+			default:
+				os.Exit(130)
+			}
+		}
+
+		if !errors.Is(err, http.ErrServerClosed) {
+			logger.Error("application stopped due to error", slog.String("error", err.Error()))
+			os.Exit(1)
+		}
 	}
```
sagikazarmark, line 229: What I meant is keep the original structure. 😄 

`if e := (run.SignalError{}); errors.As(err, &e) {`

cmd/sink-worker/main.go
```diff
@@ -163,42 +181,54 @@ func main() {
...
-	err = group.Run()
-	if e := (run.SignalError{}); errors.As(err, &e) {
-		slog.Info("received signal; shutting down", slog.String("signal", e.Signal.String()))
-	} else if !errors.Is(err, http.ErrServerClosed) {
-		logger.Error("application stopped due to error", slog.String("error", err.Error()))
+	// Run actors
+	if err = group.Run(); err != nil {
+		signalErr := &run.SignalError{}
+		if errors.As(err, signalErr) {
+			logger.Warn("received signal: shutting down", slog.String("signal", signalErr.Signal.String()))
```
sagikazarmark, line 216: Why is this warn?

cmd/sink-worker/main.go
```diff
@@ -163,42 +181,54 @@ func main() {
...
 	// Setup signal handler
 	group.Add(run.SignalHandler(ctx, syscall.SIGINT, syscall.SIGTERM))
 
-	err = group.Run()
-	if e := (run.SignalError{}); errors.As(err, &e) {
-		slog.Info("received signal; shutting down", slog.String("signal", e.Signal.String()))
-	} else if !errors.Is(err, http.ErrServerClosed) {
-		logger.Error("application stopped due to error", slog.String("error", err.Error()))
+	// Run actors
+	if err = group.Run(); err != nil {
```
sagikazarmark, line 213: This code should roughly be similar in different cmds.

cmd/sink-worker/main.go
```diff
@@ -163,42 +181,54 @@ func main() {
...
+			}
+		}
+
+		if !errors.Is(err, http.ErrServerClosed) {
+			logger.Error("application stopped due to error", slog.String("error", err.Error()))
+			os.Exit(1)
+		}
 	}
+
+	os.Exit(0)
```
sagikazarmark, line 231: That's the default IMO.

cmd/sink-worker/main.go
```diff
@@ -163,42 +181,54 @@ func main() {
...
+				os.Exit(0)
+			default:
+				os.Exit(130)
+			}
+		}
+
+		if !errors.Is(err, http.ErrServerClosed) {
+			logger.Error("application stopped due to error", slog.String("error", err.Error()))
+			os.Exit(1)
+		}
```
sagikazarmark, line 228: This needs to be en elseif

cmd/sink-worker/main.go
chrisgacsal: I agree, I do plan to make the same changes in the `server` component as well in a followup PR to avoid exploding this one.

cmd/sink-worker/main.go
chrisgacsal: I've seen projects using `warn` level for signal events to make those to "stick out" in the logs, but I ok with either approach so I'll switch it back to `info`.

sagikazarmark (approved):

PR #857, opened 2024-05-06
feat: provide ledger ID based rest API

turip:
## Overview

We want to provide REST endpoints for ledgers:

```
/api/v1/ledgers/{ledgerId=subject}/history (in period)
/api/v1/ledgers/{ledgerId=subject}/balance (as of)
/api/v1/ledgers/{ledgerId=subject}/grants
/api/v1/ledgers/{ledgerId=subject}/grants/{grantId}
/api/v1/ledgers/{ledgerId=subject}/reset (time)
```

To achieve this we need to move away from the on-demand ledger creation. This PR changes the existing behavior so that we provide the following two endpoints:

```
POST /api/v1/ledgers # create a new ledger for a specific subject
GET /api/v1/ledgers # enumerate existing ledgers
```

Then the resulting ledgerID can be used to reference the specific ledger.

Fixes #(issue)

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] New and existing unit tests pass locally with my changes


api/openapi.yaml
```diff
@@ -718,7 +718,86 @@ paths:
         default:
           $ref: "#/components/responses/UnexpectedProblemResponse"
   # Credit Balances
```
hekike, line 720: OpenAPI comment is for balance

api/openapi.yaml
```diff
@@ -718,7 +718,86 @@ paths:
...
-  /api/v1/ledgers/{creditSubjectId}/balance:
+  /api/v1/ledgers:
+    get:
+      operationId: listLedgers
+      summary: List the already defined ledgers.
+      description: |
+        List the already defined ledgers.
+      tags:
+        - Credits (Experimental)
```
hekike, line 728: Let's rename `Credits` label to `Entitlements` everywhere

internal/credit/postgres_connector/ledger.go
```diff
@@ -0,0 +1,110 @@
...
+
+	"github.com/openmeterio/openmeter/internal/credit"
+	"github.com/openmeterio/openmeter/internal/credit/postgres_connector/ent/db"
+	db_ledger "github.com/openmeterio/openmeter/internal/credit/postgres_connector/ent/db/ledger"
+	"github.com/openmeterio/openmeter/internal/credit/postgres_connector/ent/pgulid"
+	"github.com/openmeterio/openmeter/pkg/slicesx"
+)
+
+func (c *PostgresConnector) CreateLedger(ctx context.Context, namespace string, ledgerIn credit.Ledger, upsert bool) (credit.Ledger, error) {
+	dbLedger, err := transaction(ctx, c, func(tx *db.Tx) (*db.Ledger, error) {
```
hekike, line 17: have you considered always calling upsert and using OnConflict Do nothing

hekike (approved):

internal/credit/postgres_connector/ledger.go
turip: For the moment I would stay with the traditional REST schematics where we are not upserting anything. This API can be easily updated to support upserts in a non-breaking manner, however if we implement something right now we might need to do a breaking change in the future when we better understand the problem space for ledgers.

I don't see how we want to handle account deletion/closure and having an upsert call might get convoluted if we start doing things like account status, etc. I think we will see this as we are starting to address more and more use-cases. 

Also this API changes the expected behavior too (so I don't think we have value (at least for now) in upserting). If there's an external system we are doing the Entitlements for, what will happen is that:
- The external system has its own account database
- When needed it creates the ledger in our system, persists the ledger id into their account database
- They do grants based on this ID

So the Create Ledger will most probably called only on account creation. 

I might be wrong, but I think we should see actual use-cases both for upsert and account closure before adding specific APIs for there. (And we can add that if we want later on).

PR #904, opened 2024-05-14
feat: on ledger creation return the existing ledger if exists

turip:
## Overview

If a ledger for the specified subject already exists, let's return the existing ledger ID as a 409 response, so that no new query will be needed to fetch the ledger by subject.

Example response:
```json
{
    "type": "about:blank",
    "title": "Conflict",
    "status": 409,
    "detail": "ledger (default.01HXVNDJR532E8GTBVSC2XK5D4) already exitst for subject subject-1",
    "instance": "urn:request:Peters-MacBook-Pro-2.local/ZzyNVpzatY-000001",
    "conflictingEntity": {
        "id": "01HXVNDJR532E8GTBVSC2XK5D4",
        "subject": "subject-1"
    }
}
```

Fixes #(issue)

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] New and existing unit tests pass locally with my changes


api/openapi.yaml
```diff
@@ -1250,6 +1252,65 @@ components:
             type: string
           example:
             stripePaymentId: "pi_4OrAkhLvyihio9p51h9iiFnB"
+    LedgerAlreadyExistsProblem:
+      description: Ledger Exists
+      x-go-type-import:
+        path: github.com/openmeterio/openmeter/internal/credit
+      x-go-type: credit.LedgerAlreadyExistsProblemResponse
```
turip, line 1259: Duplication of type required due to this: https://github.com/deepmap/oapi-codegen/pull/1610

pkg/models/problem.go
```diff
@@ -66,27 +69,36 @@ func (p *StatusProblem) ProblemType() ProblemType {
...
-	w.WriteHeader(p.Status)
+	w.WriteHeader(problem.ProblemStatus())
 	_, _ = w.Write(buf.Bytes())
 }
 
 // NewStatusProblem will generate a problem for the provided HTTP status
 // code. The Problem's Status field will be set to match the status argument,
 // and the Title will be set to the default Go status text for that code.
-func NewStatusProblem(ctx context.Context, err error, status int) Problem {
+func NewStatusProblem(ctx context.Context, err error, status int) *StatusProblem {
```
GAlexIHU, line 101: why did you change this? (just curious)

GAlexIHU (approved):

pkg/models/problem.go
turip: Check this: https://go.dev/play/p/GGaWrq1Gu3n

The overal issue is that if an interface is composed with a struct, then json marshaling will treat that as a field member as opposed to being part of the same object.

If we want to define per service problems, like this: https://github.com/openmeterio/openmeter/pull/904/files#diff-f49bfff64b9ead64bbdc915228760a76fb213fcbc0474605aef0c0a5f47c8ad5R19 without putting them into models, we need to have access to the *StatusProblem.

Now I could have done a typecast there, but then if something changes in NewStatusProblem's implementation that might break runtime, so it's cleaner to expose the struct instead.

internal/credit/credit.go
```diff
@@ -227,12 +227,11 @@ func (e *LockErrNotObtainedError) Error() string {
...
 type LedgerAlreadyExistsError struct {
 	Namespace string
-	LedgerID  ulid.ULID
-	Subject   string
+	Ledger    Ledger
 }
 
 func (e *LedgerAlreadyExistsError) Error() string {
-	return fmt.Sprintf("ledger (%s.%s) already exitst for subject %s", e.Namespace, e.LedgerID, e.Subject)
+	return fmt.Sprintf("ledger (%s.%s) already exitst for subject %s", e.Namespace, e.Ledger.ID, e.Ledger.Subject)
```
sagikazarmark, line 234: Should we omit the namespace from the error? It's largely transparent anyway. It might be more confusing than useful.

PR #935, opened 2024-05-21
fix(deps): roll back kin-openapi for oapi-codegen compatibility

GAlexIHU:
## Overview

Downgrades `kin-openapi` for `oapi-codegen` compatibility. See [this issue](https://github.com/deepmap/oapi-codegen/issues/1533)

Fixes #(issue)

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #938, opened 2024-05-22
feat/balance api OM-573

GAlexIHU:
## Overview

Fixes #(issue)

OM-573


internal/credit/creditdriver/grant.go
```diff
@@ -249,3 +293,33 @@ func (b *builder) GetLedgerGrant() GetLedgerGrantHandler {
 		)...,
 	)
 }
+
+func mapGrantToAPI(grant credit.Grant) api.LedgerGrantResponse {
+	var featureID string
+
+	if grant.FeatureID != nil {
```
turip, line 300: Instead of these please use something like this: https://github.com/openmeterio/openmeter/blob/fcce3d310c161bd8ab17bfe0bbe88b933503dd58/pkg/defaultx/default.go#L3

internal/credit/creditdriver/grant.go
```diff
@@ -249,3 +293,33 @@ func (b *builder) GetLedgerGrant() GetLedgerGrantHandler {
...
+		featureID = string(*grant.FeatureID)
+	}
+
+	priority := int(grant.Priority)
+
+	return api.LedgerGrantResponse{
+		Amount:      float32(grant.Amount),
```
turip, line 307: Let's use float64 or decimal.

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/openmeterio/openmeter
 
 go 1.22.0
 
+replace github.com/getkin/kin-openapi => github.com/getkin/kin-openapi v0.123.0
```
turip, line 5: This is not inherited in case somebody uses our API directly from the repo. Let's make sure that the client is working without this replace (I think it only affects the code generation).

internal/credit/creditdriver/balance.go
```diff
@@ -22,10 +22,16 @@ type GetLedgerBalanceRequest struct {
 	Cutline  time.Time
 }
 
-type GetLedgerBalanceHandler httptransport.HandlerWithArgs[GetLedgerBalanceRequest, credit.Balance, GetLedgerBalaceHandlerParams]
+type GetLedgerBalanceHandlerResponse struct {
```
turip, line 25: This type seems to be unused.

internal/credit/creditdriver/balance.go
```diff
@@ -61,3 +72,79 @@ func (b *builder) GetLedgerBalance() GetLedgerBalanceHandler {
...
+		FeatureBalances: featureBalances,
+		GrantBalances:   grantBalances,
+		LastReset:       &highwatermark.Time,
+		Metadata:        &balance.Metadata,
+		Subject:         balance.Subject,
+	}
+}
+
+func mapFeatureBalanceToAPI(featureBalance credit.FeatureBalance) api.FeatureBalance {
+	id := new(string)
```
turip, line 97: let's at least wrap this logic into a convert function.

internal/credit/creditdriver/grant.go
```diff
@@ -34,7 +34,34 @@ func (b *builder) ListLedgerGrants() ListLedgerGrantsHandler {
...
-		b.CreditConnector.ListGrants,
+		func(ctx context.Context, request credit.ListGrantsParams) ([]api.LedgerGrantWithLedgerIDResponse, error) {
+			grants, err := b.CreditConnector.ListGrants(ctx, request)
+			if err != nil {
+				return nil, err
+			}
+			resp := make([]api.LedgerGrantWithLedgerIDResponse, 0, len(grants))
+			for _, grant := range grants {
+				mapped := mapGrantToAPI(grant)
+				resp = append(resp, api.LedgerGrantWithLedgerIDResponse{
```
turip, line 45: I understand that this is only used here, but it might worth placing it closer to the mapGrantToAPI, so that at least people updating the grant API don't forget to update the ledger version too.

internal/credit/postgres_connector/grant.go
```diff
@@ -272,6 +272,19 @@ func mapGrantEntity(entry *db.CreditEntry) (credit.Grant, error) {
 		return credit.Grant{}, fmt.Errorf("entry type must be grant: %s", entry.EntryType)
 	}
 
+	rollover := &credit.GrantRollover{}
+	if entry.RolloverType != nil && entry.RolloverMaxAmount != nil {
+		rollover.Type = *entry.RolloverType
+		rollover.MaxAmount = entry.RolloverMaxAmount
+	} else {
+		rollover = nil
+	}
```
turip, line 281:
```suggestion
	var rollover *credit.GrantRollover
	
	if entry.RolloverType != nil && entry.RolloverMaxAmount != nil {
	         rollover = &credit.GrantRollover{
	           Type: *entry.RolloverType,
	           MaxAmount: entry.RolloverMaxAmount,
	         }
	}
```

internal/credit/postgres_connector/test_helpers.go
```diff
@@ -0,0 +1,44 @@
+package postgres_connector
```
turip, line 1: Can we name it helpers_test.go or something, so that it's not available in normal builds only in tests?

go.mod
GAlexIHU: i'm not sure what you mean by " Let's make sure that the client is working without this replace". The bug only affects the code generation.

also https://github.com/openmeterio/openmeter/pull/935

internal/credit/creditdriver/grant.go
GAlexIHU: changed everywhere

internal/credit/creditdriver/grant.go
GAlexIHU: updated with annotations & changed

internal/credit/creditdriver/grant.go
GAlexIHU: moved it next to it

internal/credit/postgres_connector/test_helpers.go
GAlexIHU: moved it to a dedicated package

internal/credit/creditdriver/balance.go
```diff
@@ -61,3 +66,79 @@ func (b *builder) GetLedgerBalance() GetLedgerBalanceHandler {
...
+	if featureBalance.ID != nil {
+		*id = string(*featureBalance.ID)
+	} else {
+		id = nil
+	}
```
sagikazarmark, line 96: This smells. The else is unnecessary.

internal/credit/creditdriver/balance.go
```diff
@@ -61,3 +66,79 @@ func (b *builder) GetLedgerBalance() GetLedgerBalanceHandler {
...
+		*id = string(*grantBalance.ID)
+	} else {
+		id = nil
+	}
+	parentId := new(string)
+	if grantBalance.ParentID != nil {
+		*parentId = string(*grantBalance.ParentID)
+	} else {
+		parentId = nil
+	}
```
sagikazarmark, line 122: Same as above

api/openapi.yaml
```diff
@@ -955,7 +956,7 @@ paths:
               schema:
                 type: array
                 items:
-                  $ref: "#/components/schemas/LedgerGrantResponse"
+                  $ref: "#/components/schemas/LedgerGrantWithLedgerIDResponse"
```
sagikazarmark, line 959: This name is a bit weird. When do you return a response without an ID?

GAlexIHU: Fair, the 2 list endpoints had different types prior (i think), changed both to use the same

go.mod
turip: Thank you for confirming, that's what I was asking for ;-)

internal/credit/creditdriver/balance.go
```diff
@@ -61,3 +66,73 @@ func (b *builder) GetLedgerBalance() GetLedgerBalanceHandler {
...
+}
+
+func mapGrantBalanceToAPI(grantBalance credit.GrantBalance) api.LedgerGrantBalance {
+	var id *string
+	if grantBalance.ID != nil {
+		*id = string(*grantBalance.ID)
+	}
+	var parentId *string
+	if grantBalance.ParentID != nil {
+		*parentId = string(*grantBalance.ParentID)
```
turip, line 115: I think this will cause panic, as parentId is nil

internal/credit/creditdriver/balance.go
```diff
@@ -61,3 +66,73 @@ func (b *builder) GetLedgerBalance() GetLedgerBalanceHandler {
...
+		MeterSlug:           featureBalance.MeterSlug,
+		Name:                featureBalance.Name,
+		UpdatedAt:           featureBalance.UpdatedAt,
+		Usage:               featureBalance.Usage,
+		Balance:             featureBalance.Balance,
+	}
+}
+
+func mapGrantBalanceToAPI(grantBalance credit.GrantBalance) api.LedgerGrantBalance {
+	var id *string
```
turip, line 109: For these let's have a generic like this (and put that into a package please, e.g convert):

```
package main

import "fmt"

type TestID string

func ConvertString[Source, Dest ~string](src *Source) *Dest {
	if src == nil {
		return nil
	}

	out := Dest(*src)
	return &out
}

func main() {

	ID := TestID("test")
	fmt.Println(*ConvertString[TestID, string](&ID))
}
```

An alternative could be that in the openapi spec at least for the ID types we are referencing the credit types. That would make our life slightly easier and would not allow for any divergence on the struct contents.

internal/credit/creditdriver/grant.go
```diff
@@ -249,3 +276,52 @@ func (b *builder) GetLedgerGrant() GetLedgerGrantHandler {
...
+		Expiration: &api.LedgerGrantExpirationPeriod{
+			Count:    int(grant.Expiration.Count),
+			Duration: api.LedgerGrantExpirationPeriodDuration(grant.Expiration.Duration),
+		},
+		ExpiresAt: &grant.ExpiresAt,
+		FeatureID: string(defaultx.WithDefault(grant.FeatureID, credit.FeatureID(""))),
+		Id:        (*string)(grant.ID),
+		Metadata:  &grant.Metadata,
+		ParentId:  (*string)(grant.ParentID),
+		Priority:  &priority,
```
turip, line 296:
```suggestion
		Priority:  convert.ToPointer(int(grant.Priority))
```

internal/credit/creditdriver/grant.go
```diff
@@ -249,3 +276,52 @@ func (b *builder) GetLedgerGrant() GetLedgerGrantHandler {
...
+		Rollover:  grant.Rollover,
+		Type:      api.LedgerGrantType(grant.Type),
+		UpdatedAt: grant.UpdatedAt,
+		Void:      &grant.Void,
+	}
+}
+
+func mapGrantWithBalanceToAPI(grant credit.Grant) api.LedgerGrantResponse {
```
turip, line 304:
```suggestion
func mapGrantWithBalanceToAPI(grant credit.Grant) api.LedgerGrantResponse {
   res := mapGrantToAPI(grant)
   res.LedgerID = string(grant.LedgerID)
   return res
   }
```

turip: This way we at least see what we are adding as a plus

internal/credit/creditdriver/ledger.go
```diff
@@ -126,3 +134,25 @@ func (b *builder) GetLedgerHistory() GetLedgerHistoryHandler {
...
+		period = &api.Period{
+			From: entry.Period.From,
+			To:   entry.Period.To,
+		}
+	}
+	return api.LedgerEntry{
+		Id:        entryId,
+		Type:      api.LedgerEntryType(entry.Type),
+		Time:      entry.Time,
+		FeatureID: string(defaultx.WithDefault(entry.FeatureID, credit.FeatureID(""))),
```
turip, line 154:
```suggestion
		FeatureID: defaultx.WithDefault(string(entry.FeatureID), "")),
```

internal/credit/feature.go
```diff
@@ -28,6 +28,26 @@ func (e *FeatureNotFoundError) Error() string {
...
+
+type FeatureWithNameAlreadyExistsError struct {
+	Name string
+	ID   FeatureID
+}
+
+func (e *FeatureWithNameAlreadyExistsError) Error() string {
+	// Is it an issue that we leak ID on another Feature here?
+	// Shouldn't be an isue as it's namespaced.
+	return fmt.Sprintf("Feature %s with name %s already exists.", e.ID, e.Name)
```
turip, line 48: Usually go errors are starting with lowercase and don't have ending dots. 

Given that this is the API error we can use this format, but then IMHO we should be consistent with the other error messages (e.g. we should change those too).

internal/credit/creditdriver/balance.go
GAlexIHU: it wont https://go.dev/play/p/qa8fpV1Zffp

internal/credit/creditdriver/balance.go
GAlexIHU: i wasn't aware Dest ~string ... Dest(x) is possible in go 🤯

internal/credit/feature.go
GAlexIHU: changed to lowercase non-sentence

internal/credit/creditdriver/ledger.go
GAlexIHU: that actually doesnt work its a pointer

api/openapi.yaml
```diff
@@ -1406,13 +1413,16 @@ components:
           description: |
             The amount to apply. Can be positive or negative number. If applicable.
           type: number
+          format: double
           example: 100
+        featureID:
+          description: |
+            The unique feature ULID that the entry is associated with, if any.
```
tothandras, line 1420:
```suggestion
            The unique feature ULID that the entry is associated with.
```

turip (approved):

tothandras (approved):

api/openapi.yaml
tothandras: As it seems to be required, otherwise remove it from the required list.

tothandras (approved):

PR #959, opened 2024-05-31
test(balance): add breaking tests for balance calcs

GAlexIHU:
## Overview

Fixes issues related to windowed event processing. See testcase names and descriptions for details.


internal/credit/postgres_connector/balance.go
```diff
@@ -190,7 +194,7 @@ func (a *PostgresConnector) getBalance(
 			if expiresAt.Before(period.From) {
 				continue
 			}
-			if grantBalance.EffectiveAt.After(period.To) {
+			if grantBalance.EffectiveAt.After(period.To.Truncate(time.Minute)) {
```
turip, line 197: Let's add this to the docs that right now we have a 1 minute resolution.

cc @hekike please be aware of this limitation

internal/credit/postgres_connector/balance.go
```diff
@@ -217,7 +221,8 @@ func (a *PostgresConnector) getBalance(
 				// TODO: do we want this to be settable in ledger
 				FilterSubject: []string{ledger.Subject},
 			}
-			queryParams.From = &period.From
+			fromMinute := period.From.Truncate(time.Minute)
+			queryParams.From = &fromMinute
```
turip, line 225: This way we are overfetching the meters, counting even the events in the same minute happened before the period.From

internal/credit/postgres_connector/balance_test.go
```diff
@@ -44,15 +44,32 @@ func TestPostgresConnectorBalances(t *testing.T) {
 		Name:      "feature-2",
 	}
 
+	oldSetup := func(streamingConnector *mockStreamingConnector, connector credit.Connector) (credit.Ledger, error) {
```
hekike, line 47: Why is it called the old setup?

GAlexIHU:
In light of some deeper issues turned it to draft

GAlexIHU:
## Topics to discuss

### Grants changed so they take effect at the end of current window

By pushing the effectiveAt dates of grants to the end of the current window (rounding up to a minute), right after we issue a grant it wont show up in balance.
- This is expected & straightforward for balance calculation, enforcement based on balance, etc...
- For cloud this can get really confusing with current implementation as we simply display the balance info. What we should do in addition is display grants that will take effect in the next window too, mark them, and also display the features associated with them. IMO this should not be part of this API

### Lack of windowing is validated on Balance calc
`GetBalance` will error if the grants are not windowed. IMO I would keep this as without windowed grants it will return incorrect data. The issue is that we don't have a migration system. New data will be valid but it would break on old.
Suggestions?

internal/credit/postgres_connector/balance.go
```diff
@@ -193,6 +200,9 @@ func (a *PostgresConnector) getBalance(
 			if grantBalance.EffectiveAt.After(period.To) {
 				continue
 			}
+			if periodIndex != 0 && grantBalance.EffectiveAt.Equal(period.To) {
```
turip, line 203: If I understand correctly what you want to achieve here, we would only query the meter for the first feature IMHO if multiple grants are created in the same window period for multiple features. 

I would highly recommend adding a test case to multiple features with this grant setup.

internal/credit/postgres_connector/balance_test.go
```diff
@@ -490,103 +523,780 @@ func TestPostgresConnectorBalances(t *testing.T) {
...
+				balance, err := connector.GetBalance(context.Background(), credit.NewNamespacedLedgerID(namespace, ledger.ID), future.Add(windowSize*2))
+				assert.NoError(t, err)
+
+				// Both features and grants should be listed
+				assert.Equal(t, 2, len(balance.FeatureBalances))
+				assert.Equal(t, 2, len(balance.GrantBalances))
+			},
+		},
+		{
+			name:        "Should return future grants and features that have balance by time of querying even if no usage is registered",
```
turip, line 1135: Nice job on these testcases!

internal/testutils/pg_driver.go
```diff
@@ -0,0 +1,75 @@
...
+
+	// Dagger will set the POSTGRES_HOST environment variable for `make test`.
+	// If you need to run credit tests without Dagger you can set the POSTGRES_HOST environment variable.
+	// For example to use the Postgres in docker compose you can run `POSTGRES_HOST=localhost go test ./internal/credit/...`
+	host := os.Getenv("POSTGRES_HOST")
+	if host == "" {
+		t.Skip("POSTGRES_HOST not set")
+	}
+
+	return entsql.OpenDB(dialect.Postgres, pgtestdb.New(t, pgtestdb.Config{
```
turip, line 67: <3

internal/credit/postgres_connector/grant_test.go
```diff
@@ -40,12 +42,53 @@ func TestPostgresConnectorGrants(t *testing.T) {
 		description string
 		test        func(t *testing.T, connector credit.Connector, db_client *db.Client, ledger credit.Ledger)
 	}{
+		{
+			name:        "Should truncate grant effectiveAt to windowsize",
+			description: "EffectiveAt should be the end date of the window the input falls into",
+			test: func(t *testing.T, connector credit.Connector, db_client *db.Client, ledger credit.Ledger) {
+				ctx := context.Background()
```
turip, line 49: I would also add a testcase here if the effective time is already truncated. The code for truncate is good, just I would ensure that if we sometimes touch that codebase then we won't ruin the current logic.

internal/credit/postgres_connector/reset.go
```diff
@@ -20,6 +20,11 @@ type resetWithRollovedGrants struct {
 func (c *PostgresConnector) Reset(ctx context.Context, reset credit.Reset) (credit.Reset, []credit.Grant, error) {
 	ledgerID := credit.NewNamespacedLedgerID(reset.Namespace, reset.LedgerID)
 
+	// Truncate EffectiveAt to the window size
+	if effAt := reset.EffectiveAt.Truncate(c.config.WindowSize); effAt != reset.EffectiveAt {
```
turip, line 24: Given we have already calculated the truncated value we can just set it without the if.

```suggestion
	reset.EffectiveAt = reset.EffectiveAt.Truncate(c.config.WindowSize)
```



Also can you add a bit more verbose comment here why we are doing this? (e.g. clickhouse limitations, so that if somebody touches the code they have a better understanding).

internal/credit/postgres_connector/grant.go
```diff
@@ -24,6 +24,11 @@ func (c *PostgresConnector) CreateGrant(ctx context.Context, grantIn credit.Gran
 			return nil, err
 		}
 
+		// truncate effeectiveAt to start of next window
```
turip, line 27: Please add some comments detailing the issue why we need to ceil the timestamp here (this will not be obvious to others besides me and you).

internal/credit/postgres_connector/reset_test.go
```diff
@@ -34,14 +37,56 @@ func TestPostgresConnectorReset(t *testing.T) {
 	tt := []struct {
 		name        string
 		description string
-		test        func(t *testing.T, connector credit.Connector, streamingConnector *mockStreamingConnector, db_client *db.Client, ledger credit.Ledger)
+		test        func(t *testing.T, connector credit.Connector, streamingConnector *testutils.MockStreamingConnector, db_client *db.Client, ledger credit.Ledger)
 	}{
+		{
+			name:        "Shoul truncate effectiveAt to windowSize",
```
turip, line 43:
```suggestion
			name:        "Should truncate effectiveAt to windowSize",
```

internal/credit/postgres_connector/testutils/helpers.go
```diff
@@ -0,0 +1,62 @@
...
+
+func RemoveTimestampsFromFeatureBalances(featureBalances []credit.FeatureBalance) []credit.FeatureBalance {
+	for i := range featureBalances {
+		featureBalances[i].Feature.CreatedAt = nil
+		featureBalances[i].Feature.UpdatedAt = nil
+	}
+	return featureBalances
+}
+
+func ToPtr[D any](s D) *D {
```
turip, line 60: We already have this here: https://github.com/openmeterio/openmeter/blob/638d36c9b00ae77f8fc6604dd2a82135c12fef3f/pkg/convert/ptr.go#L3-L5

internal/credit/postgres_connector/grant_test.go
GAlexIHU: That happens in the default create testcase (there it's truncated)

internal/credit/postgres_connector/balance.go
GAlexIHU: I'm not sure I understand what you mean. There's a testcase `Should include usage between ledger creation and first grant`. What this bit does is burns down the first grant with that usage. Thinking about it:
- IMO this makes sense because the user would want to create ledgers and grants together (when first issuing), but it's reasonable to expect some delay between those operations, either through Cloud UI, or via API integration in selfhosted
- If they don't want to acknowledge that usage (e.g. create the ledger preemptively and later start issuing grants) they can always do reset for the ledger, then usage is only burnt down from the reset.

turip: What I meant is if at time T we have a
- Reset
- Grant for feature1 in the same window as T is in
- Grant for feature2 in the same window as T is in

We wouldn't fetch the meter for one of the features.

GAlexIHU: goccha, fixed it

turip (approved):

GAlexIHU:
here's the suggested migration:
```sql
UPDATE public.credit_entries
SET 
    effective_at = CASE 
                      WHEN entry_type IN ('GRANT', 'VOID_GRANT') THEN DATE_TRUNC('minute', effective_at + INTERVAL '1 minute' - INTERVAL '1 second')
                      ELSE DATE_TRUNC('minute', effective_at)
                   END,
    expiration_at = CASE 
                      WHEN entry_type IN ('GRANT', 'VOID_GRANT') THEN DATE_TRUNC('minute', expiration_at + INTERVAL '1 minute' - INTERVAL '1 second')
                      ELSE DATE_TRUNC('minute', expiration_at)
                    END;
```

you can test it here:
https://www.db-fiddle.com/f/8SqudRGzJYTrpaqhQHAA6A/0

PR #977, opened 2024-06-06
fix(sink-worker): prevent closing during flush operation

chrisgacsal:
## Overview

Prevent invoking `sink.Close()` from stopping `sink-worker` while `flush` operation is being performed in order to avoid events in buffer being processed twice due to inconsistent state where events data is already stored in storage (Clickhouse), but not in Kafka and/or Redis.

Use mutex to make `sink.Close()` to detect if `flush` operation is undergoing and delay shutdown until it is finished.
This change does not solve the issue where error(s) occurs during the `flush` operation which might result in data being inconsistent state.

Additional safeguards are going to be added to the `sink-worker` in followup PRs.

### DIFF
--- a/internal/sink/sink.go
+++ b/internal/sink/sink.go
@@ -290,12 +290,12 @@ func (s *Sink) persistToStorage(ctx context.Context, messages []SinkMessage) err
 	// Storage Batch insert
 	if len(batch) > 0 {
 		storageCtx, storageSpan := s.config.Tracer.Start(persistCtx, "storage-batch-insert")
+		defer storageSpan.End()
 		err := s.config.Storage.BatchInsert(storageCtx, batch)
 		if err != nil {
 			// Returning and error means we will retry the whole batch again
 			storageSpan.SetStatus(codes.Error, "failure")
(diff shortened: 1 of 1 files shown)


GAlexIHU (dismissed):

internal/sink/sink.go
```diff
@@ -640,7 +654,16 @@ func (s *Sink) ParseMessage(e *kafka.Message) (string, *serializer.CloudEventsKa
 }
 
 func (s *Sink) Close() error {
-	s.config.Logger.Info("closing sink")
+	logger := s.config.Logger.With("operation", "close")
+
+	logger.Info("closing sink")
+
+	logger.Debug("acquiring lock to prevent closing sink during flush operation")
```
hekike, line 661: Please add comment while this is necessary

chrisgacsal: Updated the PR with comments where locking is used. Ty

hekike (approved):

tothandras (approved):

sagikazarmark (approved):

PR #992, opened 2024-06-10
chore: update oapi to v2.3 OM-648

GAlexIHU:
## Overview

Updates oapi codegen to v2.3 and removes version pinning for kin-openapi dependency

Fixes #(issue)


GAlexIHU:
Updating above 2.2 is non trivial as the underlying kin-openapi changes make versions below 2.3 partially incompatible with versions after 2.3, and we cannot use latest kin-openapi versions due to our route validations in [cloud](https://github.com/openmeterio/cloud/pull/843).

PR #993, opened 2024-06-10
fix: disallow broken unique count meteraggregation for features

GAlexIHU:
## Overview

Disallowa UNQUE_COUNT for features until it's properly implemented.

Fixes #(issue)


turip (approved):

turip:
We already have this logic in place here: https://github.com/openmeterio/openmeter/blob/5c19d270be649601d593a5c81e39720fc0ad003f/internal/productcatalog/feature_connector.go#L71-L75

PR #1000, opened 2024-06-10
refactor(api): rewrite entitlements api OM-649

GAlexIHU:
## Overview

Contents:
- Write new entitlements APIs
- Breaks up old `credit` package into:
  - `credit`: calculation and core logics + entity management
  - `entitlement`: entitlements APIs (+leveraging credit)
  - `productcatalog`: currently only features
- Balance is now calculated "in-memory". We don't use ledgers and instantiated grants, everything is calculated.  
  - For performance reasons partial calculations are stored, these are called `snapshots` internally.
    - TODO: the way this is managed is not ideal, should be async later on
- Some changes in naming conventions (most things are still `connectors`, the suggestion is that their driven implementation is called `adapter` and the driving implementation is `driver`)
- Transaction management across ent instances (via custom tempaltes)
- Introduce some general concepts
  - IDs
  - entity generics (on DA lvl)
- Some global / general testutils
  - around ent/db management
  - for streaming mocking
- Misc:
  - new logging mode `prettydev` for local development
  - change meter `QueryParam` validation to make it more generic
  - some array/slice helpers

Example & proposal for new entitlements and balance APIs (v1).

You can view the up to date version in [swagger](https://editor.swagger.io/?url=https://raw.githubusercontent.com/openmeterio/openmeter/refactor/entitlements/api/openapi.yaml)

## Notes for reviewer

- the go client generation is broken, thats why the e2e tests are commented. it just doesnt support scenarios like allOf, anyOf, oneOf...... :/ 
- the API layer is kindofa mess, should be cleaned up (maybe in the followup PR)


api/openapi.yaml
```diff
@@ -1442,52 +1550,32 @@ components:
           type: string
           format: date-time
           example: "2023-01-01T00:00:00Z"
-    LedgerEntryType:
+    RecurringPeriod:
```
hekike, line 1553: This needs to have more of a reference as well. 
Like every day at 00:00 UTC. Or do we expect it's always startOf ?

api/openapi.yaml
```diff
@@ -1216,7 +1359,7 @@ components:
             tokens: "1234"
             model: "gpt-4-turbo"
         validationError: "meter not found for event"
-    CreateFeatureRequest:
+    FeatureCreateInputs:
```
hekike, line 1362: Let's add a required `key` field. This will help to work with features on UI.

api/openapi.yaml
```diff
@@ -1600,11 +1636,19 @@ components:
...
           example: "2023-01-01T00:00:00Z"
         expiration:
-          writeOnly: true
           allOf:
-            - $ref: "#/components/schemas/LedgerGrantExpirationPeriod"
+            - $ref: "#/components/schemas/ExpirationPeriod"
         rollover:
-          $ref: "#/components/schemas/LedgerGrantRollover"
+          anyOf:
+            - type: boolean
```
hekike, line 1643: Is this a remaining amount rollover? Description would be helpful

api/openapi.yaml
```diff
@@ -738,335 +738,478 @@ paths:
...
         "401":
           $ref: "#/components/responses/UnauthorizedProblemResponse"
+        "501":
+          $ref: "#/components/responses/NotImplementedProblemResponse"
         default:
           $ref: "#/components/responses/UnexpectedProblemResponse"
-    post:
-      operationId: createLedgerGrant
-      summary: Create a grant on a specific ledger.
+  /api/v1/subjects/{subjectIdOrKey}/entitlements/{entitlementId}/balance:
```
hekike, line 1009: Thinking loudly: if we have one entitlement per feature, we can allow `entitlementIdOrFeatureKey` param. Which again helps to check balance on UI side.

api/openapi.yaml
```diff
@@ -1256,172 +1399,137 @@ components:
...
-          maximum: 599
-          example: 400
-        detail:
+            The feature to grant.
           type: string
-          description: A human-readable explanation specific to this occurrence of the problem.
-          example: "body must be a JSON object"
-        instance:
+          example: 01ARZ3NDEKTSV4RRFFQ69G5FAV
+        measureUsageFrom:
```
hekike, line 1417: Make it optional, default to `createdAt`.

hekike, line 1417: As it's related to period I'd call it `periodAnchor`

api/openapi.yaml
```diff
@@ -738,335 +738,478 @@ paths:
...
         "401":
           $ref: "#/components/responses/UnauthorizedProblemResponse"
         "404":
           $ref: "#/components/responses/NotFoundProblemResponse"
         default:
           $ref: "#/components/responses/UnexpectedProblemResponse"
-    delete:
-      operationId: voidLedgerGrant
-      summary: Void ledger grant
+  /api/v1/subjects/{subjectIdOrKey}/entitlements/{entitlementId}/reset:
```
hekike, line 1136: This is updating the `periodAnchor`. Wondering if this need to be a separate endpoint or it's an edit on the entitlement

api/openapi.yaml
```diff
@@ -738,335 +738,478 @@ paths:
...
-          description: LedgerGrant created.
+        "200":
+          description: The entitlement balance.
           content:
             application/json:
               schema:
-                $ref: "#/components/schemas/LedgerGrantResponse"
+                type: object
+                properties:
+                  balance:
```
hekike, line 1037: Let's also add an `overage: number`.

api/openapi.yaml
```diff
@@ -1256,172 +1399,137 @@ components:
...
-        A ledger represented in open meter. A ledger must be assigned to a single
-        subject.
+        Entitles a subject to use a feature.
       required:
-        - subject
+        - featureId
+      additionalProperties: false
       properties:
-        subject:
+        featureId:
```
hekike, line 1412: we should consider adding a `limitType: 'soft' | 'hard'` so later in SDKs we can decide behavior when balance=0

api/openapi.yaml
GAlexIHU: We might have oversimplified with "fix 12am on X day", we can also introduce an anchor time as you suggested.

api/openapi.yaml
GAlexIHU: I think the idea was to push till later that decision

api/openapi.yaml
GAlexIHU: The idea here was actually different, it was so we can have a "different first period" (longer shorter)

api/openapi.yaml
GAlexIHU: `usage` - `balance` = overage

api/openapi.yaml
```diff
@@ -1442,144 +1695,67 @@ components:
...
         - effectiveAt
-        - expiration
+        - expiration # Does it have to expire?
       properties:
-        type:
-          $ref: "#/components/schemas/LedgerGrantType"
         amount:
           description: |
-            The amount to grant. Can be positive or negative number.
+            The amount to grant. Should be a positive number.
```
hekike, line 1755: /s The amount to grant. Must be a positive decimal number.

api/openapi.yaml
```diff
@@ -1600,111 +1776,106 @@ components:
...
-        - type
+        - hasAccess
       properties:
-        type:
-          $ref: "#/components/schemas/LedgerGrantRolloverType"
-        maxAmount:
-          description: |
-            Maximum amount to rollover.
+        hasAccess:
+          description: Whether the subject has access to the feature.
```
hekike, line 1848: Extend:

> Whether the subject has access to the feature.
> Always true for static and boolean entitlements.
> True for metered entitlements with non-zero balance or when the entitlement is unlimited.

api/openapi.yaml
```diff
@@ -1600,111 +1776,106 @@ components:
...
+          format: double
+          example: 100
+        usage:
+          description: Total usage of the feature in the period. Includes overages.
+          readOnly: true
+          type: number
+          format: double
+          example: 50
+        overage:
+          description: The overage of a metered entitlement.
```
hekike, line 1865: > The overage of a metered entitlement.
> Usage not covered by the grants.

api/openapi.yaml
```diff
@@ -1600,111 +1776,106 @@ components:
...
-      type: string
-      enum:
-        - REMAINING_AMOUNT
-        - ORIGINAL_AMOUNT
-      example: ORIGINAL_AMOUNT
-    LedgerGrantExpirationPeriod:
+          format: double
+          example: 100
+        usage:
+          description: Total usage of the feature in the period. Includes overages.
```
hekike, line 1859: > The total usage of a metered entitlement for the entitlement period.
> This is the full usage, including overage not covered by grants.

cmd/server/main.go
```diff
@@ -533,3 +565,56 @@ func initNamespace(config config.Configuration, namespaces ...namespace.Handler)
...
+func initPGClients(ctx context.Context, config config.PostgresConfig) (*struct {
+	driver                 *entDialectSQL.Driver
+	entitlementDBClient    *entitlementdb.Client
+	productcatalogDBClient *productcatalogdb.Client
+	creditDBClient         *creditdb.Client
+}, error) {
+	if err := config.Validate(); err != nil {
+		return nil, fmt.Errorf("invalid postgres config: %w", err)
+	}
+	driver, err := entutils.GetPGDriver(config.URL)
```
hekike, line 578: Is this going to open connection pools per connectors or share the connection pool between connectors?

e2e/e2e_test.go
```diff
@@ -33,9 +36,13 @@ func initClient(t *testing.T) *api.ClientWithResponses {
 }
 
 func TestMain(m *testing.M) {
-	// Make sure OpenMeter is ready
-	// TODO: replace this with some sort of health check
-	time.Sleep(15 * time.Second)
+	wait := os.Getenv("TEST_WAIT_ON_START")
+
+	if b, err := strconv.ParseBool(wait); err == nil && b {
```
hekike, line 41: you can workaround this to put a `testutils.EventuallyWithTf` before that calls healthcheck or queries meter

e2e/e2e_test.go
```diff
@@ -544,246 +559,245 @@ func TestCredit(t *testing.T) {
...
-	var ledgerID credit.LedgerID
-	ledgerMeta := map[string]string{
-		"test": "data",
-	}
-	t.Run("Create Ledger", func(t *testing.T) {
-		resp, err := client.CreateLedgerWithResponse(context.Background(), api.CreateLedgerJSONRequestBody{
-			Subject:  subject,
-			Metadata: ledgerMeta,
-		})
+	// FIXME: oapi-codegen is broken
```
hekike, line 562: I assume this is required for PR

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/AppsFlyer/go-sundheit v0.5.0
 	github.com/ClickHouse/clickhouse-go/v2 v2.25.0
 	github.com/XSAM/otelsql v0.31.0
+	github.com/alpacahq/alpacadecimal v0.0.4
```
hekike, line 12: good neighbors :)

internal/productcatalog/feature_connector.go
```diff
@@ -0,0 +1,151 @@
...
+	Offset          int
+	Limit           int
+	OrderBy         FeatureOrderBy
+}
+
+type DBCreateFeatureInputs struct {
+	Name                string
+	Key                 string
+	Namespace           string
+	MeterSlug           string
```
hekike, line 48: Shouldn't this be a pointer as meter is optional?

internal/productcatalog/feature_connector.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+}
+
+func (c *featureConnector) CreateFeature(ctx context.Context, feature CreateFeatureInputs) (Feature, error) {
+	meter, err := c.meterRepo.GetMeterByIDOrSlug(ctx, feature.Namespace, feature.MeterSlug)
+	if err != nil {
+		return Feature{}, &models.MeterNotFoundError{MeterSlug: feature.MeterSlug}
+	}
+
+	validAggregations := []models.MeterAggregation{
```
hekike, line 84: can be a const top of file

pkg/framework/entutils/testutils/ent1/db/example1.go
hekike, line 1: what are these example1,2,3,4... ?

internal/credit/balance_connector.go
```diff
@@ -0,0 +1,428 @@
...
+		logger: log,
+
+		// TODO: make configurable
+		snapshotGracePeriod: time.Hour,
+	}
+}
+
+type balanceConnector struct {
+	// grants and balance snapshots are managed in this same package
+	gc  GrantDBConnector
```
hekike, line 48: Let's write these out. gc to me is garbage collector

internal/productcatalog/postgresadapter/feature.go
```diff
@@ -0,0 +1,164 @@
...
+		Save(ctx)
+
+	if err != nil {
+		return productcatalog.Feature{}, err
+	}
+
+	return mapFeatureEntity(entity), nil
+}
+
+func (c *featureDBAdapter) FindByKey(ctx context.Context, namespace string, key string, includeArchived bool) (*productcatalog.Feature, error) {
```
hekike, line 49: Get by id and get by key can be merged. It usually helps to handle featureIdOrKey usecases. Like:

```
func (r Repository) GetByIdOrSlug(ctx context.Context, orgId string, idOrSlug string) (*db.Subject, error) {
	return r.ent.Subject.Query().
		WithMeter().
		Where(
                         Where(db_feature.Namespace(namespace)),
			entreport.Or(
				entreport.ID(idOrSlug),
				entreport.Slug(idOrSlug),
			),
		).
		Only(ctx)
}
```

internal/server/router/product_catalog.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"net/http"
+
+	"github.com/openmeterio/openmeter/api"
+)
+
+// List features
+// (GET /api/v1/features)
+func (a *Router) ListFeatures(w http.ResponseWriter, r *http.Request, params api.ListFeaturesParams) {
+	if !a.config.EntitlementsEnabled {
```
hekike, line 12: why not pass a NoopProductCatalogConnector in main instead?

internal/streaming/testutils/streaming.go
```diff
@@ -12,6 +12,7 @@ import (
 )
 
 type MockStreamingConnectorParams struct {
+	// TODO: get rid of this, it's not used anymore
 	DefaultHighwatermark time.Time
```
hekike, line 16: is it still used?

pkg/models/error.go
```diff
@@ -17,3 +17,12 @@ type NamespaceNotFoundError struct {
 func (e *NamespaceNotFoundError) Error() string {
 	return fmt.Sprintf("namespace not found: %s", e.Namespace)
 }
+
+// TODO: this should be picked up in a general server error handler
+type GenericUserError struct {
```
hekike, line 22: when should we use this?

internal/credit/engine_test.go
```diff
@@ -0,0 +1,972 @@
...
+
+		ID:          "grant-1",
+		Amount:      100.0,
+		Priority:    1,
+		EffectiveAt: t1,
+		Expiration: credit.ExpirationPeriod{
+			Duration: credit.ExpirationPeriodDurationDay,
+			Count:    30,
+		},
+		// ExpiresAt: time.Now().AddDate(0, 0, 30),
```
hekike, line 35: why is it commented out?

internal/credit/owner_connector.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/openmeterio/openmeter/internal/streaming"
+	"github.com/openmeterio/openmeter/pkg/framework/entutils"
+)
+
+type OwnerConnector interface {
```
hekike, line 12: who is the "owner"?

internal/credit/postgresadapter/ent/schema/balance_snapshot.go
```diff
@@ -0,0 +1,56 @@
...
+	return []ent.Mixin{
+		entutils.NamespaceMixin{},
+		entutils.TimeMixin{}, // maybe we don't need this
+	}
+}
+
+// Fields of the CreditGrant.
+func (BalanceSnapshot) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("owner_id").GoType(credit.GrantOwner("")).Immutable().SchemaType(map[string]string{
```
hekike, line 28: why is it owner? is it a subject?

internal/entitlement/balance_connector.go
```diff
@@ -0,0 +1,304 @@
...
+	nsOwner := credit.NamespacedGrantOwner{
+		Namespace: entitlementID.Namespace,
+		ID:        credit.GrantOwner(entitlementID.ID),
+	}
+	res, err := e.bc.GetBalanceOfOwner(ctx, nsOwner, at)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get balance of entitlement %s: %w", entitlementID.ID, err)
+	}
+
+	meterSlug, params, err := e.oc.GetOwnerQueryParams(ctx, nsOwner)
```
turip, line 89: If I understand correctly, during `e.bc.GetBalanceOfOwner`, we are relying on the meter data to calculate grant burndown and overage. Thus, we have the usage data there, too.

Question: given that we already have that data queried there, can't we just return the usage as part of the balance information?

(This would allow this connector to be a really thin one, as I think the history has the same issue).

internal/entitlement/entitlemeng_grant.go
```diff
@@ -0,0 +1,39 @@
+package entitlement
```
turip, line 1: typo in filename

internal/productcatalog/feature.go
```diff
@@ -64,8 +70,8 @@ type Feature struct {
 	MeterGroupByFilters *map[string]string `json:"meterGroupByFilters,omitempty"`
 
 	// Read-only fields
-	Archived *bool `json:"archived,omitempty"`
+	ArchivedAt *time.Time `json:"archivedAt,omitempty"`
```
turip, line 73: <3

turip (approved):

api/openapi.yaml
```diff
@@ -738,335 +773,454 @@ paths:
...
             enum:
-              - subject
-              - createdAt
               - id
-            default: id
+              - createdAt
+              - updatedAt
+            default: updatedAt
+        - $ref: "#/components/parameters/includeDeleted"
+        # - $ref: "#/components/parameters/queryFilterSubject"
```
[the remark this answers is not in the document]
GAlexIHU: its currently not implemented

api/openapi.yaml
```diff
@@ -738,335 +773,454 @@ paths:
...
-  /api/v1/ledgers:
+  /api/v1/grants:
     get:
-      operationId: listLedgers
-      summary: List the already defined ledgers.
+      operationId: listGrants
+      summary: List grants
       description: |
-        List the already defined ledgers.
+        List all grants.
```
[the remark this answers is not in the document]
GAlexIHU: we dont have any filters

cmd/server/main.go
GAlexIHU: `*sql.DB` and `*driver.Connector` manage pooling, thats part of driver, so on, its a single pool

e2e/e2e_test.go
GAlexIHU: tbh i'd be happy to hear suggestions... 
TLDR for those not familiar: `oapi-codegen` cannot resolve complex types (like `oneOf`, `anyOf` ...). we can monkeypatch-it by manually typing out the types (which i disdain) or we can give up our aspirations of using those types (that i also disdain)... thoughts?

internal/productcatalog/feature_connector.go
GAlexIHU: well yes, but i'd push that to next PR, currently it isn't (as it wasn't previously)

internal/productcatalog/feature_connector.go
GAlexIHU: IMO should be config to the connector, feature doesn't implement any querying, it doesn't really know what aggregation types it can support....

pkg/framework/entutils/testutils/ent1/db/example1.go
GAlexIHU: used in the transaction tests (in `entutils`), there are two, example1 and example2

internal/server/router/product_catalog.go
GAlexIHU: because we don't have a NoopProductCatalogConnector and i'm not convinced we need one, we should figure out conditional module loading instead IMO

internal/streaming/testutils/streaming.go
GAlexIHU: removed

pkg/models/error.go
GAlexIHU: I just added it for cases where I didn't want to type out a domain specific error

internal/credit/engine_test.go
GAlexIHU: was left there, deleted

internal/credit/owner_connector.go
GAlexIHU: whoever owns the grant, currently only entitlements. the idea was that in the future if we want e.g. feature specific grants, or a general credit system, we won't have to change a thing just reimplement this interface.

internal/credit/postgresadapter/ent/schema/balance_snapshot.go
GAlexIHU: for current usecases it transaltes to an entitlement, see the comment above for details

api/openapi.yaml
```diff
@@ -1216,24 +1370,35 @@ components:
...
       description: |
         A feature is a feature or service offered to a customer.
         For example: CPU-Hours, Tokens, API Calls, etc.
       required:
+        - key
         - name
-        - meterSlug
-      additionalProperties: false
       properties:
+        key:
```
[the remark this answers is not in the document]
GAlexIHU: we have `subjectKey` and `meterSlug` as of now, they are a bit different. Key is an external reference and slug is a regex validated string. IMO this is also closer to slug as the point is for it to be a human readable id.

internal/entitlement/postgresadapter/ent/schema/entitlement.go
```diff
@@ -0,0 +1,54 @@
...
+func (Entitlement) Mixin() []ent.Mixin {
+	return []ent.Mixin{
+		entutils.IDMixin{},
+		entutils.NamespaceMixin{},
+		entutils.MetadataAnnotationsMixin{},
+		entutils.TimeMixin{},
+	}
+}
+
+// Fields of the CreditGrant.
```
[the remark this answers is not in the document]
GAlexIHU: fixed, thx

internal/entitlement/grant_owner_adapter.go
```diff
@@ -0,0 +1,164 @@
...
+
+type entitlementGrantOwner struct {
+	fdb    productcatalog.FeatureDBConnector
+	edb    EntitlementDBConnector
+	urdb   UsageResetDBConnector
+	mr     meter.Repository
+	logger *slog.Logger
+}
+
+func NewEntitlementGrantOwnerAdapter(
```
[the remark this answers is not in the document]
GAlexIHU: some docs for terminology definitely would be great...

The idea is fairly simple: _a grant has an owner: it's what we grant to_ 

Currently we grant to entitlements, so owner always translates to an entitlement. In the future we might want to do grants, balance, credits... for a feature directly, or maybe implement a generic credit system with a balance. For those cases the `Owner` would be different. The benefit is that we won't have to change a thing just implement the `OwnerConnector` interface for those as well.

internal/credit/balance_connector.go
```diff
@@ -0,0 +1,428 @@
...
+	}
+	if balance.At.Before(periodStart) {
+		// This is an inconsistency check. It can only happen if we lost our snapshot for the last reset.
+		//
+		// The engine doesn't manage rollovers at usage reset so it cannot be used to calculate GrantBurnDown across resets.
+		return nil, fmt.Errorf("last valid balance snapshot %s is before current period start at %s, no snapshot was created for reset", balance.At, periodStart)
+	}
+
+	// get all relevant grants
+	grants, err := m.gc.ListActiveGrantsBetween(ctx, owner, balance.At, at)
```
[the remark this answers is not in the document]
GAlexIHU: you mean this part?
```go
// get all relevant grants
grants, err := m.gc.ListActiveGrantsBetween(ctx, owner, period.From, period.To)
// these grants might not be present in the starting balance so lets fill them
for _, grant := range grants {
	if _, ok := balance.Balances[grant.ID]; !ok {
		if grant.ActiveAt(period.From) {
			// This is only possible in case the grant becomes active exactly at the start of the current period
			balance.Balances.Set(grant.ID, grant.Amount)
		} else {
			balance.Balances.Set(grant.ID, 0.0)
		}
	}
}
```

internal/credit/balance_connector.go
```diff
@@ -0,0 +1,428 @@
...
+		if _, ok := err.(*OwnerNotFoundError); ok {
+			return nil, err
+		}
+		return nil, fmt.Errorf("failed to get current usage period start for owner %s at %s: %w", owner.ID, at, err)
+	}
+	if at.Before(periodStart) {
+		return nil, &models.GenericUserError{Message: fmt.Sprintf("reset at %s is before current usage period start %s", at, periodStart)}
+	}
+
+	balance, err := m.getLastValidBalanceSnapshotForOwnerAt(ctx, owner, at)
```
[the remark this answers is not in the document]
GAlexIHU: the link for me opens the entire diff, could please you re-comment or quote? :D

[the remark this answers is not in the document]
GAlexIHU: good catch!
added breaking test + fix

internal/entitlement/balance_connector.go
GAlexIHU: we can't: usage is measured for the whole period while the Engine only calculates since the last snapshot

internal/entitlement/grant_owner_adapter.go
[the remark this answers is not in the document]
turip: Can we name this parent or parentEntity something? What owner means for me (but I might be alone with this) is the actual person owning something ;-)

internal/credit/balance_connector.go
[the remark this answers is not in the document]
turip: yes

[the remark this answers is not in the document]
GAlexIHU: deduped, see elsewhere

internal/credit/balancesnapshot_connector.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/openmeterio/openmeter/pkg/framework/entutils"
+)
+
+type BalanceSnapshotDBConnector interface {
```
sagikazarmark, line 11: Is this...a repository?

internal/credit/grant_connector.go
```diff
@@ -0,0 +1,199 @@
...
+	Amount           float64
+	Priority         uint8
+	EffectiveAt      time.Time
+	Expiration       ExpirationPeriod
+	Metadata         map[string]string
+	ResetMaxRollover float64
+	Recurrence       *Recurrence
+}
+
+type GrantConnector interface {
```
sagikazarmark, line 22: Is this a repository?

internal/credit/grant_connector.go
```diff
@@ -0,0 +1,199 @@
...
+type ListGrantsParams struct {
+	Namespace      string
+	OwnerID        *GrantOwner
+	IncludeDeleted bool
+	Offset         int
+	Limit          int
+	OrderBy        GrantOrderBy
+}
+
+type DBCreateGrantInput struct {
```
sagikazarmark, line 49: Not sure I like this name? How about something like `CreateGrant`?

internal/credit/postgresadapter/balance_snapshot.go
```diff
@@ -0,0 +1,75 @@
+package postgresadapter
```
sagikazarmark, line 1: I generally prefer a single adapter package (ie. `creditadapter`).

internal/credit/postgresadapter/transaction.go
```diff
@@ -0,0 +1,44 @@
...
+
+	"github.com/openmeterio/openmeter/internal/credit"
+	"github.com/openmeterio/openmeter/internal/credit/postgresadapter/ent/db"
+	"github.com/openmeterio/openmeter/pkg/framework/entutils"
+)
+
+// We implement entuitls.TxUser[T] and entuitls.TxCreator here
+// There ought to be a better way....
+
+func (e *grantDBADapter) Tx(ctx context.Context) (context.Context, *entutils.TxDriver, error) {
```
sagikazarmark, line 16: You could return a generic transaction interface to decouple it from ent.

sagikazarmark (commented):
General feedback:

- I'm more in favor of calling repositories repositories, not connectors (those are already present here and in other codebases)
- I prefer a single adapter package for each domain, unless there is a very good reason to do otherwise.
- I prefer keeping storage layer details out of type names (eg. no `DBInput` and `DBConnector` type names)

GAlexIHU:
> General feedback:
> 
> * I'm more in favor of calling repositories repositories, not connectors (those are already present here and in other codebases)

I'm okay with repos going on

> * I prefer a single adapter package for each domain, unless there is a very good reason to do otherwise.

I prefer adapter/functionality, these packages do a lot and the adapters are already fairly long

> * I prefer keeping storage layer details out of type names (eg. no `DBInput` and `DBConnector` type names)

Calling them Repo solves it, but due to name collisions  we have to differentiate in some cases

internal/credit/postgresadapter/ent/schema/balance_snapshot.go
hekike: The problem is that this way we loose the foreign key.
I'm wondering that evenn if connector is generic PG adapter should be specific to use-case.

internal/credit/expiration.go
```diff
@@ -0,0 +1,53 @@
...
+	// Duration The expiration period duration like month.
+	Duration ExpirationPeriodDuration `json:"duration,omitempty"`
+}
+
+func (c ExpirationPeriod) GetExpiration(t time.Time) time.Time {
+	switch c.Duration {
+	case ExpirationPeriodDurationHour:
+		return t.Add(time.Hour * time.Duration(c.Count))
+	case ExpirationPeriodDurationDay:
+		return t.AddDate(0, 0, int(c.Count))
```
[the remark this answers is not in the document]
hekike: I'd enforce UTC everywhere. In cloud we have an account level TZ, potentially we can bring that over but we still need to align them somehiw.

sagikazarmark:
> > * I prefer a single adapter package for each domain, unless there is a very good reason to do otherwise.
> 
> I prefer adapter/functionality, these packages do a lot and the adapters are already fairly long

This is what we followed so far (except from a few cases), so let's not break pattern.

> Calling them Repo solves it, but due to name collisions we have to differentiate in some cases

That's fine, just don't call it `DB`. Call it `Model` or `Entity` or something to that effect.

internal/credit/postgresadapter/ent/schema/balance_snapshot.go
GAlexIHU: thats not a bad option, we can do that 👍  
i'm personally not that worried about missing the FKeys, I thought about it a bit when deciding to do it this way and couldn't really find a scenario where it would bite us

internal/productcatalog/feature_connector.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/openmeterio/openmeter/pkg/framework/entutils"
+	"github.com/openmeterio/openmeter/pkg/models"
+)
+
+type CreateFeatureInputs struct {
+	Name                string             `json:"name"`
+	Key                 string             `json:"key"`
+	Namespace           string             `json:"namespace"`
+	MeterSlug           string             `json:"meterSlug"`
+	MeterGroupByFilters *map[string]string `json:"meterGroupByFilters"`
```
turip, line 17: Map itself is a pointer. If you just remove the star most of the code will just work fine. I have seen that oapi codegen does the same thing, but still the pointer is unnecessary.

internal/productcatalog/feature_connector.go
turip: Putting aside the theoretical part, usually these are defined as globals as @hekike suggested.

api/openapi.yaml
```diff
@@ -738,335 +773,454 @@ paths:
...
           $ref: "#/components/responses/UnexpectedProblemResponse"
+  /api/v1/subjects/{subjectIdOrKey}/entitlements:
     post:
-      operationId: createLedger
-      summary: Creates the specified ledger
+      operationId: createEntitlement
+      summary: Create entitlement
       description: |
-        Create or update the specified ledger.
+        Create an entitlement for a subject.
```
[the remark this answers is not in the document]
GAlexIHU: will do as part of general documentation task

api/openapi.yaml
GAlexIHU: will rewrite api descriptions as part of documentation effort

e2e/e2e_test.go
GAlexIHU: fixed the labour-intensive way

api/openapi.yaml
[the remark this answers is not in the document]
GAlexIHU: had a discussion on it today with @hekike & @tothandras and the decision was to stick with key

internal/productcatalog/postgresadapter/feature.go
GAlexIHU: i think in general having two unique identifiers being used is confusing and a bad design. if we use keys then we could / should manage all user initiated logic with keys IMO (that would make IDs obsolete for the most part), or at least be more intentional about when to use which and why.

hekike (dismissed):

internal/credit/engine.go
```diff
@@ -0,0 +1,446 @@
...
+// 1. Grants with higher priority are burned down first
+// 2. Grants with earlier expiration date are burned down first
+func prioritizeGrants(grants []Grant) error {
+	if len(grants) == 0 {
+		// we don't do a thing, return early
+		// return fmt.Errorf("no grants to prioritize")
+		return nil
+	}
+
+	// 2. Grants with earlier expiration date are burned down first
```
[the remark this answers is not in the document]
GAlexIHU: good catch i wasnt aware

so actually there's a test for it which passes "burn down grant that expires fitst". Also, i cannot produce a breaking test with the old setup...

https://github.com/openmeterio/openmeter/blob/refactor/entitlements/internal/credit/engine_test.go#L422

https://github.com/openmeterio/openmeter/blob/refactor/entitlements/internal/credit/engine_test.go#L454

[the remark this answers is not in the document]
turip: Yes, I don't think it's worth pursuing it more other than just using the Stable version so that we are not running into any corner case.

internal/productcatalog/feature_connector.go
```diff
@@ -0,0 +1,151 @@
...
+	found, err := c.db.FindByKey(ctx, feature.Namespace, feature.Name, false)
+	if err != nil {
+		if _, ok := err.(*FeatureNotFoundError); !ok {
+			return Feature{}, err
+		}
+	} else {
+		return Feature{}, &FeatureWithNameAlreadyExistsError{Name: feature.Name, ID: found.ID}
+	}
+
+	return c.db.CreateFeature(ctx, FeatureRepoCreateFeatureInputs(feature))
```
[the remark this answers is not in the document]
turip: Let's agree that we disagree: I don't have issues with the different types' existence, but until they diverge, I would rather just add a type alias to signify that they aren't necessarily the same.

api/openapi.yaml
```diff
@@ -1244,184 +1438,272 @@ components:
...
+          $ref: "#/components/schemas/RecurringPeriod"
+    EntitlementSharedFields:
+      allOf:
+        - $ref: "#/components/schemas/SharedMetaFields"
+        - $ref: "#/components/schemas/EntitlementCreateSharedFields"
+        - type: object
+          required:
+            - subjectKey
+          properties:
+            subjectkey:
```
turip, line 1485:
```suggestion
            subjectKey:
```

Given that https://github.com/openmeterio/openmeter/blob/0651acb69cb1be32ca44f2cf59d08fefcff72d74/internal/entitlement/entitlement.go#L14

Also we seem to miss [MeasureUsageFrom](https://github.com/openmeterio/openmeter/blob/0651acb69cb1be32ca44f2cf59d08fefcff72d74/internal/entitlement/entitlement.go#L13).

api/types/entitlement.go
```diff
@@ -0,0 +1,66 @@
...
+	Config *string `json:"config"`
+
+	// FeatureId The feature the subject is entitled to use
+	FeatureId string `json:"featureId"`
+
+	// IsUnlimited If unlimited=true the subject can use the feature an unlimited amount.
+	IsUnlimited *bool `json:"isUnlimited,omitempty"`
+
+	// IssueAfterReset You can issue usage automatically after reset. This usage is not rolled over.
+	IssueAfterReset *float32 `json:"issueAfterReset,omitempty"`
```
turip, line 16: Shouldn't this be float64?

Also I think SubjectKey is missing.

api/openapi.yaml
GAlexIHU: yes, not everything is or will be implemented in this PR, thanks for the typo

tothandras (approved):

PR #1011, opened 2024-06-15
refactor(sink-worker): add improvements

chrisgacsal:
## Overview

Refactor `sink-worker` adding tiny improvements/fixes:
* proper `context.Context` propagation
* properly close Kafka consumer on exit
* set exit code on error
* remove unnecessary signal handling in `sink.Sink`

### DIFF
--- a/cmd/sink-worker/main.go
+++ b/cmd/sink-worker/main.go
@@ -199,10 +199,12 @@ func main() {
 	group.Add(run.SignalHandler(ctx, syscall.SIGINT, syscall.SIGTERM))
 
 	err = group.Run()
+
 	if e := (run.SignalError{}); errors.As(err, &e) {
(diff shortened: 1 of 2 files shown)


cmd/sink-worker/main.go
```diff
@@ -162,41 +166,46 @@ func main() {
 		logger.Error("failed to initialize sink worker", "error", err)
 		os.Exit(1)
 	}
-
-	var group run.Group
+	defer func() {
+		_ = sink.Close()
```
hekike, line 170: shouldn't we log error?

hekike (approved):

cmd/sink-worker/main.go
chrisgacsal: Actually we do within `sink.Close()` function. I'll push a followup PR which will remove `error` retrun for `sink.Close()` as does not return error anyway.

PR #1042, opened 2024-06-21
feat(entitlements): implement different entitlement types

GAlexIHU:
## Overview

- Introduce the different entitlement types (boolean | static | metered)
- Implement meter-less features
- Store UsagePeriod fields
- Extract Entitlement Create input to schema to make client generation work
- Clean up type comments

Fixes #(issue)


internal/entitlement/connector.go
```diff
@@ -33,68 +33,98 @@ type EntitlementConnector interface {
...
+	entitlementRepo  EntitlementRepo
+	featureConnector productcatalog.FeatureConnector
 }
 
 func NewEntitlementConnector(
-	entitlementBalanceConnector EntitlementBalanceConnector,
 	entitlementRepo EntitlementRepo,
 	featureConnector productcatalog.FeatureConnector,
-) EntitlementConnector {
+	meteredEntitlementConnector SubTypeConnector,
```
turip, line 47: Most probably not for this PR, but this is getting out of hand, esp if we have to maintain this initialization sequence in multiple components. Can we have a single helper initialize the whole feature and return the connectors?

internal/entitlement/connector.go
```diff
@@ -33,68 +33,98 @@ type EntitlementConnector interface {
...
 }
 
 func (c *entitlementConnector) CreateEntitlement(ctx context.Context, input CreateEntitlementInputs) (Entitlement, error) {
-	// TODO: check if the feature exists, if it is compatible with the type, etc....
 	feature, err := c.featureConnector.GetFeature(ctx, models.NamespacedID{Namespace: input.Namespace, ID: input.FeatureID})
 	if err != nil {
 		return Entitlement{}, &productcatalog.FeatureNotFoundError{ID: input.FeatureID}
 	}
-	if feature.ArchivedAt != nil {
+	if feature.ArchivedAt != nil && feature.ArchivedAt.Before(time.Now()) {
```
turip, line 65: This is a fun use case; you can deprecate/archive a feature in the future.

internal/entitlement/entitlement.go
```diff
@@ -1,55 +1,109 @@
...
 
 import (
-	"fmt"
 	"time"
 
 	"github.com/openmeterio/openmeter/pkg/models"
+	"github.com/openmeterio/openmeter/pkg/slicesx"
 )
 
+type HasType interface {
```
turip, line 10:
```suggestion
type TypedEntitlement interface {
```

internal/entitlement/connector.go
```diff
@@ -33,68 +33,98 @@ type EntitlementConnector interface {
...
+	return connector.GetValue(ent, at)
 }
 
 func (c *entitlementConnector) ListEntitlements(ctx context.Context, params ListEntitlementsParams) ([]Entitlement, error) {
 	return c.entitlementRepo.ListEntitlements(ctx, params)
 }
+
+func (c *entitlementConnector) getTypeConnector(inp HasType) (SubTypeConnector, error) {
+	entitlementType := inp.GetType()
+	switch entitlementType {
```
turip, line 120:
```suggestion
	switch inp.GetType() {
```

internal/entitlement/static/static_entitlement.go
```diff
@@ -0,0 +1,26 @@
+package staticentitlement
```
turip, line 1: nit: I would name the file entitlement.go as it's obvious that it's the static one.

internal/entitlement/postgresadapter/ent/schema/entitlement.go
```diff
@@ -25,11 +26,19 @@ func (Entitlement) Mixin() []ent.Mixin {
...
+		field.Enum("entitlement_type").Values(entitlement.EntitlementType("").StrValues()...).Immutable(),
 		field.String("feature_id").Immutable().SchemaType(map[string]string{
 			dialect.Postgres: "char(26)",
 		}),
 		field.String("subject_key").Immutable(),
-		field.Time("measure_usage_from").Immutable(),
+		field.Time("measure_usage_from").Optional().Nillable().Immutable(),
+		field.Float("issue_after_reset").Optional().Nillable().Immutable(),
+		field.Bool("is_soft_limit").Optional().Nillable().Immutable(),
+		field.String("config").Optional().Nillable().SchemaType(map[string]string{
```
turip, line 37: Based on this comment: https://github.com/openmeterio/openmeter/blob/55fb3fcc153f7d1db926043d31ecd0ded26ad521/api/api.gen.go#L390

Config should a json object, but as far as I see:
- In the schema it's a text object
- In the code here https://github.com/openmeterio/openmeter/blob/96b5deeb7369c9e314fd5fb942da4d3c9fc87ff6/internal/entitlement/static/static_entitlement.go#L10

What's the idea here? If we expect it to be json, then we should validate it at some point, right?

internal/entitlement/httpdriver/parser.go
```diff
@@ -0,0 +1,115 @@
...
+		return nil, err
+	}
+
+	return &api.EntitlementMetered{
+		CreatedAt:       &metered.CreatedAt,
+		DeletedAt:       metered.DeletedAt,
+		FeatureId:       metered.FeatureID,
+		Id:              &metered.ID,
+		IsUnlimited:     convert.ToPointer(false), // implement
+		IssueAfterReset: metered.IssuesAfterReset,
```
turip, line 30: As far as I see it's not implemented just stored, right?

turip: Also if this is used to auto issue metered entitlements after auto resets we should have the same solution for string and bools, or what's the plan here? cc @hekike

internal/entitlement/connector.go
GAlexIHU: Technically yes, though in practice we always set it for now.

internal/entitlement/postgresadapter/ent/schema/entitlement.go
GAlexIHU: My thinking was I don't know what format we'll want it to be. The api description says it should be json. We'll probably be safer if we enforce it to be json for now. 
I can see a rational behind not limiting it to being json and I didn't want to really tackle that, that's why i pushed it off.
I'll just enforce it to be json.

internal/entitlement/httpdriver/parser.go
GAlexIHU: > As far as I see it's not implemented just stored, right?

yes, it will be in a followup
> Also if this is used to auto issue metered entitlements after auto resets we should have the same solution for 
string and bools, or what's the plan here?

wdym same solution for static and bools? what would that mean?

internal/entitlement/connector.go
GAlexIHU: we return it too thats why its separate

internal/entitlement/static/static_entitlement.go
GAlexIHU: i usually optimize names for my fuzzy finder but i suppose thats a unique concern to me (for that it finds it easier if it has this name)

Nevertheless i agree it would be nicer to call them all just entitlement.go, so i'll rename them

internal/entitlement/postgresadapter/ent/schema/entitlement.go
turip: I think this https://github.com/openmeterio/openmeter/pull/1042/files#diff-6c08e65eef0374988297d041bd50f87b385613a7e478c3812cb5312cc3e1f1e3R1915 should be an object typed member. It feels strange that we are embedding JSON string into a json response.

GAlexIHU: enforced to be json

internal/entitlement/connector.go
GAlexIHU: will squeeze in in one of the followups

turip (dismissed):

turip (approved):

PR #1065, opened 2024-06-25
Add support for using featureKey with entitlements

GAlexIHU:
## Overview

- adds support for using feature key when interacting with entitlements
- stores UsagePeriod

Fixes #(issue)


internal/entitlement/connector.go
```diff
@@ -57,50 +57,74 @@ func NewEntitlementConnector(
 	}
 }
 
-func (c *entitlementConnector) CreateEntitlement(ctx context.Context, input CreateEntitlementInputs) (Entitlement, error) {
-	feature, err := c.featureConnector.GetFeature(ctx, models.NamespacedID{Namespace: input.Namespace, ID: input.FeatureID})
-	if err != nil {
-		return Entitlement{}, &productcatalog.FeatureNotFoundError{ID: input.FeatureID}
+func (c *entitlementConnector) CreateEntitlement(ctx context.Context, input CreateEntitlementInputs) (*Entitlement, error) {
+	// ID has precedence over key
+	idOrKey := input.FeatureID
```
turip, line 62: I think the inputs should already get a single FeatureIDorKey field, and this should be handled in the httpdriver layer.

This would also align the PR with the GetEntitlementValue's signature

internal/entitlement/postgresadapter/ent/schema/entitlement.go
```diff
@@ -30,6 +30,7 @@ func (Entitlement) Fields() []ent.Field {
 		field.String("feature_id").Immutable().SchemaType(map[string]string{
 			dialect.Postgres: "char(26)",
 		}),
+		field.String("feature_key").Immutable(),
```
turip, line 33: I don't think it's worth storing it here. I would rather prefer that the connector resolves the ID and we persits that data.

I wouldn't want to allow for a situation where a feature gets archived, then a new feature is created with the same key. Let's make sure that inside the DB we are not representing this convenience feature.

internal/entitlement/postgresadapter/entitlement.go
```diff
@@ -159,6 +188,10 @@ func mapEntitlementEntity(e *db.Entitlement) *entitlement.Entitlement {
 	}
 
 	if e.UsagePeriodAnchor != nil && e.UsagePeriodInterval != nil {
+		ent.GenericProperties.UsagePeriod = &entitlement.UsagePeriod{
+			Anchor:   *convert.SafeToUTC(e.UsagePeriodAnchor),
```
turip, line 192:
```suggestion
			Anchor:   e.UsagePeriodAnchor.In(time.UTC)
```

nit: given that you already done the nil check, I wouldn't involve all the dereferring magic here.

internal/entitlement/postgresadapter/entitlement.go
```diff
@@ -159,6 +188,10 @@ func mapEntitlementEntity(e *db.Entitlement) *entitlement.Entitlement {
...
+			Interval: entitlement.UsagePeriodInterval(*e.UsagePeriodInterval),
+		}
 		ent.UsagePeriod = &entitlement.UsagePeriod{
```
turip, line 195: Why do we store it twice?

internal/entitlement/repository.go
```diff
@@ -21,3 +23,17 @@ type EntitlementRepo interface {
 	entutils.TxCreator
 	entutils.TxUser[EntitlementRepo]
 }
+
+type CreateEntitlementRepoInputs struct {
+	Namespace       string          `json:"namespace"`
+	FeatureID       string          `json:"featureId"`
+	FeatureKey      string          `json:"featureKey"`
```
turip, line 30: As above, I don't think this is a good idea, let's have stable references in the DB.

internal/productcatalog/feature_connector.go
```diff
@@ -118,13 +117,13 @@ func (c *featureConnector) ListFeatures(ctx context.Context, params ListFeatures
...
-	feature, err := c.featureRepo.GetByID(ctx, featureID)
+func (c *featureConnector) GetFeature(ctx context.Context, namespace string, idOrKey string) (*Feature, error) {
+	feature, err := c.featureRepo.GetByIdOrKey(ctx, namespace, idOrKey, true)
 	if err != nil {
-		return Feature{}, err
+		return nil, err
 	}
-	if feature.Namespace != featureID.Namespace {
-		return Feature{}, &FeatureNotFoundError{ID: featureID.ID}
+	if feature.Namespace != namespace {
```
turip, line 125: This is not needed as the GetByIdOrKey's idempotent property should be that the namespace is already filtered for. The previous implementation didn't have this check, so that's why it has been stuck here.

internal/productcatalog/feature_connector.go
```diff
@@ -118,13 +117,13 @@ func (c *featureConnector) ListFeatures(ctx context.Context, params ListFeatures
 	return c.featureRepo.ListFeatures(ctx, params)
 }
 
-func (c *featureConnector) GetFeature(ctx context.Context, featureID models.NamespacedID) (Feature, error) {
-	feature, err := c.featureRepo.GetByID(ctx, featureID)
+func (c *featureConnector) GetFeature(ctx context.Context, namespace string, idOrKey string) (*Feature, error) {
+	feature, err := c.featureRepo.GetByIdOrKey(ctx, namespace, idOrKey, true)
```
turip, line 121: Please add an options struct at least for the includeArchived as true/false const arguments are extremely hard to understand what they are representing. (or you can add a type IncludeArchive bool so that the next one reading this code won't have to look up the function signature) ;)

turip (changes requested):

internal/entitlement/postgresadapter/ent/schema/entitlement.go
GAlexIHU: So actually the problem has deeper roots than implementation details:

- If we want to allow entitlements to be accessible by featureKey, then it's necessary to limit that a user can only be entitled to a given featureKey once. Otherwise we cannot return a simple value.
- If we limit that then after archiving a feature the user can only be entitled again if the previous is deleted.

The issue is that this is not yet enforced (which i'll add) but after that these archivation scenarios are resolved. (this has some implications on the checks we do at create-time but nothing too complex)

internal/productcatalog/feature_connector.go
GAlexIHU: I agree 4 arguments is starting to get on the longer side, but signature wise a bool is no different than a string and we don't do custom types for that.

internal/entitlement/repository.go
GAlexIHU: see other thread

internal/entitlement/postgresadapter/entitlement.go
GAlexIHU: mishap, thx

GAlexIHU:
implemented in another PR

PR #1067, opened 2024-06-26
feat: entitlement auto reset support pt1

turip:
This patch ensures that for metered entitlements we have the next reset time captured. Also we are providing the required function to perform the which can be supported as a cron job in the OpenMeter solution.

## Overview

This patch:
- adds a new Nillable DB field to store the next reset's time
- this allows us to quickly detect which entitlements need the reset right now


turip:
Failures seem to be stemming from the upstream PR: https://github.com/openmeterio/openmeter/pull/1068/files

internal/entitlement/metered/grant_owner_adapter.go
```diff
@@ -121,50 +121,70 @@ func (e *entitlementGrantOwner) GetPeriodStartTimesBetween(ctx context.Context,
 	return times, nil
 }
 
-func (e *entitlementGrantOwner) EndCurrentUsagePeriod(ctx context.Context, owner credit.NamespacedGrantOwner, at time.Time) error {
```
GAlexIHU, line 124: thx for hygiene 👍

internal/entitlement/metered/grant_owner_adapter.go
```diff
@@ -121,50 +121,70 @@ func (e *entitlementGrantOwner) GetPeriodStartTimesBetween(ctx context.Context,
...
+	er := e.entitlementRepo.WithTx(ctx, tx)
+
+	entitlementEntity, err := er.GetEntitlement(ctx, owner.NamespacedID())
+	if err != nil {
+		return err
+	}
+
+	if entitlementEntity.UsagePeriod == nil {
+		e.logger.Warn("metered entitlement does not have a usage period set", "entitlement", owner.NamespacedID().String())
+		return nil
```
GAlexIHU, line 162: I don't think this is the reasonable default behavior, it's just convenient for you here

It should attempt the update based on the inputs & fail if it's not possible. Also, it's arguments could include the entire usageperiod

internal/entitlement/connector.go
```diff
@@ -88,7 +88,29 @@ func (c *entitlementConnector) CreateEntitlement(ctx context.Context, input Crea
...
+		if err != nil {
+			return nil, err
+		}
+		usagePeriod = &UsagePeriodWithNextReset{
+			UsagePeriod: *input.UsagePeriod,
+			NextReset:   nextResetAt,
+		}
+	}
+
+	ent, err := c.entitlementRepo.CreateEntitlement(ctx, CreateEntitlementRepoInputs{
```
GAlexIHU, line 103: why are the interfaces different now (repo create vs create)?

internal/entitlement/entitlement.go
```diff
@@ -78,17 +85,32 @@ type GenericProperties struct {
...
+	UsagePeriod *UsagePeriodWithNextReset `json:"usagePeriod,omitempty"`
 }
 
 type UsagePeriod struct {
 	Anchor   time.Time           `json:"anchor"`
 	Interval UsagePeriodInterval `json:"interval"`
 }
 
+func (u UsagePeriod) GetNextReset() (time.Time, error) {
+	rec := recurrence.Recurrence{
```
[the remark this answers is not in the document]
turip: It would, but it uses a slightly different schematics of Anchor/Period vs Anchor/Interval. I wouldn't make this PR more complex than this.

internal/entitlement/connector.go
turip: As the calculations should happen inside the entitlement connector not at the repo side.

internal/entitlement/metered/grant_owner_adapter.go
turip: This should not happen, but agree, changed the behavior.

api/openapi.yaml
```diff
@@ -1529,7 +1537,23 @@ components:
...
+        - lastReset
+        - currentUsagePeriod
+      properties:
+        lastReset:
+          description: |
+            The last time the meter was reset.
+          type: string
+          format: date-time
+          example: "2023-01-01T00:00:00Z"
+        currentUsagePeriod:
```
GAlexIHU, line 1555: currentUsagePeriod is meaningful for all entitlements not just meteted

internal/entitlement/entitlement.go
```diff
@@ -84,14 +87,46 @@ type GenericProperties struct {
...
+	UsagePeriod        *UsagePeriod `json:"usagePeriod,omitempty"`
+	CurrentUsagePeriod *api.Period  `json:"currentUsagePeriod,omitempty"`
 }
 
 type UsagePeriod struct {
 	Anchor   time.Time           `json:"anchor"`
 	Interval UsagePeriodInterval `json:"interval"`
 }
 
+func (u UsagePeriod) ToRecurringPeriod() api.RecurringPeriod {
```
GAlexIHU, line 99: i don't like api being referred in the core internal package, lets do the mapping in the adapter/driver layer

internal/entitlement/metered/entitlement.go
```diff
@@ -42,12 +49,22 @@ func ParseFromGenericEntitlement(model *entitlement.Entitlement) (*Entitlement,
 		return nil, &entitlement.InvalidValueError{Message: "UsagePeriod is required", Type: model.EntitlementType}
 	}
 
+	if model.LastReset == nil {
```
GAlexIHU, line 52: but what if it's never been reset?

internal/entitlement/metered/entitlement.go
```diff
@@ -23,6 +24,12 @@ type Entitlement struct {
 
 	// UsagePeriod defines the recurring period for usage calculations.
 	UsagePeriod entitlement.UsagePeriod `json:"usagePeriod,omitempty"`
+
+	// CurrentPeriod defines the current period for usage calculations.
+	CurrentUsagePeriod api.Period `json:"currentUsagePeriod,omitempty"`
```
GAlexIHU, line 29: - lets not reference api in the core implementations
- shouldn't this be optional?

internal/entitlement/metered/grant_owner_adapter.go
```diff
@@ -121,50 +121,69 @@ func (e *entitlementGrantOwner) GetPeriodStartTimesBetween(ctx context.Context,
...
+func (e *entitlementGrantOwner) updateEntitlementUsagePeriod(ctx context.Context, tx *entutils.TxDriver, owner credit.NamespacedGrantOwner, params credit.EndCurrentUsagePeriodParams) error {
+	er := e.entitlementRepo.WithTx(ctx, tx)
+
+	entitlementEntity, err := er.GetEntitlement(ctx, owner.NamespacedID())
+	if err != nil {
+		return err
+	}
+
+	if entitlementEntity.UsagePeriod == nil {
+		return fmt.Errorf("entitlement=%s, namespace=%s does not have a usage period set, cannot guess interval", owner.ID, owner.Namespace)
```
GAlexIHU, line 161: nit: lets not use the word guess

internal/entitlement/postgresadapter/ent/schema/entitlement.go
```diff
@@ -35,8 +36,10 @@ func (Entitlement) Fields() []ent.Field {
 		field.Float("issue_after_reset").Optional().Nillable().Immutable(),
 		field.Bool("is_soft_limit").Optional().Nillable().Immutable(),
 		field.JSON("config", map[string]interface{}{}).Optional(),
-		field.Enum("usage_period_interval").Values(entitlement.UsagePeriodInterval("").StrValues()...).Optional().Nillable().Immutable(),
-		field.Time("usage_period_anchor").Optional().Nillable().Immutable(),
+		field.Enum("usage_period_interval").Values(recurrence.RecurrencePeriod("").Values()...).Optional().Nillable().Immutable(),
```
GAlexIHU, line 39: in the db you use recurrence.RecurrencePeriod but in the types its entitlement.UsagePeriodInterval. lets make these consistent

internal/entitlement/postgresadapter/entitlement.go
```diff
@@ -175,6 +184,13 @@ func mapEntitlementEntity(e *db.Entitlement) *entitlement.Entitlement {
 		IsSoftLimit:      e.IsSoftLimit,
 	}
 
+	switch {
+	case len(e.Edges.UsageReset) > 0:
+		ent.LastReset = convert.ToPointer(e.Edges.UsageReset[0].ResetTime.In(time.UTC))
+	case e.MeasureUsageFrom != nil:
+		ent.LastReset = convert.ToPointer(e.MeasureUsageFrom.In(time.UTC))
+	}
```
GAlexIHU, line 192: nit: i don't prefer unrelated switch cases, i'd much rather use plain `if`s

internal/entitlement/postgresadapter/entitlement.go
```diff
@@ -186,12 +202,19 @@ func mapEntitlementEntity(e *db.Entitlement) *entitlement.Entitlement {
...
 	if e.UsagePeriodAnchor != nil && e.UsagePeriodInterval != nil {
-		ent.GenericProperties.UsagePeriod = &entitlement.UsagePeriod{
+		ent.UsagePeriod = &entitlement.UsagePeriod{
 			Anchor:   e.UsagePeriodAnchor.In(time.UTC),
 			Interval: entitlement.UsagePeriodInterval(*e.UsagePeriodInterval),
 		}
 	}
 
+	if e.CurrentUsagePeriodEnd != nil && e.CurrentUsagePeriodStart != nil {
+		ent.CurrentUsagePeriod = &api.Period{
```
GAlexIHU, line 212: lets not use api in our data access layer

GAlexIHU (approved):

PR #1071, opened 2024-06-26
fix: include grants issued for resettime into calculations

GAlexIHU:
## Overview

- fixes the edgecase where a grant is issued for the same time as the last reset happened

Fixes #(issue)

PR #1070, opened 2024-06-26
Store & Retrieve Metadata on Entitlement Entities

GAlexIHU:
## Overview

Stores and retrieves metadata for entitlement, grant, feature

Fixes #(issue)

## Notes for reviewer

- needs this to be merged first https://github.com/openmeterio/openmeter/pull/1068

PR #1069, opened 2024-06-26
feat: make recurrence anchor optional for grant create with default o…

GAlexIHU:
…f effectiveAt

## Overview

- Makes anchor of recurring grants optional with a default of the effectiveAt time.

Fixes #(issue)

## Notes for reviewer
- needs this PR to be merged first https://github.com/openmeterio/openmeter/pull/1068

PR #1072, opened 2024-06-26
Entitlement Improvements

GAlexIHU:
## Overview

This PR combines 4 other PRs 1 commit / PR for the following PRs:
- https://github.com/openmeterio/openmeter/pull/1071
- https://github.com/openmeterio/openmeter/pull/1070
- https://github.com/openmeterio/openmeter/pull/1069
- https://github.com/openmeterio/openmeter/pull/1066

Fixes #(issue)


hekike (approved):

internal/entitlement/httpdriver/parser.go
```diff
@@ -28,7 +28,7 @@ func (parser) ToMetered(e *entitlement.Entitlement) (*api.EntitlementMetered, er
 		Id:              &metered.ID,
 		IsUnlimited:     convert.ToPointer(false), // implement
 		IssueAfterReset: metered.IssuesAfterReset,
-		Metadata:        nil, // implement
+		Metadata:        &metered.Metadata,
```
turip, line 31: Please make sure that we are not taking the address of a nil map

GAlexIHU: ent populates that field in all cases and it always sets it to an empty map when there's no value (the field is not nillable), so it will not result in an error. do you wan't me to add safe dereferencing now?

turip (approved):

PR #1079, opened 2024-06-27
feat: return history from last reset

turip:
## Overview

This patch returns the EntitlementHistory starting from the last reset (or if there's not restet the startMeasurementFrom interval).

### DIFF
--- a/internal/entitlement/metered/balance.go
+++ b/internal/entitlement/metered/balance.go
@@ -95,38 +95,27 @@ func (e *connector) GetEntitlementBalanceHistory(ctx context.Context, entitlemen
 	// TODO: we should guard against abuse, getting history is expensive
 
 	// validate that we're working with a metered entitlement
-	ent, err := e.entitlementRepo.GetEntitlement(ctx, entitlementID)
+	entRepoEntity, err := e.entitlementRepo.GetEntitlement(ctx, entitlementID)
 	if err != nil {
 		return nil, credit.GrantBurnDownHistory{}, err
 	}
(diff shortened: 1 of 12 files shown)


internal/entitlement/metered/balance.go
```diff
@@ -39,7 +39,7 @@ const (
 )
 
 type BalanceHistoryParams struct {
-	From           time.Time
+	From           *time.Time
 	To             time.Time
```
GAlexIHU, line 43: it looks weird to me that To is required but from isn't. lets set both in here

internal/entitlement/metered/balance.go
```diff
@@ -95,26 +95,27 @@ func (e *connector) GetEntitlementBalanceHistory(ctx context.Context, entitlemen
...
+	entRepoEntity, err := e.entitlementRepo.GetEntitlement(ctx, entitlementID)
 	if err != nil {
 		return nil, credit.GrantBurnDownHistory{}, err
 	}
-	_, err = ParseFromGenericEntitlement(ent)
-	if err != nil {
-		return nil, credit.GrantBurnDownHistory{}, err
+
+	if entRepoEntity == nil {
+		return nil, credit.GrantBurnDownHistory{}, &entitlement.NotFoundError{EntitlementID: entitlementID}
```
GAlexIHU, line 104: can this actually happen?

turip: Nope, but I would prefer to be a bit defensive regardless in case somebody refactors the code.

GAlexIHU (approved):

PR #1093, opened 2024-06-28
feat(api): allow filtering entitlements by feature IDs

chrisgacsal:
## Overview

Add `FeatureIDs` to `entitlement.ListEntitlementsParams` to allow filtering _entitlements_ based on _features_ they are belong to.

### DIFF
--- a/internal/entitlement/connector.go
+++ b/internal/entitlement/connector.go
@@ -20,6 +20,7 @@ const (
 type ListEntitlementsParams struct {
 	Namespace      string
 	SubjectKey     string
+	FeatureIDs     []string
 	Limit          int
 	Offset         int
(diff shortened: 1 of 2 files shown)


turip (approved):

internal/entitlement/connector.go
```diff
@@ -20,6 +20,7 @@ const (
 type ListEntitlementsParams struct {
 	Namespace      string
 	SubjectKey     string
+	FeatureIDs     []string
```
GAlexIHU, line 23: lets call it Keys if it filters for keys

chrisgacsal: Nice catch. Will open a PR with fix. Actually we need to filter based on _ids_ not _keys_.

PR #1105, opened 2024-06-28
feat: include entitlement creation in a single transaction

GAlexIHU:
## Overview

Ensures that entitlement creation happens as a single atomic operation

Fixes #(issue)

## Notes for reviewer
We'll have to refactor later to a more generic pattern


turip (approved):

turip:
We will most probably rework txn handling, but for the time being it gets the job done.

turip (approved):

PR #1106, opened 2024-06-28
feat: add support for 1d meters

turip:
The entitlement engine will use the meter's window resultion instead of defaulting to minute.

### DIFF
--- a/internal/credit/owner_connector.go
+++ b/internal/credit/owner_connector.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/openmeterio/openmeter/internal/streaming"
 	"github.com/openmeterio/openmeter/pkg/framework/entutils"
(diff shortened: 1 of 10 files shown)


internal/credit/owner_connector.go
```diff
@@ -7,15 +7,22 @@ import (
...
 	"github.com/openmeterio/openmeter/pkg/framework/entutils"
+	"github.com/openmeterio/openmeter/pkg/models"
 )
 
 type EndCurrentUsagePeriodParams struct {
 	At           time.Time
 	RetainAnchor bool
 }
 
+type OwnerMeter struct {
```
GAlexIHU, line 18: i think we can use simply meter, thats the core of the product after all

GAlexIHU (approved):

PR #1115, opened 2024-06-30
fix: make sure grant fetching fetches active grants

turip:
## Overview

The previous algorithm didn't properly determine the grants in a certain time-rage. This patch ensures that the repo works as expected.

If we are missing grants that would lead to invalid balances and also snapshots that are inconsistent and thus returning 500 errors.

## Impact

Some queries requiring balance were returning inconsistent values due to grants being missed.

### DIFF
--- a/internal/credit/postgresadapter/grant.go
+++ b/internal/credit/postgresadapter/grant.go
@@ -109,18 +109,22 @@ func (g *grantDBADapter) ListGrants(ctx context.Context, params credit.ListGrant
 }
 
 func (g *grantDBADapter) ListActiveGrantsBetween(ctx context.Context, owner credit.NamespacedGrantOwner, from, to time.Time) ([]credit.Grant, error) {
+
 	query := g.db.Grant.Query().
-		Where(db_grant.And(db_grant.OwnerID(owner.ID), db_grant.Namespace(owner.Namespace))).
 		Where(
-			db_grant.Or(
-				db_grant.And(db_grant.EffectiveAtLT(from), db_grant.ExpiresAtGT(from)),
-				db_grant.And(db_grant.EffectiveAtGTE(from), db_grant.EffectiveAtLT(to)),
-				db_grant.EffectiveAt(from),
+			db_grant.And(
+				// Filter by owner, namespace
+				db_grant.OwnerID(owner.ID),
+				db_grant.Namespace(owner.Namespace),
+				// We are not interested in grants that either expired before the period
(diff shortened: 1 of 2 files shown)


internal/credit/postgresadapter/grant.go
```diff
@@ -109,18 +109,22 @@ func (g *grantDBADapter) ListGrants(ctx context.Context, params credit.ListGrant
...
+				// Filter by owner, namespace
+				db_grant.OwnerID(owner.ID),
+				db_grant.Namespace(owner.Namespace),
+				// We are not interested in grants that either expired before the period
+				db_grant.Not(db_grant.EffectiveAtGT(to)),
+				// We are not interested in grants that are not yet effective
+				db_grant.Not(db_grant.ExpiresAtLT(from)),
+				// We are not interested in deleted or voided grants
+				db_grant.Or(db_grant.DeletedAtGTE(to), db_grant.DeletedAtIsNil()),
+				db_grant.Or(db_grant.VoidedAtGTE(to), db_grant.VoidedAtIsNil()),
```
GAlexIHU, line 125: this query is quite tricky to get right (i've also rewritten it like 5 times by now), i'm gonna type the rules and then see if we agree in the query:
- this query has to return all grants that are active at ANY point in time in the queried period
- the period input can be nonsensical (e.g. from can be after to, that still has to work)
- the period can be a single point in time

Usually the periods we use are inclusive => exclusive (start => finish). In this case that's not quite the case, maybe the method could be renamed to `ListAllRelevantGrantsBetween`. We do have to list grants with deletedAt & voidedAt = to as otherwise they would never be accounted in any calculation:
Grant deleted (or voided) at t2
Query1: t1=>t2       (this should include those grants)
Query2: t2=>t3      (this should not include those grant)

internal/credit/postgresadapter/grant.go
```diff
@@ -129,6 +133,12 @@ func (g *grantDBADapter) ListActiveGrantsBetween(ctx context.Context, owner cred
 
 	grants := make([]credit.Grant, 0, len(entities))
 	for _, entity := range entities {
+		// Let's prefilter these as they are not really active, but will make the engine
+		// think they are active for that single point in time.
+		if entity.EffectiveAt.Equal(entity.ExpiresAt) {
+			continue
+		}
```
GAlexIHU, line 140: this looks really weird to me, lets not do this (?)
What was the underlying issue?

PR #1126, opened 2024-07-01
fix: remove redundant error check

csatib02:
## Overview

- Removes a redundant error check that was previously removed in #880 but was reverted.


GAlexIHU:
Hello @csatib02, nice catch, thank you for your contribution!

turip (approved):

PR #1128, opened 2024-07-01
feat: fix issues when grant expires the same time when a reset happens

turip:
## Overview

This patch adds the clock package that we can use to model complex scenarios over time. The first regression test illustrates the usage.

This patch also fixes the underlying issue:
- if a grant expires same time as the reset

### DIFF
--- a/pkg/clock/clock_test.go
+++ b/pkg/clock/clock_test.go
@@ -1,0 +1,23 @@
+package clock_test
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/openmeterio/openmeter/internal/testutils"
+	"github.com/openmeterio/openmeter/pkg/clock"
+)
+
+func TestClock(t *testing.T) {
+	clock.SetTime(testutils.GetRFC3339Time(t, "2024-06-30T15:39:00Z"))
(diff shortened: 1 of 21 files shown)


GAlexIHU:
* if a grant expires same time as the reset

pkg/clock/clock_test.go
```diff
@@ -0,0 +1,23 @@
...
+func TestClock(t *testing.T) {
+	clock.SetTime(testutils.GetRFC3339Time(t, "2024-06-30T15:39:00Z"))
+	defer clock.ResetTime()
+
+	now := clock.Now()
+	diff := now.Sub(testutils.GetRFC3339Time(t, "2024-06-30T15:39:00Z"))
+	if diff < 0 {
+		diff = -diff
+	}
+	assert.True(t, diff < time.Second)
```
GAlexIHU, line 22: i'm not sure i understand the point of this test

GAlexIHU (approved):

pkg/clock/clock_test.go
turip: This test just ensures that the calculation is pseudo correct. If I would have messed up the sign of the drift this would fail.

GAlexIHU (approved):

PR #1135, opened 2024-07-02
feat(server): return event count as prom metrics

hekike:
Return event count since UTC midnight in Prometheus format to help vendors monitor data quality.


chrisgacsal (approved):
LGTM! Left a tiny suggestion regarding to the query building.

PR #1138, opened 2024-07-02
refactor: split ingest and kafka specific config

turip:
## Overview

This allows us to reuse the Kafka configuration object in other parts of the system/other configuration files.

## Notes for reviewer

The mapstructure tag is required so that viper merges the two objects as expected.


chrisgacsal (approved):
LGTM!

PR #1141, opened 2024-07-03
fix(router): namespace driver default

hekike:
- fix router namespace adapter namespace
- validate router config


turip (approved):

GAlexIHU (commented):
What if entitlements are disabled? We don't have noop implementations for all of those, and i'm not sure writing them is the right approach (in solving conditional modules)

PR #1144, opened 2024-07-04
Support feature creation by meter Id OM-719

GAlexIHU:
## Overview

Includes (per commit):
- make task to run e2e tests
- removes `*map[string]string` for feature `meterGroupByFilters` and replaces it with a custom type
- Proposed solution for [OM-719 : Create Feature with MeterID](https://linear.app/openmeter/issue/OM-719/create-feature-with-meterid)

- This PR changes the API which I don't like that much, another option would be to add an extra input property `meterId` and then we'd document which has precedence, but I that doesn't seem the best either.
- MeterID shouldn't be added to the Feature directly (IMO) as in most cases MeterID doesn't exist... the idea is to provide a convenience when creating a feature


internal/productcatalog/httpdriver/feature.go
```diff
@@ -80,19 +81,25 @@ type CreateFeatureHandler httptransport.Handler[CreateFeatureHandlerRequest, Cre
 func (h *featureHandlers) CreateFeature() CreateFeatureHandler {
 	return httptransport.NewHandler(
 		func(ctx context.Context, r *http.Request) (productcatalog.CreateFeatureInputs, error) {
-			featureIn := productcatalog.CreateFeatureInputs{}
-			if err := commonhttp.JSONRequestBodyDecoder(r, &featureIn); err != nil {
-				return featureIn, err
+			parsedBody := api.CreateFeatureJSONRequestBody{}
+			emptyFeature := productcatalog.CreateFeatureInputs{}
```
turip, line 85: we should switch to pointers (regardless, I am fine merging this)

pkg/convert/ptr.go
```diff
@@ -32,3 +32,16 @@ func SafeToUTC(t *time.Time) *time.Time {
...
 	})
 }
+
+// Header represents generic primitives with a header like map, slice, array...
+type Header[E any] interface {
+	map[string]E | []E
+}
+
+// Safely dereference a pointer to a slice or map
+func DerefHeaderPtr[E any, T Header[E]](header *T) T {
```
turip, line 42:
```suggestion
func DerefCompositeTypePtr[E any, T Header[E]](header *T) T {
```

The header is misleading for me. According to the [go spec](https://go.dev/ref/spec):

```
Composite types—array, struct, pointer, function, interface, slice, map, and channel types—may be constructed using type literals.
```

It's not a perfect match, but IMHO better then header.

turip (changes requested):

pkg/convert/ptr.go
GAlexIHU: So what I was referring to was that they're nillable cause they point to their header struct in runtime
https://github.com/golang/go/blob/master/src/runtime/map.go#L109
https://github.com/golang/go/blob/master/src/runtime/slice.go (slice header has since been "removed", e.g. [see this depracation](https://pkg.go.dev/reflect#SliceHeader), now it's just slice)

turip (approved):

GAlexIHU:
Closing due to inactivity

PR #1170, opened 2024-07-08
feat: add measureUsageFrom input OM-751

GAlexIHU:
## Overview

Proposed API for specifying start of measurement for metered entitlements:
- either a timestamp
- or the current period start
- or the default of createdAt if not specified

Fixes OM-751

## Notes for reviewer
- Would it make sense to change the default to the current period start & add another enum value for createdAt? I'm not sure the ENUM side of the API is intuitive


internal/entitlement/httpdriver/entitlement.go
```diff
@@ -94,6 +94,29 @@ func (h *entitlementHandler) CreateEntitlement() CreateEntitlementHandler {
 				if v.Metadata != nil {
 					request.Metadata = *v.Metadata
 				}
+				if v.MeasureUsageFrom != nil {
+					measureUsageFrom := &entitlement.MeasureUsageFromInput{}
+					apiTime, err := v.MeasureUsageFrom.AsMeasureUsageFromTime()
+					if err == nil {
+						measureUsageFrom.FromTime(apiTime)
```
turip, line 101: Missing error check

turip (approved):

PR #1178, opened 2024-07-09
feat: allow access to entitlement parser

turip:
## Overview

This allows us to reuse the value to api parsing logic in the Cloud without code duplication.


internal/entitlement/httpdriver/entitlement.go
```diff
@@ -222,6 +196,34 @@ func (h *entitlementHandler) GetEntitlementValue() GetEntitlementValueHandler {
 	)
 }
 
+func MapEntitlementValueToAPI(entitlementValue entitlement.EntitlementValue) (api.EntitlementValue, error) {
```
GAlexIHU, line 199: nit: maybe move to parser.go?

GAlexIHU (dismissed):

internal/entitlement/httpdriver/entitlement.go
turip: Done.

GAlexIHU (approved):

PR #1199, opened 2024-07-15
feat(sink-worker): enable kafka client logging

chrisgacsal:
## Overview

Allow setting debug context for Kafka client configuration in order to get more detailed logging emitted by the `librdkafka`.

See: https://github.com/confluentinc/librdkafka/blob/master/INTRODUCTION.md#debug-contexts

## Notes for reviewer

This change allows us to have a better insights how Kafka clients interact with brokers and helps with spotting potentional issues which are not (properly) surfaced.

### DIFF
--- a/config/ingest.go
+++ b/config/ingest.go
@@ -64,6 +64,10 @@ type KafkaConfiguration struct {
 	// in case of large clusters where changes are more frequent.
 	// This value must not be set to value lower than 10s.
 	TopicMetadataRefreshInterval pkgkafka.TimeDurationMilliSeconds
+
+	// Enable contexts for extensive debugging of librdkafka.
+	// See: https://github.com/confluentinc/librdkafka/blob/master/INTRODUCTION.md#debug-contexts
+	DebugContexts pkgkafka.DebugContexts
 }
 
 func (c KafkaConfiguration) Validate() error {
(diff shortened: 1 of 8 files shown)


config/ingest.go
```diff
@@ -64,6 +64,10 @@ type KafkaConfiguration struct {
 	// in case of large clusters where changes are more frequent.
 	// This value must not be set to value lower than 10s.
 	TopicMetadataRefreshInterval pkgkafka.TimeDurationMilliSeconds
+
+	// Enable contexts for extensive debugging of librdkafka.
+	// See: https://github.com/confluentinc/librdkafka/blob/master/INTRODUCTION.md#debug-contexts
+	DebugContexts []string
```
sagikazarmark, line 70: Is this an exhaustive list? Do we validate invalid values?

config/ingest.go
```diff
@@ -134,6 +138,10 @@ func (c KafkaConfiguration) CreateKafkaConfig() kafka.ConfigMap {
 		config["metadata.max.age.ms"] = 3 * c.TopicMetadataRefreshInterval
 	}
 
+	if c.DebugContexts != nil {
+		config["debug"] = strings.Join(c.DebugContexts, ",")
+	}
```
sagikazarmark, line 143: Empty list may result in an empty string. Might be better to check for length?

sagikazarmark (dismissed):

config/ingest.go
chrisgacsal: Valid. Will fix it promptly.

config/ingest.go
chrisgacsal: Fixed.

config/ingest.go
chrisgacsal: Fixed.

sagikazarmark (approved):

PR #1232, opened 2024-07-23
Unify Ent DB Schemas in central place

GAlexIHU:
## Overview

Moves the separate ent db schemas under `internal/ent` and uses that db client in all adapter implementations

Until commit 75cd48a93ca9ebe01644ab4e2720a831707d2a63 it doesn't contain any changes to the underlying schemas

## Notes for reviewer

- still looking into how to validate PG schema changes across versions


cmd/server/main.go
```diff
@@ -575,56 +573,26 @@ func initNamespace(config config.Configuration, namespaces ...namespace.Handler)
...
-	driver                 *entDialectSQL.Driver
-	entitlementDBClient    *entitlementdb.Client
-	productcatalogDBClient *productcatalogdb.Client
-	creditDBClient         *creditdb.Client
-}, error,
+func initPGClients(config config.PostgresConfig) (
+	*entDialectSQL.Driver,
+	*db.Client,
+	error,
 ) {
```
turip, line 580: Can we please just return a named struct?

```suggestion

type pgClients struct {
    driver *entDialectSQL.Driver
	client *db.Client
}

func initPGClients(config config.PostgresConfig) (
	*pgClients,
	error,
) {
```

GAlexIHU: IMO 3 return values is ok but sure

turip (approved):

PR #1234, opened 2024-07-23
refactor(api): return meta object instead of id

chrisgacsal:
## Overview

Update Notification API to return metadata of the resources in responses
instead of their identifiers as:

* there are resources (e.g. `feature`) which may have two unique
  identifiers wher one generated by the system (`id`) and the other provided by
  the user (`key`)
* this response structure allows us to add support for resource
  *expand* feature where users can request the API to expand certain
  fields in the response like returning the full resources instead of
  their limited representaion (including only their metadata like
  unique identifiers)

Additionally rename Notification API related API resources using `Generic` prefix to `Common`.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -2738,7 +2738,7 @@ components:
 
     NotificationChannelWebhookCreateRequest:
       allOf:
-        - $ref: "#/components/schemas/GenericNotificationChannelCreateRequest"
+        - $ref: "#/components/schemas/NotificationChannelCommonCreateRequest"
         - type: object
           description: "Request for updating existing webhook notification channel."
           required:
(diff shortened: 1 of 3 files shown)


turip (approved):

api/openapi.yaml
```diff
@@ -1859,6 +1859,25 @@ components:
               example: "2023-01-01T00:00:00Z"
         - $ref: "#/components/schemas/FeatureCreateInputs"
         - $ref: "#/components/schemas/SharedMetaFields"
+
+    FeatureMeta:
+      type: object
+      description: |
+        Limited representation of a feature resource which includes only its unique identifiers (id, key).
+      required:
+        - id
```
GAlexIHU, line 1868: key is required

GAlexIHU (changes requested):

GAlexIHU (approved):

PR #1239, opened 2024-07-23
refactor(api): notification event payload

chrisgacsal:
## Overview

Add notification event type playload to the OepnAPI specification. Currently only payload for notifications with `entitlements.balance.threshold` type is supported.

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -3078,66 +3078,63 @@ components:
           $ref: "#/components/schemas/NotificationEventPayload"
 
     NotificationEventPayload:
-      description: |
-        The actual payload sent to channel as part of the notification event.
+      description: The actual payload sent to channel as part of the notification event.
+      oneOf:
+        - $ref: "#/components/schemas/NotificationEventBalanceThresholdPayload"
+      discriminator:
+        propertyName: type
+        mapping:
+          entitlements.balance.threshold: "#/components/schemas/NotificationEventBalanceThresholdPayload"
 
-        The `data` field contains the information specific to the notification event
-        which may vary based on the `type` of the event.
+    NotificationEventCommonPayload:
+      description: Common fields for notification event payload.
       type: object
       required:
         - id
         - type
         - timestamp
-        - data
       properties:
         id:
           description: A unique identifier for the notification event the payload belongs to.
           type: string
           example: "01J2KNP1YTXQRXHTDJ4KPR7PZ0"
         type:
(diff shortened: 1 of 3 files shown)


api/openapi.yaml
```diff
@@ -3078,66 +3078,90 @@ components:
...
+            - data
+          properties:
+            data:
+              type: object
+              required:
+                - entitlement
+                - feature
+                - subject
+                - balance
+              properties:
```
tothandras, line 3125: `threshold` seems to be missing

api/openapi.yaml
```diff
@@ -3078,66 +3078,90 @@ components:
...
+                entitlement:
+                  $ref: "#/components/schemas/EntitlementMetered"
+                feature:
+                  $ref: "#/components/schemas/Feature"
+                subject:
+                  $ref: "#/components/schemas/Subject"
+                balance:
```
tothandras, line 3132: Isn't there a balance type?

api/openapi.yaml
```diff
@@ -3078,66 +3078,90 @@ components:
           $ref: "#/components/schemas/NotificationEventPayload"
 
     NotificationEventPayload:
-      description: |
-        The actual payload sent to channel as part of the notification event.
+      description: The actual payload sent to channel as part of the notification event.
+      oneOf:
+        - $ref: "#/components/schemas/NotificationEventBalanceThresholdPayload"
+      discriminator:
```
tothandras, line 3084: 👏🏻

api/openapi.yaml
chrisgacsal: Nice catch!

api/openapi.yaml
chrisgacsal: So I was kinda hasitant to use `EntitlementValue` directly as it is shared across all the entitlement types meaning it has fields for all teh entitlement types while we only need `metered` here. So it might be confusing for users to see optinal fields in the spec which will be never used for `entitlements.balance.threshold` payload.

It would be awesome to have entitlement specific `Entitlement[Metered|Static|Boolean]Value` resources in the spec to support better reusability, but making this type of change out of scope for this PR.

So I am ok with using `EntitlementValue` here, but I'd improve on this so make the API spec more clean.

@GAlexIHU

api/openapi.yaml
chrisgacsal: Fixed.

tothandras (approved):

PR #1240, opened 2024-07-23
Add manual diff tests for schema changes

GAlexIHU:
## Overview

Adds a new dagger function for running diff tests between the current HEAD (including uncommitted) and a provided `baseRef`.  Example command:
```
dagger call --source .:default etoe diff --baseRef 40f70fc1dd4153050038de513aba2a1d8ae5e9d5
```
The only test written for now asserts that a given change doesn't result a database schema change via ent automigrate.

Currently this is not included in the CI but it would be reasonably easy to do so. The reason for that is that `baseRef` has to have `tools/migrate` present, which main does not have, only this branch.

## Notes for reviewer

Later on it would be easy to generalize the setup process (currently its linked to this single usecase)


ci/e2e.go
```diff
@@ -49,6 +63,39 @@ func (m *Ci) Etoe(
...
 
+func (e *Etoe) Diff(
+	ctx context.Context,
+	// The base ref against which to run the diffs
+	baseRef string,
+) error {
+	db := postgres()
+
+	// Migrate DB with base state
+	_, err := goModule().
```
sagikazarmark, line 74: Might make sense to use `goModuleSource("")`

ci/e2e.go
```diff
@@ -49,6 +63,39 @@ func (m *Ci) Etoe(
...
+		Sync(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to migrate base ref %s: %w", baseRef, err)
+	}
+
+	_, err = goModule().
+		WithSource(e.Ci.Source).
+		Container().
+		WithServiceBinding("postgres", db).
+		WithEnvVariable("POSTGRES_URL", "postgres://postgres:postgres@postgres:5432/postgres").
```
sagikazarmark, line 92: These could be consolidated into a single container with the above. It would improve caching.

GAlexIHU: consolidated them to single container

GAlexIHU: WithExec doesn't provide me a proper shell so I couldn't find a way to go back to the original revision (i couldn't get the current head sha to env or an external variable) without using a hack like a local tag i name or a util script, so i separated them again

sagikazarmark: You can `WithExec([]string{"sh", "-c", "git something something"})

e2e/diff/no_schema_diff_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+func setup(t *testing.T) {
+	if PG_URL == "" {
+		t.Skip("POSTGRES_URL not set")
+	}
+}
+
+func TestNoSchemaDiffOnMigrate(t *testing.T) {
+	setup(t)
+	driver, err := entutils.GetPGDriver(PG_URL)
```
sagikazarmark, line 37: You could merge these into a single function (if reusability is needed). You can drop the global variable that way.

I'd also use `t.Helper()` to make sure the correct line is reported in case of an error.

tools/migrate/main.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+func main() {
+	PG_URL := os.Getenv("POSTGRES_URL")
+	if PG_URL == "" {
+		PG_URL = "postgres://postgres:postgres@localhost:5432/postgres"
+	}
+
+	logger := slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelDebug}))
+	slog.SetDefault(logger)
```
sagikazarmark, line 19: Do you know if ent uses a global logger instance?

tools/migrate/main.go
```diff
@@ -0,0 +1,34 @@
...
+	"context"
+	"log/slog"
+	"os"
+
+	"github.com/openmeterio/openmeter/internal/ent/db"
+	"github.com/openmeterio/openmeter/pkg/framework/entutils"
+)
+
+func main() {
+	PG_URL := os.Getenv("POSTGRES_URL")
```
sagikazarmark, line 13:
```suggestion
	pgURL := os.Getenv("POSTGRES_URL")
```

e2e/diff/no_schema_diff_test.go
```diff
@@ -0,0 +1,98 @@
...
+var PG_URL string
+
+func TestMain(m *testing.M) {
+	PG_URL = os.Getenv("POSTGRES_URL")
+
+	os.Exit(m.Run())
+}
+
+const (
+	DIR = "./tmp/"
```
sagikazarmark, line 26:
```suggestion
	dir = "./tmp/"
```

sagikazarmark: Maybe use `os.MkTemp`?

e2e/diff/no_schema_diff_test.go
```diff
@@ -0,0 +1,98 @@
...
+	migrate, err := schema.NewMigrate(driver, schema.WithDir(dir))
+	if err != nil {
+		t.Fatalf("failed to create migrate %s", err)
+	}
+	err = migrate.Diff(context.Background(), entMigrate.Tables...)
+	if err != nil {
+		t.Fatalf("failed to diff schema %s", err)
+	}
+
+	files, err := dir.Files()
```
sagikazarmark, line 66: I'm not sure if that's possible here, but if it is, using some dir walking method would be more elegant.

e2e/diff/no_schema_diff_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"testing"
+
+	"ariga.io/atlas/sql/migrate"
+	"entgo.io/ent/dialect/sql/schema"
+
+	entMigrate "github.com/openmeterio/openmeter/internal/ent/db/migrate"
+	"github.com/openmeterio/openmeter/pkg/framework/entutils"
+)
+
+var PG_URL string
```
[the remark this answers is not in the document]
sagikazarmark: Wonder why linter wasn't triggered. Maybe that rule isn't enabled yet?

e2e/diff/.gitignore
```diff
@@ -0,0 +1 @@
+tmp/**
```
sagikazarmark, line 1: I think this wouldn't be necessary with MkTemp

ci/e2e.go
sagikazarmark: Let's chat about this sometime. Happy to help you get this sorted.

GAlexIHU: What I meant is rather that I could only do it conceptually similar to this
```go
// Stash local changes
WithExec([]string{"git", "stash", "save", "-u", "-m", "diffteststash"}).
WithExec([]string{"ash", "-c", "echo $(git rev-parse HEAD) > /tmp/revision"}).
// Migrate
WithExec([]string{"git", "checkout", "-f", baseRef}).
WithExec([]string{"go", "run", "./tools/migrate"}).
// Return to original version
WithExec([]string{"ash", "-c", "git checkout $(cat /tmp/revision)"}).
WithExec([]string{"ash", "-c", "git stash apply stash^diffteststash || true"}).
// Run tests
WithExec([]string{"go", "test", "-v", "./e2e/diff/..."}).
WithExec([]string{"git", "stash", "apply"}).
```
which i consider to be hacky. I don't know of a way to get the revision value without using a workaround like this, that's why I decided to separate them again. (I wanted to write it to environment to extract it with `.EnvVariable("x")`, but the string arglist is not a proper shell and when using a subshell i don't have access to the parent env...)

(If a similar approach doesn't bother you then we could use it, but I'd personally rather just have two containers)

tools/migrate/main.go
GAlexIHU: I don't think it logs anything by itself, only if you add logging through a custom hook. `slog.SetDefault(logger)` is just there for good measure, but yes, ineffective as is right now.

e2e/diff/no_schema_diff_test.go
GAlexIHU: We could do a `filepath.Walk(...)` directly on the os for the tmp dir, but after using `.MkdirTemp()` the contents will always be flat

sagikazarmark (dismissed):

PR #1241, opened 2024-07-23
Adding a standardized pagination implementation to entitlement entities OM-779

GAlexIHU:
## Overview

The aim of this PR is to add a standardized pagination interface (Request Response) to the external and internal APIs. Currently included APIs:
- /api/v1/entitlements
- /api/v1/features
- /api/v1/grants

## Notes for reviewer
- Limit & Offset are kept for BW compatibility (which results in a little mess) but it doesn't break the API. Page has preference over limit&offset. 
- Due to supporting both, page and pagesize have no default values.


tothandras (approved):

tothandras (dismissed):

api/openapi.yaml
```diff
@@ -1860,6 +1868,36 @@ components:
         - $ref: "#/components/schemas/FeatureCreateInputs"
         - $ref: "#/components/schemas/SharedMetaFields"
 
+    FeaturePaginatedUnion:
```
sagikazarmark, line 1871: How about FeatureList or Features or FeatureWuery|ListResults?

api/openapi.yaml
```diff
@@ -3213,6 +3309,26 @@ components:
...
+      name: page
+      in: query
+      required: false
+      description: Page number to return
+      schema:
+        type: integer
+        minimum: 1
+        default: 1
+        example: 1
+    queryPageSize:
```
sagikazarmark, line 3322: Hos is this different from limit? I think I like that better.

GAlexIHU: You mean semantically? `limit` suggests the resultset will have a certain size, `pageSize` suggests that if i'm on `page` 5 with `pageSize` 20 I would get results 80-99 (i.e. its the "paging increment")

api/openapi.yaml
GAlexIHU: i'm fine with `ListFeaturesResponse`, renamed it

tothandras (approved):
Let's get this merged first

api/client/node/clients/entitlement.ts
```diff
@@ -1,5 +1,6 @@
 import { components, operations } from '../schemas/openapi.js'
 import { RequestOptions, BaseClient, OpenMeterConfig } from './client.js'
+import { Paginated } from './pagination.js'
```
tothandras, line 3: 🙏🏻 thanks!

PR #1272, opened 2024-07-30
feat: skaffold balance worker

turip:
## Overview

The balance worker will be responsible for emitting balance snapshots. For now we are using the sarama library, but as a seperate PR we will transition to the confluent supported one.

The event handler is unimplemented, so that the application logic will be visible in the seperate PR.


sagikazarmark (approved):

chrisgacsal (approved):
Nice work! :shipit:

PR #1273, opened 2024-07-30
refactor: move sink models to a seperate package

turip:
## Overview

This will allow us to have handlers that can iteract with the sink messages.


chrisgacsal (approved):
LGTM!

PR #1274, opened 2024-07-30
feat: initialize entitlements in balance worker

turip:
## Overview

Given that this would copy paste most of the code from server's main this patch moves some client initializaiton code into internal for better reusability.

### DIFF
--- a/internal/cmd/clickhouse/client.go
+++ b/internal/cmd/clickhouse/client.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/ClickHouse/clickhouse-go/v2"
+
 	"github.com/openmeterio/openmeter/config"
 )
 
--- a/internal/cmd/clickhouse/client.go
+++ b/internal/cmd/clickhouse/client.go
@@ -1,0 +1,40 @@
+package streaming
+
+import (
+	"crypto/tls"
+	"fmt"
+	"time"
+
+	"github.com/ClickHouse/clickhouse-go/v2"
+	"github.com/openmeterio/openmeter/config"
+)
+
+func GetClient(config config.Configuration) (clickhouse.Conn, error) {
+	options := &clickhouse.Options{
+		Addr: []string{config.Aggregation.ClickHouse.Address},
+		Auth: clickhouse.Auth{
+			Database: config.Aggregation.ClickHouse.Database,
+			Username: config.Aggregation.ClickHouse.Username,
+			Password: config.Aggregation.ClickHouse.Password,
+		},
+		DialTimeout:      time.Duration(10) * time.Second,
+		MaxOpenConns:     5,
+		MaxIdleConns:     5,
+		ConnMaxLifetime:  time.Duration(10) * time.Minute,
+		ConnOpenStrategy: clickhouse.ConnOpenInOrder,
+		BlockBufferSize:  10,
+	}
+	// This minimal TLS.Config is normally sufficient to connect to the secure native port (normally 9440) on a ClickHouse server.
+	// See: https://clickhouse.com/docs/en/integrations/go#using-tls
+	if config.Aggregation.ClickHouse.TLS {
+		options.TLS = &tls.Config{}
+	}
+
+	// Initialize ClickHouse
+	clickHouseClient, err := clickhouse.Open(options)
+	if err != nil {
+		return nil, fmt.Errorf("init clickhouse client: %w", err)
+	}
+
+	return clickHouseClient, nil
+}
--- a/internal/cmd/postgres/client.go
+++ b/internal/cmd/postgres/client.go
@@ -1,0 +1,39 @@
+package postgres
+
+import (
+	"context"
+	"fmt"
+
+	"entgo.io/ent/dialect/sql"
+	"github.com/openmeterio/openmeter/config"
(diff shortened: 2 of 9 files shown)


internal/common/entitlement/connectors.go
```diff
@@ -0,0 +1,96 @@
...
+
+type ConnectorsOptions struct {
+	DatabaseClient     *db.Client
+	StreamingConnector streaming.Connector
+	Logger             *slog.Logger
+	MeterRepository    meter.Repository
+	Publisher          publisher.TopicPublisher
+}
+
+func GetConnectors(opts ConnectorsOptions) *Connectors {
```
GAlexIHU, line 39: i don't think `internal/common/entitlement` is warranted. If you want to use it for server setup it could go to `internal/server` somewhere, if its needed each time entitlements is used it could go into `internal/entitlement`... I think "common" is a bit like util, it easily gets out of hand. Wdyt?

internal/cmd/postgres/client.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/openmeterio/openmeter/internal/ent/db"
+	"github.com/openmeterio/openmeter/pkg/framework/entutils"
+)
+
+type Clients struct {
+	Driver *sql.Driver
+	Client *db.Client
+}
+
+func GetClients(config config.PostgresConfig) (*Clients, error) {
```
GAlexIHU, line 18: Q: do you think the subpackages are warranted? (i.e. `cmd/xy`), couldn't it all just be under `internal/cmd`? (do you think it reasonable to have multiple files per package?)

internal/common/entitlement/connectors.go
turip: It's for component setups (not just server, but balance-worker, notifications worker); I didn't want to put this into `internal/entitlement` as it also provides product catalog + credit. But I am completely fine putting it there into a registry package or something

internal/cmd/postgres/client.go
turip: I am fine merging those just I wanted to prevent this becoming a kitchen sink package.

internal/common/entitlement/connectors.go
turip: On the secound tought I would rather have an `internal/registry` package that can yield these connector structs, such as Entitlements (that includes credit + product catalog) or the streaming connector.

internal/cmd/postgres/client.go
sagikazarmark: Splitting up later is always easier than merging things back, so I'm also on the side of keeping things together, until it doesn't make sense.

internal/cmd/clickhouse/client.go
```diff
@@ -0,0 +1,41 @@
...
+	"crypto/tls"
+	"fmt"
+	"time"
+
+	"github.com/ClickHouse/clickhouse-go/v2"
+
+	"github.com/openmeterio/openmeter/config"
+)
+
+func GetClient(config config.Configuration) (clickhouse.Conn, error) {
```
sagikazarmark, line 13: Not sure if it's a good idea, but this could be a function on the struct behind `config.Aggregation.ClickHouse`.

That approach worked pretty well with the polymorphic configurations (dedupe).

It doesn't have to be the final location though. But at least keeps the input (config) and the constructor together for now.

WDYT?

internal/common/entitlement/connectors.go
sagikazarmark: I generally prefer keeping wiring in main, but if it's too big a burden we can think about introducing some alternative dependency injection strategy.

I've been experimenting with this a while ago: https://github.com/sagikazarmark/go-service-locator

There is Google's wire too: https://github.com/google/wire

Uber has fx: https://github.com/uber-go/fx

I'd like to keep it as simple and as framework independent as possible though.

I'd propose going with something simple for now and then sitting down and talking through some long-term options in the coming weeks.

internal/cmd/clickhouse/client.go
turip: I would love to take that approach. I used this split because we had this discussion with @chrisgacsal when doing similar things in the cloud code-base and it felt strange to him to add these to the config structs. But for simplicity's sake I will take that approach for now.

internal/common/entitlement/connectors.go
turip: Ok, for now I have created a repository package for this one only. If we end up having more such components where you need to always initialize, we can put additional stuff there.

It's included in this [PR](https://github.com/openmeterio/openmeter/pull/1278), I will close this (I didn't wanted to force push this one with the new contents).

internal/cmd/postgres/client.go
turip: I have dialed back the size of the refactor, given that what is the most problematic is the magic constants in the clickhouse options. In this [PR](https://github.com/openmeterio/openmeter/pull/1278) I have moved the options part to the config package (like we do for kafka) and then it's fine to initialize the clickhouse client in the main.

internal/cmd/clickhouse/client.go
turip: I took the approach to provide the options from the config package. That's aligned with what we do for kafka.

PR #1277, opened 2024-07-30
fix: allow getting entitlements by feature id

turip:
## Overview

The filtering was bugous and used the entitlement ID for searches instead of the featureID.


GAlexIHU:
Nope its not bogus, that is the intended behavior, that's why the name of the argument is `idOrFeatureKey` not `featureIdOrKey`

turip:
🥇

PR #1278, opened 2024-07-30
feat: add entitlements initialization - v2

turip:
## Overview

This patch ensures that entitlements are initialized for the balance workers.

This patch moves the clickhouse initialization's options part to the config package, as we have some magic constants there that should become config options.

It moves the entitlements initialization to a registry package, as at most of the places we are initializing credit, entitlements, and product catalog together: that's just a boilerplate we should not copy.

This is based on feedback from https://github.com/openmeterio/openmeter/pull/1274

### DIFF
--- a/cmd/balance-worker/main.go
+++ b/cmd/balance-worker/main.go
@@ -11,8 +11,10 @@ import (
 	"syscall"
 	"time"
 
+	"entgo.io/ent/dialect/sql"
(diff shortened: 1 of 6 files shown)


GAlexIHU (approved):
👍

sagikazarmark (approved):

PR #1280, opened 2024-07-31
Feat/batch ingest events for balance calculation

GAlexIHU:
## Overview

Fixes #(issue)


GAlexIHU:
DLQ should have maxretries implemented

internal/entitlement/balanceworker/worker.go
```diff
@@ -65,9 +67,14 @@ type highWatermarkCacheEntry struct {
...
 
+type connectors struct {
+	entitlement *registry.Entitlement
+	repo        BalanceWorkerRepository
+}
+
 type Worker struct {
 	opts       WorkerOptions
-	connectors *registry.Entitlement
+	connectors connectors
```
turip, line 77:
```suggestion
	entitlement *registry.Entitlement
	repo BalanceWorkerRepository
```

I don't think we need another level of grouping here, WDYT?

internal/entitlement/postgresadapter/entitlement.go
```diff
@@ -26,7 +28,12 @@ type entitlementDBAdapter struct {
 	db *db.Client
 }
 
-func NewPostgresEntitlementRepo(db *db.Client) entitlement.EntitlementRepo {
+type repo interface {
```
turip, line 31: maybe not, but I would prefer to not have this interface here depending on entitlement + balanceworker.

Instead, we should not have interface constraints here, and whenever we want to use the return value of NewPostgresEntitlementRepo the checking should happen there.

That would hugely decrease the dependency hell we have between the packages.

This interface constraining also necessitates the splitting of registry/builder just because we are using the repo directly inside the balanceworker.

internal/event/spec/parser.go
```diff
@@ -64,9 +64,9 @@ func newCloudEventFromSpec(meta *EventTypeSpec, spec EventSpec) event.Event {
 	return ev
 }
 
-// ParseCloudEvent unmarshals a single CloudEvent into the given payload
+// ParseCloudEvent unmarshals and validates a single CloudEvent into the given payload
```
turip, line 67: <3

internal/watermill/middleware/dlq.go
```diff
@@ -0,0 +1,96 @@
...
+	msg.Metadata.Set(DLQErrorReasonKey, err.Error())
+	msg.Metadata.Set(DLQOriginalTopicKey, message.SubscribeTopicFromCtx(msg.Context()))
+
+	// Don't intercept error from publish. Can't help you if the publisher is down as well.
+	return dlq.pub.Publish(dlq.topic, msg)
+}
+
+func (dlq deadLetterQueue) Middleware(h message.HandlerFunc) message.HandlerFunc {
+	// Capture return values from the handler
+	return func(msg *message.Message) (events []*message.Message, err error) {
```
turip, line 74: Please don't use named returns: Named returns only make sense if you are trying to manipulate return values in a defer when using recover, but panic handling is handled by other middleware.

Please don't use defer as this code can be replaced by:

```
func(msg * ...)
  outMessages, err := h(msg)
  
 // code in defer
```

internal/watermill/middleware/dlq.go
```diff
@@ -0,0 +1,96 @@
...
+	DLQOriginalTopicKey = "original_topic"
+)
+
+// DLQMsgProcessorFunc determines whether any messages should be published to the DLQ based on the original message and the error.
+// Even if DLQMsgProcessorFunc returns a non-nil slice of messages, if error is nil no messages will be published to the DLQ.
+type DLQMsgProcessorFunc func(original *message.Message, err error) []*message.Message
+
+// defaultProcessor listens for RetryableErrors and publishes the messages contained in that
+var defaultProcessor DLQMsgProcessorFunc = func(original *message.Message, err error) []*message.Message {
+	if retryable, ok := err.(watermill.RetryableError); ok {
```
turip, line 24: Please use the errorsx.ErrorAs (or errors.As), this will not work properly if any wrapping occurs.

internal/entitlement/balanceworker/worker.go
```diff
@@ -121,30 +128,30 @@ func New(opts WorkerOptions) (*Worker, error) {
 		middleware.Recoverer,
 	)
 
-	if opts.PoisonQueue != nil {
-		poisionQueue, err := middleware.PoisonQueue(opts.Publisher, opts.PoisonQueue.Topic)
+	if opts.DLQ != nil {
+		dlq, err := wmmiddleware.DLQ(opts.Publisher, opts.DLQ.Topic, nil)
```
turip, line 132: We should use the poison queue implementation from Watermill and add additional middleware that does the splitting in this specific case.

I don't want to maintain a full Dlq implementation when the library provides one.

From the Schemantics perspective I would expect 3 cases from the PoisionQueue + DLQ implementation:
- If an error happens, and it is a known non-retriable error => drop from DLQ
- If an ingest split is needed, then let's only DLQ the batch items for failing entries
- Else DLQ

internal/entitlement/balanceworker/ingesthandler.go
```diff
@@ -4,38 +4,73 @@ import (
...
+			continue
 		}
 
 		result = append(result, messages...)
 	}
 
-	return result, nil
+	var err error
+	if len(failedEntitlementInfo) > 0 {
+		err = w.newEntitlementProcessingFailedError(failedEntitlementInfo, handlingError)
```
turip, line 66: After thinking about this PR a bit, instead of an abstract solution, I would rather include the splitting here, as right now we don't need this anywhere else. If we need the same logic somewhere else we can abstract this, but I wouldn't do it just now.

internal/entitlement/postgresadapter/entitlement.go
GAlexIHU: I think registry & builder being combined was the issue actually. If you think about the direction of dependencies, implementations depend on abstractions so it's only right that we depend here on the interfaces. The issue was that registry depended in the both & then it was used in packages defining further interfaces we wanted to implement. 

I'm ok with using two type asserts instead of `repo` but IMO that's a technicality, but i'm confident this is the correct direction of dependencies.

internal/watermill/middleware/dlq.go
GAlexIHU: I'm OK with the change, but thats a stylistic preference / convention

(technically the two [arent't the same](https://go.dev/play/p/_LksNUvLsOl) but we don't need that here (nor can we use it due to the nil check a few lines up in `publishWithErr`)

internal/entitlement/balanceworker/worker.go
GAlexIHU: I don't understand what you mean by not wanting to maintain a full DLQ implementation, as you see its roughly 50 lines split into two methods

About the expected behavior:
- the main difference between the current implementation and watermill's poisonqueu is that the current implementation only DLQs "known retryable" errors
- i see the rational behind managing it as an opt out rather than an opt in, so we could change the default to something like
```go
// defaultProcessor listens for RetryableErrors and publishes the messages contained in that
var defaultProcessor DLQMsgProcessorFunc = func(original *message.Message, err error) []*message.Message {
	if retryable, ok := errorsx.ErrorAs[watermill.RetryableError](err); ok {
		return retryable.RetryMessages()
	}

	return []*message.Message{original}
}
```
which i think gives best of both

thoughts?

internal/entitlement/balanceworker/worker.go
GAlexIHU: i think i prefer the word "connectors" as `w.connectors.entitlement` reads better to me than `w.entitlement` simply semantically, but I'm fine with either

internal/entitlement/postgresadapter/entitlement.go
turip: We can talk about this IRL, but I don't think that this interface should be mandated here. Right now this is how it is, I was just signaling that this is not something I am confident is a good approach.

internal/entitlement/balanceworker/worker.go
turip: My toughts are that the PoisionQueue implementation is perfect in watermill. The only thing we need is such a handler for this specific use-case that handles the specific case of ingress batch split.

internal/entitlement/balanceworker/worker.go
turip: I prefer to not to have repos in a struct called connectors ;-)

internal/watermill/middleware/dlq.go
GAlexIHU: on a more philosophical note, if we rely on the error data/methods for our behavior we should probably not include wrapped errors, but I don't realistically see it turning into an issue

PR #1285, opened 2024-08-01
fix: exclude deleted entitlements from autoreset

GAlexIHU:
## Overview

Deleted entitlements weren't excluded from the autoreset which resulted in frequent errors


turip (approved):
Nice!

tothandras (approved):

PR #1287, opened 2024-08-02
Add getEntitlementById to public API & fix pagination bugs OM-857

GAlexIHU:
## Overview

- Adds new EP in OpenAPI & Clients
- Fixes pagination validation
- Sets pagination defaults

Fixes #OM-857


internal/entitlement/httpdriver/entitlement.go
```diff
@@ -328,6 +329,10 @@ func (h *entitlementHandler) ListEntitlements() ListEntitlementsHandler {
 					return entitlement.EntitlementType(s)
 				}),
 			}
+			if !p.Page.IsZero() {
```
turip, line 332: I don't know how many other places we have the same optional structure, but we can consider creating a helper that takes a pageNumber pointer and pageSize pointer and constructs the page from that.

turip (approved):

internal/entitlement/httpdriver/entitlement.go
GAlexIHU: We could do a helper for it, the reason this is separate is because we also have to support limit & offset

We could do a helper method that takes the limit, offset, page, pageSize pointers and constructs the output....

GAlexIHU: In practice we should never reimplement this, its only there to support the legacy limit/offset pagination as well

PR #1289, opened 2024-08-02
Add custom retry message processing for DLQ

GAlexIHU:
## Overview

Let's continue the DLQ implementation discussion & changes over here (branched form https://github.com/openmeterio/openmeter/pull/1280) to unblock the ingest event batching

### DIFF
--- a/internal/entitlement/balanceworker/ingesthandler.go
+++ b/internal/entitlement/balanceworker/ingesthandler.go
@@ -10,6 +10,7 @@ import (
 	"github.com/openmeterio/openmeter/internal/event/spec"
 	"github.com/openmeterio/openmeter/internal/productcatalog"
 	"github.com/openmeterio/openmeter/internal/sink/flushhandler/ingestnotification"
+	"github.com/openmeterio/openmeter/internal/watermill"
 	"github.com/openmeterio/openmeter/pkg/slicesx"
 )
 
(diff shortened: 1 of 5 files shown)


pkg/slicesx/chunk.go
```diff
@@ -0,0 +1,28 @@
...
+
+	var chunks [][]T
+	chunk := make([]T, 0, size)
+
+	for i, item := range s {
+		chunk = append(chunk, item)
+
+		if len(chunk) == size || i == len(s)-1 {
+			chunks = append(chunks, chunk)
+			chunk = make([]T, 0, size)
```
turip, line 23: Please add back the filtering for the I == len(s)-1 case, so that we are not allocating unnecesarily at the end of the iteration.

internal/watermill/middleware/dlq.go
```diff
@@ -0,0 +1,97 @@
...
+type deadLetterQueue struct {
+	topic string
+	pub   message.Publisher
+
+	msgProcessor DLQMsgProcessorFunc
+}
+
+// DLQ provides a middleware that salvages unprocessable messages and published them on a separate topic.
+// The main middleware chain then continues on, business as usual.
+func DLQ(pub message.Publisher, topic string, msgProcessor DLQMsgProcessorFunc) (message.HandlerMiddleware, error) {
```
turip, line 41: I don't think it's worth creating a DLQ implementation for this. I would rather have the following logic in the event handler:

```
- try to calculate the entitlements
- collect the successfully and unsuccessfully generated entitlements
- publish the successfully generated snapshots
- publish a EventBatchedIngest event with the remainder into the dlq topic
```

@sagikazarmark what do you think?

@GAlexIHU Given that I did a lot of refactoring around the handlers, I can take over this PR and wrap it up if you are fine with this.

internal/entitlement/balanceworker/ingesthandler.go
```diff
@@ -4,38 +4,73 @@ import (
...
 		messages, err := w.handleEntitlementUpdateEvent(
 			ctx,
-			entitlement,
+			NamespacedID{Namespace: entitlement.Namespace, ID: entitlement.EntitlementID},
 			spec.ComposeResourcePath(entitlement.Namespace, spec.EntityEvent),
 		)
 		if err != nil {
-			return nil, err
+			// TODO: add error information too
+			handlingError = multierror.Append(handlingError, err)
```
sagikazarmark, line 56: Can we use `errors.Join` instead?

internal/watermill/middleware/dlq.go
sagikazarmark: Sounds good to me, although I don't have a lot of context.

What you are suggesting sounds like "business logic" to me. DLQ is generally a good approach for blindly retrying without clogging up the original queue.

The above proposal sounds like a more "clever" retry approach where we only retry what needs to be retried. That sounds like business logic to me and DLQ may not be the best approach there.

GAlexIHU: @turip I'm okay with you taking over, I've left this PR stale for the past some time no not much lost there

on the DLQ implementation topic, I think having a generic `RetryableError` or something of the sort will only make more sense as we go forward, but I'm fine either way for now

GAlexIHU:
Since redundant, supersede based on verdict of this discussion https://github.com/openmeterio/openmeter/pull/1280#discussion_r1699994899

Closing due to inactivity

PR #1314, opened 2024-08-07
refactor: use watermill event bus

turip:
## Overview

This patch introduces watermill [event bus](https://github.com/ThreeDotsLabs/watermill/blob/master/components/cqrs/event_bus.go#L68) usage for internal event publishing. This allows us to have a cleaner way of publishing messages, by eliminating publishing boilerplate and by making the eventbus to handle topic <-> message assignments.


internal/entitlement/balanceworker/worker.go
```diff
@@ -169,78 +169,88 @@ func (w *Worker) Close() error {
...
 		w.opts.Logger.Warn("missing CloudEvents type, ignoring message")
 		return nil, nil
 	}
 
 	switch ceType {
 	// Entitlement events
-	case entitlement.EntitlementCreatedEvent{}.Spec().Type():
-		event, err := spec.ParseCloudEventFromBytes[entitlement.EntitlementCreatedEvent](msg.Payload)
-		if err != nil {
+	case entitlement.EntitlementCreatedEvent{}.EventName():
```
turip, line 180: The subscriber/consumer part is intentionally left the old way, as I don't want to further complicate this PR, it will be a separate one.

internal/entitlement/events.go
```diff
@@ -39,32 +35,54 @@ func (e entitlementEvent) Validate() error {
...
+	_ marshaler.Event = EntitlementCreatedEvent{}
 
-func (e EntitlementCreatedEvent) Spec() *spec.EventTypeSpec {
-	return &entitlementCreatedEventSpec
-}
+	entitlementCreatedEventName = metadata.GetEventName(metadata.EventType{
+		Subsystem: EventSubsystem,
+		Name:      "entitlement.created",
+		Version:   "v1",
+	})
```
sagikazarmark, line 45: I still think these should be reduced to a single string identifier, but it's not a big deal for now.

sagikazarmark (approved):

sagikazarmark (approved):

PR #1316, opened 2024-08-07
refactor: subscribers

turip:
## Overview

This patch adopts watermill fully for subscribers/consumers.

There are two design decisions that I had made:
- We are using the NoPublishing handlers
- We are not using the EventGroupProcessor

## NoPublishing handlers

The whole implementation was based on the idea to use the [EventGroupProcessor](https://github.com/ThreeDotsLabs/watermill/blob/c77ca102e5cdfbfc85bbef5aa073364d9cc1d656/components/cqrs/event_processor_group.go#L115) so that we can use handlers and generics to find the type of the message.

The NoPublishing handlers are used because:
- Most of our consumers will not publish messages using the [Router](https://watermill.io/docs/messages-router/) component. 
- The EventGroupProcessor assumes a NoPublishing handler due to the Command and Query Responsibility Segregation (CQRS) abstraction.

The only issue is when we want to use the Router as it was intended: to transform messages. But:
-  router doesn't support the EventBus (so we would need to use the lower level publishing interface)
- router can only publish to a specific topic, and even they [are recommending](https://github.com/ThreeDotsLabs/watermill/blob/master/message/router.go#L250-L251) to manually publish to support more topic

## Not using the EventGroupProcessor

One event group processor can consume from a single topic. In the case of balance worker we want to consume from the system events and the ingest events topic, which might match. For this we might or might not need to create two EventGroupProcessors.

For DLQ we would need to have a separate handler, as eventGroupProcessor uses a different abstraction.

Due to this, the implementation contains a minimalistic handler suite, that relies on the eventGroupProcessor's abstractions, but can work as a NonPublishingHandler, so that we can utilize the router object fully.

## Default consumer implementation

A default consumer setup is added to promote code reuse. If any consumer would require a more complex setup, we can always create specific consumer setups.


internal/watermill/router/router.go
```diff
@@ -0,0 +1,105 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		router.AddMiddleware(
+			poisionQueue,
+		)
+
+		poisionQueueProcessor := nopublisher.NoPublisherHandlerToHandlerFunc(dlqHandler)
+		if opts.DLQ.Throttle.Enabled {
```
turip, line 91: Let's add the following:
- DLQ permanent error after some time

sagikazarmark (approved):

PR #1325, opened 2024-08-09
fix: relax entitlement snapshot event validation

turip:
## Overview

Static entitlements doesn't have current period set, so let's not mandate that.

This patch also renames balance to value, as the previous naming was misleading.


GAlexIHU (commented):
I mean... static entitlements wouldn't have balance either, right? How is `e.Operation` set, couldn't that be the issue?

turip:
> I mean... static entitlements wouldn't have balance either, right? How is `e.Operation` set, couldn't that be the issue?

Yes, I think I have used a bad naming here, this is actually an entitlement Value so I have renamed the field to reflect that. The issue was a static entitlement, where I made the assumption that the period is set.

Either ways, you are also right regarding the Operation: we should validate that this is properly filled.

GAlexIHU (approved):
👍

PR #1344, opened 2024-08-12
Implement usage carryover

GAlexIHU:
## Overview

Adds backward compatible API for preserving overage at a reset (both manual and scheduled).
Overage at the end of a period has already been calculated, this PR simply adds the calculated value to the snapshot marking the reset (which is sufficient as the engine is already equipped to handle snapshots with overage).
To achieve this some `engine` internals had to be restructured and exposed from the package, it might make sense at some point to slightly rewrite the engine to follow a more functional style.

## Note:
Previously grants with negative values have been allowed and they resulted in a non-defined behavior. Creating them now fails validation while any previous such grants are simply ignored by the calculation logic going forward.


turip (commented):
Overall, this makes sense, but I don't know what is the exact user story we are aiming for. 

If the overage should only be carried over in case of manual resets, I think this PR is fine.

However, I don't know what is expected in the case of autoresets? Do we expect the autoreset to always not carry over the overage? 

From that perspective we should either agree on the default behavior for auto resets or make the reset behavior part of the entitlement itself. Then the autoreset job can choose the right overage behavior.

The existing field in the reset request can still serve as a manual reset override of the behavior defined in the entitlement itself.

@hekike WDYT?

GAlexIHU:
> However, I don't know what is expected in the case of autoresets? Do we expect the autoreset to always not carry over the overage?
> 
> From that perspective we should either agree on the default behavior for auto resets or make the reset behavior part of the entitlement itself. Then the autoreset job can choose the right overage behavior.
> 
> The existing field in the reset request can still serve as a manual reset override of the behavior defined in the entitlement itself.

If you check that's exactly what it does, the default behavior is defined on the entitlement and it can be overriden when the user executes a manual reset
https://github.com/openmeterio/openmeter/blob/feat/usage-carriover/internal/entitlement/metered/reset.go#L40

GAlexIHU:
> Overall, this makes sense, but I don't know what is the exact user story we are aiming for.

For the user story, you can check [this PR](https://github.com/openmeterio/landing/pull/155), but TLDR: we're in a sense imitating the running balance of a `usagePeriod`-less entitlement while circumventing data retention limits.

turip (approved):
Sorry, my bad. I have missed that part somehow :/

PR #1347, opened 2024-08-12
Add CI checks in separate PR

GAlexIHU:
### Adds checks to validate the integrity of migrations:
1. Simple up-down-up on empty database (part of `go test ...`)
2. Validates that ent schema was generated (`internal/ent/db` in sync with `internal/ent/schema`)
3. Validates that migrations directory is in sync with schema (no diff)
4. Runs `atlas lint` for last 10 migration files checking for migrations being non-destructive and linear (10 is arbitrary)
5. Validates atlas checksum integrity (migration files weren't compromised)

The checks can also be run locally via `make migrate-check`

### Fixes the atlas nix derivation to use platform specific version
Note: it's templated locally, not the ideal solution but I have some skill issues with nix

## Follow Up Items
- We should test that migrations work not just up-down-up but with a base of the latest release

## Notes for reviewer
The checks are implemented in dagger, and for dependency management a nix-in-dagger solution was used. This is somewhat questionable, the 2 main drawbacks are:
- doesn't really fit in the conceptual paradigm of dagger
- nix core + devenv deps have to be pulled no matter how lightweight the environment is otherwise

(this latter is decently mitigated by caching but still)

The benefit form my point of view is that it's an exact replica the local environment by using the same derivation(s)

### DIFF
--- a/ci/migrate.go
+++ b/ci/migrate.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/openmeterio/openmeter/ci/internal/dagger"
 	"github.com/sourcegraph/conc/pool"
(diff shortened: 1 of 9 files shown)


ci/migrate.go
```diff
@@ -0,0 +1,57 @@
...
+	// Always lint last 10 migrations
+	p.Go(syncFunc(nix.
+		WithServiceBinding("postgres", postgresNamed("last-10")).
+		WithExec([]string{"sh", "-c", "nix develop --impure .#dagger -c atlas migrate --env ci lint --latest 10"}),
+	))
+	// Validate checksum is intact
+	p.Go(syncFunc(nix.
+		WithServiceBinding("postgres", postgresNamed("validate")).
+		WithExec([]string{"sh", "-c", "nix develop --impure .#dagger -c atlas migrate --env ci validate"}),
+	))
```
sagikazarmark, line 46: I would probably just use this image: https://hub.docker.com/r/arigaio/atlas

ci/migrate.go
```diff
@@ -0,0 +1,57 @@
...
+
+func (m *Migrate) Check(
+	ctx context.Context,
+) error {
+	nix := nix(m.Source)
+	p := pool.New().WithErrors().WithContext(ctx)
+
+	// Always validate schema is generated
+	p.Go(syncFunc(nix.
+		WithExec([]string{"sh", "-c", "nix develop --impure .#dagger -c go generate ./internal/ent/..."}).
```
sagikazarmark, line 28: You could use the go moule to generate the code and then extract the directory using `.Container().Directory()`.

ci/migrate.go
```diff
@@ -0,0 +1,57 @@
...
+		WithExec([]string{"sh", "-c", "nix develop --impure .#dagger -c sh -c 'if [[ -n $(git status --porcelain ./internal/ent) ]]; then echo \"Ent schema wasnt generated\"; exit 1; fi' "}),
```
sagikazarmark, line 29: You can either use Git here, but you could also try using `dagger.Directory.Diff` to see if there is a diff between the first and the last version. That would utilize caching more.

PR #1360, opened 2024-08-14
test: add integration test suite for Notification API

chrisgacsal:
## Overview

Add integration test suite for Notification API.

### DIFF
--- a/ci/e2e.go
+++ b/ci/e2e.go
@@ -79,3 +79,20 @@ func postgres() *dagger.Service {
 		WithExposedPort(5432).
 		AsService()
 }
+
+const (
+	SvixJWTSingingSecret = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpYXQiOjE3MjI5NzYyNzMsImV4cCI6MjAzODMzNjI3MywibmJmIjoxNzIyOTc2MjczLCJpc3MiOiJzdml4LXNlcnZlciIsInN1YiI6Im9yZ18yM3JiOFlkR3FNVDBxSXpwZ0d3ZFhmSGlyTXUifQ.PomP6JWRI62W5N4GtNdJm2h635Q5F54eij0J3BU-_Ds"
+)
+
+func svix() *dagger.Service {
+	return dag.Container().
+		From(fmt.Sprintf("svix/svix-server:%s", svixVersion)).
+		WithEnvVariable("WAIT_FOR", "true").
+		WithEnvVariable("SVIX_QUEUE_TYPE", "memory").
+		WithEnvVariable("SVIX_CACHE_TYPE", "memory").
+		WithEnvVariable("SVIX_DB_DSN", "postgres://postgres:postgres@postgres:5432/postgres?sslmode=disable").
+		WithEnvVariable("SVIX_JWT_SECRET", SvixJWTSingingSecret).
+		WithServiceBinding("postgres", postgres()).
+		WithExposedPort(8071).
+		AsService()
+}
--- a/test/notification/channel.go
+++ b/test/notification/channel.go
@@ -1,0 +1,168 @@
+package notification
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/openmeterio/openmeter/internal/notification"
+	notificationwebhook "github.com/openmeterio/openmeter/internal/notification/webhook"
+	"github.com/openmeterio/openmeter/pkg/models"
+)
+
+func NewCreateChannelInput(name string) notification.CreateChannelInput {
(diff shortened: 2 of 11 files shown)


turip (approved):

ci/e2e.go
```diff
@@ -79,3 +79,20 @@ func postgres() *dagger.Service {
...
 		AsService()
 }
+
+const (
+	SvixJWTSingingSecret = "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpYXQiOjE3MjI5NzYyNzMsImV4cCI6MjAzODMzNjI3MywibmJmIjoxNzIyOTc2MjczLCJpc3MiOiJzdml4LXNlcnZlciIsInN1YiI6Im9yZ18yM3JiOFlkR3FNVDBxSXpwZ0d3ZFhmSGlyTXUifQ.PomP6JWRI62W5N4GtNdJm2h635Q5F54eij0J3BU-_Ds"
+)
+
+func svix() *dagger.Service {
+	return dag.Container().
+		From(fmt.Sprintf("svix/svix-server")).
```
sagikazarmark, line 89: Should we pin the version? (I usually like pinning)

(If not, `fmt.Sprintf` is not necessary)

ci/e2e.go
```diff
@@ -79,3 +79,20 @@ func postgres() *dagger.Service {
...
+		WithEnvVariable("WAIT_FOR", "true").
+		WithEnvVariable("SVIX_QUEUE_TYPE", "memory").
+		WithEnvVariable("SVIX_CACHE_TYPE", "memory").
+		WithEnvVariable("SVIX_DB_DSN", "postgres://postgres:postgres@postgres:5432/postgres?sslmode=disable").
```
sagikazarmark, line 93: I guess this is ok for now, but I wonder if we should create a separate database in postgres for svix.

turip (approved):

test/notification/channel.go
```diff
@@ -0,0 +1,168 @@
+package notification
```
sagikazarmark, line 1: Any particular reason why these files aren't in `_test.go` files.

More importantly: why aren't they next to the code they are testing?

sagikazarmark: To clarify: if (for some reason unknown to me) it does make sense to have a separate test package, I'd still prefer having them closer to the original notification code.

I definitely don't like having a high level test package (not a huge fan of e2e either, but there is a good reason for that)

ci/e2e.go
chrisgacsal: We should. I jsut forgot to add the version here. Will fix it.

ci/e2e.go
chrisgacsal: I was thinking about it as well. My plan is to create separate schema in Postgres for svix so we do not need to run multiple instances of Postgres, but the data for OM and Svix would be still separated. Will implement this in a separate PR.

test/notification/channel.go
chrisgacsal: > Any particular reason why these files aren't in _test.go files.

There is a single test file called `notification_test.go` which includes all the tests covered by this package.
There are multiple test "suites" in dedicated files like `channel.go` which cannot be directly tested as they require inicialization. Initialization is done in `notification_test.go` and the test "suites" are run from there.

Not particualrly keen on this structure, but my focus was on having test coverage and so much how the tests are structured. But I personally prefer to have a discussion around how we write/run integration/e2e tests just to have some guidelines.

> More importantly: why aren't they next to the code they are testing?

Again, this is something we need to chat about. I perfer unit tests stored next to the code, but having separate integration and e2e test packages. But I ok with whatever structure the team happy with.

chrisgacsal: I prefer to keep it as is until there is a decision how we want to sturucture our tests. Afterwards I'll make this test package aligned with the guidlines.

turip (approved):

tothandras:
I agree with @sagikazarmark that tests files should be next to the business logic that're being tested.

test/notification/testenv.go
```diff
@@ -0,0 +1,174 @@
...
+	logger := slog.Default().WithGroup("notification")
+
+	pgClient, err := NewPGClient(fmt.Sprintf(PostgresURLTemplate, postgresHost))
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		if err != nil {
+			if err := pgClient.Close(); err != nil {
+				logger.Error("failed to close postgres client", slog.String("error", err.Error()))
```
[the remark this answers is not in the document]
chrisgacsal: Valid. This is a copy paste error as closing the client should be done here: https://github.com/openmeterio/openmeter/pull/1360/files/f7081a80479a140938967c887eed9c00e58c5b76#diff-958d95e79f8cca9469807f1e7833d5b839526e6cb9def0b0ca3bb94687f917d0R162-R164

PR #1363, opened 2024-08-14
feat: resolve full subject entity

turip:
## Overview

This patch allows for resolving the full subject entity. This mostly applies for cloud for now, but in case we are introducing entity based subject support into the opensource version it can be useful too.


chrisgacsal (dismissed):
LGTM!

chrisgacsal (approved):

PR #1366, opened 2024-08-14
Implement entitlement overrides OM738

GAlexIHU:
## Overview

Users need a 0 downtime option to change a subject's entitlements. This is especially important for plan upgrade/downgrade scenarios.

Fixes #OM738

## Notes for Reviewers
- Tx Handling is still messy, I don't want to convolute that refactor with this feature


internal/entitlement/connector.go
```diff
@@ -159,26 +160,82 @@ func (c *entitlementConnector) CreateEntitlement(ctx context.Context, input Crea
...
+	if feature.ID != oldEnt.FeatureID {
+		return nil, &models.GenericUserError{Message: "Feature in path and body do not match"}
+	}
+
+	// Do the override in TX
+	return entutils.StartAndRunTx(ctx, c.entitlementRepo, func(ctx context.Context, tx *entutils.TxDriver) (*Entitlement, error) {
+		ctx = entutils.NewTxContext(ctx, tx)
+
+		// Delete previous entitlement
+		// FIXME: we publish an event during this even if we fail later
```
turip, line 216: This should be fine for now, but I agree that we should revamp trns handling, and provide hooks for publishing only if the whole transaction seems ok to commit (it's not 100% but better than the current situation)

turip (approved):
👍

PR #1376, opened 2024-08-15
refactor(api): notification events sending

chrisgacsal:
## Overview

* remove endpoint for notification event sending
* add endpoint for testing notification rule by sending notification event with test data


internal/ent/schema/notification.go
```diff
@@ -138,6 +138,13 @@ func (NotificationEvent) Fields() []ent.Field {
 			SchemaType(map[string]string{
 				dialect.Postgres: "jsonb",
 			}),
+		field.String("annotations").
+			GoType(notification.Annotations{}).
+			ValueScanner(AnnotationsValueScanner).
```
turip, line 143: Do we need a custom value scanner for map[string]interface{} scanning?

turip (approved):

PR #1380, opened 2024-08-15
feat: add subject and feature filters

turip:
## Overview

This patch relies on PostgreSQL's jsonb support, so it's not portable at all, but allows to filter by annotation values.

Additional risk is that the GIN index can only be defined on certian PGSQL types such as jsonb, so we will have to check how the query planner copes with complex queries where we need to use the GIN and other indexes (such as SELECT by namespace, created_at and a json field).


chrisgacsal (approved):
Awesome work! :shipit:

PR #1392, opened 2024-08-19
refactor: add utils for Notification API

chrisgacsal:
## Overview

Add utils for working with types from Notification API.

### DIFF
--- a/internal/notification/utils.go
+++ b/internal/notification/utils.go
@@ -1,0 +1,50 @@
+package notification
+
+// ChannelTypes returns a set of ChannelType from Channel slice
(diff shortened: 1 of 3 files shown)


internal/notification/utils.go
```diff
@@ -0,0 +1,50 @@
...
+		if _, ok := seen[status.State]; ok {
+			continue
+		}
+
+		seen[status.State] = struct{}{}
+		types = append(types, status.State)
+	}
+
+	return types
+}
```
sagikazarmark, line 50: These functions would be good candidates for iterators once we upgrade to Go 1.23

sagikazarmark (dismissed):

PR #1397, opened 2024-08-20
refactor: postgres/ent driver initialization

chrisgacsal:
## Overview

* split `ent` and `postgres/pgx` driver initialization into separate packages (`entutils/driver` and `pgdriver`)
* use `pgxpool` to have a better control over client connection pool to database
* allow instrumenting database client with custom tracing providers
* deprecate previously used `GetPGDriver` helper for backward compatibility

### DIFF
--- a/cmd/balance-worker/main.go
+++ b/cmd/balance-worker/main.go
@@ -10,7 +10,6 @@ import (
 	"syscall"
 	"time"
 
-	"entgo.io/ent/dialect/sql"
 	health "github.com/AppsFlyer/go-sundheit"
 	healthhttp "github.com/AppsFlyer/go-sundheit/http"
 	"github.com/ClickHouse/clickhouse-go/v2"
@@ -30,7 +29,6 @@ import (
 	semconv "go.opentelemetry.io/otel/semconv/v1.17.0"
 
 	"github.com/openmeterio/openmeter/config"
-	"github.com/openmeterio/openmeter/internal/ent/db"
 	entitlementpgadapter "github.com/openmeterio/openmeter/internal/entitlement/adapter"
 	"github.com/openmeterio/openmeter/internal/entitlement/balanceworker"
 	"github.com/openmeterio/openmeter/internal/meter"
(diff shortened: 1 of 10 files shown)


cmd/balance-worker/main.go
```diff
@@ -211,14 +210,42 @@ func main() {
 	}
 
 	// Dependencies: postgresql
-	pgClients, err := initPGClients(conf.Postgres)
+
+	// Initialize Postgres driver
+	var postgresDriver *pgdriver.Driver
+	postgresDriver, err = pgdriver.NewPostgresDriver(conf.Postgres.URL,
```
sagikazarmark, line 216:
```suggestion
	postgresDriver, err = pgdriver.NewPostgresDriver(
	conf.Postgres.URL,
```

cmd/balance-worker/main.go
```diff
@@ -211,14 +210,42 @@ func main() {
 	}
 
 	// Dependencies: postgresql
-	pgClients, err := initPGClients(conf.Postgres)
+
+	// Initialize Postgres driver
+	var postgresDriver *pgdriver.Driver
```
sagikazarmark, line 215: Not entirely sure why this variable has to be initialized.

cmd/balance-worker/main.go
```diff
@@ -211,14 +210,42 @@ func main() {
...
+
+	// Initialize Ent driver
+	entPostgresDriver := entdriver.NewEntPostgresDriver(postgresDriver.DB())
+	defer func() {
+		if err = entPostgresDriver.Close(); err != nil {
+			logger.Error("failed to close ent driver", "error", err)
+		}
+	}()
+
+	entPostgresClient := entPostgresDriver.Client()
```
sagikazarmark, line 240: Isn't this just an ent client at this point? I mean, it's not postgres specific, is it?

pkg/framework/pgdriver/driver.go
```diff
@@ -0,0 +1,103 @@
...
+		},
+	}
+
+	for _, opt := range opts {
+		opt.apply(o)
+	}
+
+	if o.ctx == nil {
+		o.ctx = context.Background()
+	}
```
sagikazarmark, line 86: I don't think anything is wrong with making this a mandatory argument.

chrisgacsal: Yeah, I was a bit hesitant about having context.Context in constructorm but I agree that in this case it makes sense to enforce it.

cmd/balance-worker/main.go
chrisgacsal: I do not think it makes any difference, but will rename it.

cmd/balance-worker/main.go
chrisgacsal: I do not think ther eis anything wrong with preinitializing the variable. Sometimes it makes the code more readable, but I have no strong prefenrence so I'll update this to remove the initialization.

cmd/balance-worker/main.go
chrisgacsal: Fixed.

turip (approved):

PR #1408, opened 2024-08-22
Fix bug with recurrence

GAlexIHU:
## Overview

There was an issue in which if the recurrence time was the same as the creation date of the entitlement OR a reset time, we calculated that grant twice


chrisgacsal (dismissed):
Awesome! thanks for the fix.

chrisgacsal (approved):

PR #1409, opened 2024-08-22
fix: Notification API

chrisgacsal:
## Overview

Fix multiple issues in Notification API:

* webhook signing secret validation
* custom header not being set at notification channel creation
* balance threshold validation to allow >100% values for PERCENT type
* deleting notification channels assigned to active notification rules

### DIFF
--- a/internal/notification/webhook/webhook.go
+++ b/internal/notification/webhook/webhook.go
@@ -77,9 +77,8 @@ func (i CreateWebhookInput) Validate() error {
 	}
 
 	if i.Secret != nil {
-		secret, _ := strings.CutPrefix(*i.Secret, SigningSecretPrefix)
-		if _, err := base64.StdEncoding.DecodeString(secret); err != nil {
-			return errors.New("invalid secret: must be base64 encoded")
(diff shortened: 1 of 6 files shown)


internal/notification/webhook/webhook.go
```diff
@@ -262,3 +261,16 @@ func New(config Config) (Handler, error) {
 
 	return handler, nil
 }
+
+func ValidateSigningSecret(secret string) error {
+	s, _ := strings.CutPrefix(secret, SigningSecretPrefix)
+	if len(s) < 32 || len(s) > 100 {
+		return errors.New("secret length must be between 32 to 100 chars without the optional prefix")
```
tothandras, line 268:
```suggestion
		return errors.New("secret length must be between 32 to 100 characters without the optional prefix")
```

tothandras (approved):

internal/notification/webhook/webhook.go
chrisgacsal: Will fix this in a follow up PR.

PR #1431, opened 2024-08-26
refactor: remove internal

turip:
## Overview

Maintaining the openmeter mapping to internal was quite a burden: this patch moves all the packages from internal to openmeter.


sagikazarmark (approved):
LGTM :shipit:

PR #1433, opened 2024-08-26
fix: nil error on event creation

turip:
## Overview

This causes panics at certain places, let's enforce that the logger is passed in correctly.

Fixes #(issue)

### DIFF
--- a/cmd/notification-service/main.go
+++ b/cmd/notification-service/main.go
@@ -315,7 +315,7 @@ func main() {
 		Repository:       notificationRepo,
 		Webhook:          notificationWebhook,
 		FeatureConnector: entitlementConnRegistry.Feature,
-		Logger:           logger.WithGroup("notification"),
+		Logger:           logger.With(slog.String("subsystem", "notification")),
 	})
 	if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/notification-service/main.go
```diff
@@ -315,6 +315,7 @@ func main() {
 		Repository:       notificationRepo,
 		Webhook:          notificationWebhook,
 		FeatureConnector: entitlementConnRegistry.Feature,
+		Logger:           logger.WithGroup("notification"),
```
sagikazarmark, line 318: Note that this causes all attribute keys passed to this logger to be prefixed with `notification.`. I don't have all the context, but I'm not sure that's desirable.

We used `logger.With(slog.String("subsystem", "notification"))` in other places.

turip: Thank you for pointing out, I have changed the code accordingly.

sagikazarmark (approved):

cmd/notification-service/main.go
chrisgacsal: I used the `WithGroup` deliberately here. Will revert this change.

PR #1439, opened 2024-08-28
feat: add recalculation metrics

turip:
## Overview

We can use these metrics to determine the average time required to recalculate an entitlement in balance worker.

Fixes OM-877


GAlexIHU (approved):
lgtm, just one q: are you also building dashboards for all the metrics we're adding?

turip:
> lgtm, just one q: are you also building dashboards for all the metrics we're adding?

Yes, I want to start adding dashboards to datadog, just first I want to make sure that the required metrics are available.

PR #1448, opened 2024-08-30
refactor(ent): extensions

hekike:
Turn ent templates into extensions to make them sharable.


openmeter/ent/generate.go
```diff
@@ -1,3 +1,3 @@
 package entdb
 
-//go:generate go run -mod=mod entgo.io/ent/cmd/ent  generate --target ./db  --feature sql/upsert --feature sql/lock --feature sql/versioned-migration --template ../../pkg/framework/entutils/expose.tpl --template ../../pkg/framework/entutils/paginate.tpl ./schema
+//go:generate go run -mod=mod entc.go
```
chrisgacsal, line 3:
```suggestion
//go:generate go run -mod=readonly entc.go
```

chrisgacsal: Just to avoid automatic pkg updates during `go generate`.

chrisgacsal (approved):

pkg/framework/entutils/entexpose/entexpose.go
```diff
@@ -0,0 +1,26 @@
+package entexpose
```
GAlexIHU, line 1: nit: i'd keep `entexpose` and `entpaginate` in the same package, maybe simply `extensions`

GAlexIHU (approved):
Neat!

pkg/framework/entutils/entexpose/entexpose.go
hekike: I prefer them separate, as nit I'll skip on it

PR #1451, opened 2024-09-01
feat(pagination): add helpers

hekike:
Add helpers to:

- Parse query params to `Page`
- Create `PagedResponse` from existing with transforming items type

### DIFF
--- a/pkg/pagination/pagination.go
+++ b/pkg/pagination/pagination.go
@@ -75,12 +75,23 @@ type PagedResponse[T any] struct {
 	Items      []T  `json:"items"`
 }
 
-func NewPagedResponseWith[OUT any, IN any](resp PagedResponse[IN], items []OUT) PagedResponse[OUT] {
+// NewPagedResponse creates a new PagedResponse with the given page, totalCount and items.
+func NewPagedResponseWith[OUT any, IN any](resp PagedResponse[IN], m func(in IN) (OUT, error)) (PagedResponse[OUT], error) {
+	items := make([]OUT, 0, len(resp.Items))
+	for _, inItem := range resp.Items {
+		item, err := m(inItem)
+		if err != nil {
+			return PagedResponse[OUT]{}, err
+		}
+
+		items = append(items, item)
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/pagination/pagination.go
```diff
@@ -50,6 +75,25 @@ type PagedResponse[T any] struct {
 	Items      []T  `json:"items"`
 }
 
+// NewPagedResponse creates a new PagedResponse with the given page, totalCount and items.
+func NewPagedResponseWith[OUT any, IN any](resp PagedResponse[IN], m func(in IN) (OUT, error)) (PagedResponse[OUT], error) {
```
GAlexIHU, line 79: I'm not convinced we need this. e.g.

```go
apiRes := pagination.PagedResponse{
  Page: res.Page,
  TotalCount: res.TotalCount,
  Items: slicesx.Map(...)
}
```

looks fine to me...

pkg/pagination/pagination.go
```diff
@@ -20,6 +20,31 @@ type Page struct {
...
+func NewPage(pageNumber int, pageSize int) Page {
+	return Page{
+		PageSize:   pageSize,
+		PageNumber: pageNumber,
+	}
+}
+
+// NewPageFromRef creates a new Page from pointers to pageNumber and pageSize.
+// Useful for handling query parameters.
+func NewPageFromRef(pageNumber *int, pageSize *int) Page {
```
GAlexIHU, line 33: IMO if you want this utility you should move it to `pkg/framework/commonhttp`, it's an API specific thing we do

hekike: I agree, but this is true for `PaginatedResponse` true + `NewPageFromRef` doesn't enforce is a query param, we would need to make an interface for it.

pkg/pagination/pagination.go
hekike: The challenge is if that then we have to copy struct properties everywhere in router which is easy to forget + error prone if we add new properties to PagedResponse like hasNextPage

sagikazarmark (approved):

PR #1497, opened 2024-09-10
Meter event validation

sagikazarmark:
## Overview

Refactor event validation to make it usable in other places than the sink-worker.

### DIFF
--- a/openmeter/meter/validate.go
+++ b/openmeter/meter/validate.go
@@ -1,0 +1,68 @@
+package meter
+
+import (
+	"errors"
+	"fmt"
+	"strconv"
+
+	"github.com/cloudevents/sdk-go/v2/event"
+	"github.com/oliveagle/jsonpath"
+)
+
+// ValidateEvent validates an event against a meter.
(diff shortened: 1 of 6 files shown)


openmeter/meter/validate.go
```diff
@@ -0,0 +1,68 @@
...
+	if rawValue == nil {
+		return errors.New("event data value cannot be null")
+	}
+
+	// Aggregation specific value validation
+	switch meter.Aggregation {
+	// UNIQUE_COUNT aggregation requires string property value
+	case MeterAggregationUniqueCount:
+		switch rawValue.(type) {
+		case string, float64:
```
tothandras, line 42: `float64` is also acceptable? Based on the comment, it shouldn't be.

tothandras (changes requested):

openmeter/meter/validate.go
sagikazarmark: That part of the code is just moved somewhere else, it's not actually new. I'm not sure what's the correct behavior.

chrisgacsal (approved):
LGTM!

PR #1507, opened 2024-09-13
test: reuse examples in quickstart

sagikazarmark:
## Overview

This PR ensures that the examples in the quick start guide are the same used in tests to make sure the quick start guide works.

### DIFF
--- a/quickstart/README.md
+++ b/quickstart/README.md
@@ -26,67 +26,19 @@ docker compose up -d
 Ingest usage events in [CloudEvents](https://cloudevents.io/) format:
 
 ```sh
-curl -X POST http://localhost:8888/api/v1/events \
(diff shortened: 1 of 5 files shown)


hekike:
To reduce the impact of event body is not visible anymore in .md, we can link from README to:
https://openmeter.io/docs/get-started/quick-start

Where we have events.

sagikazarmark:
It's kind of the point though: use a single source of truth for those events and make sure changing them in one place propagates to everywhere else. So linking to the docs isn't any better either. The docs should have the same instructions.

chrisgacsal (approved):

GAlexIHU:
@hekike As an option, we could template the quickstart README with the JSONs if that sound better

PR #1523, opened 2024-09-16
feat: add invoicing database structure

turip:
## Overview

This patch contains the base entities (DB) required to start working on the invoicing stack.

Right now the target is manually added invoice line items, later extending into subscription based line items.

Stripe and OpenMeter provider configs are added to the scheme.


openmeter/ent/schema/billing.go
```diff
@@ -0,0 +1,358 @@
...
+				"postgres": "jsonb",
+			}),
+		field.Bool("default").
+			Default(false),
+	}
+}
+
+func (BillingProfile) Edges() []ent.Edge {
+	return []ent.Edge{
+		edge.To("billing_invoices", BillingInvoice.Type),
```
hekike, line 56: Please set DB field name. Ent default is weird.

openmeter/ent/schema/billing.go
```diff
@@ -0,0 +1,358 @@
...
+func (BillingProfile) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("key").
+			NotEmpty().
+			Immutable(),
+		field.String("provider_config").
+			GoType(provider.Configuration{}).
+			ValueScanner(ProviderConfigValueScanner).
+			SchemaType(map[string]string{
+				"postgres": "jsonb",
```
hekike, line 43: let's normalize config to enable migrations, indexing etc.

`workflow_auto_advance`, `workflow_due_after`, e.g.

turip: Added a new patch with the requested changes.

openmeter/ent/schema/billing.go
turip: That is not required, as the reference is to the id field. At the from side I have set the filed properly. If you check the SQL migrations you will see that there are no weird field names present.

openmeter/ent/schema/billing.go
```diff
@@ -0,0 +1,418 @@
+package schema
```
hekike, line 1: should we separate files per table?

openmeter/ent/schema/billing.go
```diff
@@ -0,0 +1,418 @@
...
+	return []ent.Mixin{
+		entutils.IDMixin{},
+		entutils.NamespaceMixin{},
+		entutils.TimeMixin{},
+	}
+}
+
+func (BillingProfile) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("key").
```
hekike, line 36: nit: wondering if we should create a key Ent mixin like we have for ID

openmeter/ent/schema/billing.go
```diff
@@ -0,0 +1,418 @@
...
+			NotEmpty().
+			Immutable(),
+		field.String("provider_config").
```
hekike, line 39: should we separate config per provider? So least we can check who has which provider setup

openmeter/ent/schema/billing.go
```diff
@@ -0,0 +1,418 @@
...
+		entutils.MetadataAnnotationsMixin{},
+	}
+}
+
+func (BillingInvoiceItem) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("invoice_id").
+			Optional().
+			SchemaType(map[string]string{
+				"postgres": "char(26)",
```
hekike, line 205: nit: same, mixin?

openmeter/ent/schema/billing.go
```diff
@@ -0,0 +1,418 @@
...
+			Immutable(),
+		field.String("billing_profile_id").
+			NotEmpty().
+			Immutable().
+			SchemaType(map[string]string{
+				"postgres": "char(26)",
+			}),
+		field.Time("voided_at").
+			Optional(),
+		field.String("currency").
```
hekike, line 292: we will use lot's of currencies. Mixin?

openmeter/ent/schema/billing.go
```diff
@@ -0,0 +1,418 @@
...
+		index.Fields("namespace", "id"),
+		index.Fields("namespace", "key"),
+		index.Fields("namespace", "customer_id"),
+		index.Fields("namespace", "due_date"),
+		index.Fields("namespace", "status"),
+	}
+}
+
+func (BillingInvoice) Edges() []ent.Edge {
+	return []ent.Edge{
```
hekike, line 340: TODO to reference customer entity

openmeter/ent/schema/billing.go
```diff
@@ -0,0 +1,418 @@
...
+		edge.From("billing_workflow_config", BillingWorkflowConfig.Type).
+			Ref("billing_invoices").
+			Field("workflow_config_id").
+			Unique().
+			Required(),
+		edge.To("billing_invoice_items", BillingInvoiceItem.Type),
+	}
+}
+
+type providerReferenceSerde[T any] struct {
```
hekike, line 356: Is this a generic helper?

hekike (approved):

openmeter/ent/schema/billing.go
turip: We usually do seperate file per service. That's why I haven't split it.

openmeter/ent/schema/billing.go
turip: We can, I will do it as a separate PR.

openmeter/billing/provider/config.go
```diff
@@ -0,0 +1,63 @@
...
+import "fmt"
+
+// Type specifies the provider used for billing
+type Type string
+
+const (
+	// TypeOpenMeter specifies the OpenMeter billing provider, which is a dummy billing provider mostly useful for testing and
+	// initial OpenMeter assessment
+	TypeOpenMeter Type = "openmeter"
+	// TypeStripe specifies the Stripe billing provider, which is a real billing provider that can be used in production
```
hekike, line 12: This is an invoice provider, no? Not billing

openmeter/ent/schema/billing.go
turip: Even with ent we can query the provider_config -> type, so I wouldn't do that.

openmeter/ent/schema/billing.go
turip: The invoice ID is used only once. I wouldn't want to create a parameterized mixin for ulid.

openmeter/ent/schema/billing.go
turip: This is used by the parser to extract the jsonb field based on the type field. I wouldn't abstract it further as the value scanner needs to be custom either ways.

openmeter/billing/provider/config.go
turip: The openmeter will be a billing provider that is dummy (e.g. simulates all the functions including payments etc). Right now we don't have the distinction between tax, payments and invoicing providers, and I am reluctant to include that abstraction until we have at least one such use-case.

openmeter/ent/schema/billing.go
turip: My concern with mixins for this is that in case of plans, I don't know if we would want to make that field immutable. With mixins you cannot fine-tune these things (e.g. change the currency in a draft plan).

PR #1525, opened 2024-09-16
chore(api): support style on query params in spec

tothandras:
Add TypeSpec patches to support `style` on `@query` decorators.
See: https://github.com/microsoft/typespec/issues/4440


sagikazarmark (approved):
I reviewed every bit of this PR and I can honestly say, I haven't seen a better crafted, more beautiful PR before.

Very well done!

PR #1529, opened 2024-09-17
Adds improved transaction handling

GAlexIHU:
## Overview

- Adds nested transaction support (SQL `SAVEPOINT`)
- Separate transaction handling from ent
- Add utilities to remove `entutils.TxUser[T]` and `entutils.TxCreator` from repo signatures (move to internals)

## Notes for reviewer

1. Reference implementation for a repository is inside the entitlements package
2. Reference implementation on how to use `transaction` package is in `credit/balance`
3. All changes are backwards compatible with previous transaction handling

### DIFF
--- a/openmeter/entitlement/metered/grant_owner_adapter.go
+++ b/openmeter/entitlement/metered/grant_owner_adapter.go
@@ -12,6 +12,7 @@ import (
 	"github.com/openmeterio/openmeter/openmeter/productcatalog/feature"
 	"github.com/openmeterio/openmeter/openmeter/streaming"
 	"github.com/openmeterio/openmeter/pkg/framework/entutils"
+	"github.com/openmeterio/openmeter/pkg/framework/transaction"
 	"github.com/openmeterio/openmeter/pkg/models"
 )
 
@@ -41,7 +42,7 @@ func NewEntitlementGrantOwnerAdapter(
 
 func (e *entitlementGrantOwner) GetMeter(ctx context.Context, owner grant.NamespacedOwner) (*grant.OwnerMeter, error) {
 	// get feature of entitlement
-	entitlement, err := getRepoMaybeInTx(ctx, e.entitlementRepo, e.entitlementRepo).GetEntitlement(ctx, owner.NamespacedID())
+	entitlement, err := e.entitlementRepo.GetEntitlement(ctx, owner.NamespacedID())
 	if err != nil {
 		e.logger.Debug(fmt.Sprintf("failed to get entitlement for owner %s in namespace %s: %s", string(owner.ID), owner.Namespace, err))
 		return nil, &grant.OwnerNotFoundError{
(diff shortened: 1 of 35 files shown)


openmeter/entitlement/metered/grant_owner_adapter.go
```diff
@@ -201,14 +209,18 @@ func (e *entitlementGrantOwner) updateEntitlementUsagePeriod(ctx context.Context
...
-// FIXME: this is a terrible hack using select for udpate...
-func (e *entitlementGrantOwner) LockOwnerForTx(ctx context.Context, tx *entutils.TxDriver, owner grant.NamespacedOwner) error {
-	return e.entitlementRepo.WithTx(ctx, tx).LockEntitlementForTx(ctx, owner.NamespacedID())
+func (e *entitlementGrantOwner) LockOwnerForTx(ctx context.Context, owner grant.NamespacedOwner) error {
+	// If we're not in a transaction this method has to fail
+	_, err := transaction.GetDriverFromContext(ctx)
+	if err != nil {
+		return fmt.Errorf("lock owner for tx must be called in a transaction: %w", err)
+	}
+	return e.locker.LockForCtx(ctx, owner.NamespacedID().ID)
```
turip, line 218: Why don't we lock the owner's record instead of relying on a separate lock table?

GAlexIHU: Because we might be updating the entitlement entity itself.

turip: AFAIK you can still update while having a select for update lock on the row.

turip:
```
postgres=# CREATE TABLE test (
postgres(#   id numeric,
postgres(#   value text);
CREATE TABLE

postgres=# insert into test values (1, 'a');
INSERT 0 1
postgres=# insert into test values (2, 'b');
INSERT 0 1

postgres=# BEGIN TRANSACTION ISOLATION LEVEL SERIALIZABLE;
BEGIN

postgres=*# SELECT  * FROM test where id = 1 FOR UPDATE;
 id | value
----+-------
  1 | a
(1 row)

postgres=*# update test set value = 'xxx' where id = 1;
UPDATE 1

postgres=*# commit;
COMMIT
postgres=#

```

GAlexIHU: Checking [the docs](https://www.postgresql.org/docs/current/explicit-locking.html#LOCKING-ROWS) you're right 

(though one small thing, SERIALIZABLE will make concurrent TXs error if the record changed since they started acquiring their locks)

Should you like me to revert to the prev implementation?

turip: I was testing with SERIALIZABLE, but we should not use that for transaction coordination.

openmeter/ent/schema/lock.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"entgo.io/ent"
+	"entgo.io/ent/dialect"
+	"entgo.io/ent/schema/field"
+	"entgo.io/ent/schema/index"
+
+	"github.com/openmeterio/openmeter/pkg/framework/entutils"
+)
+
+type LockTable struct {
```
turip, line 12: Please remove the table definition too. Otherwise looks good.

GAlexIHU: Thanks for catching, updated now

turip (approved):

PR #1537, opened 2024-09-18
Add entitlements APIs and resources to TypeSpec

GAlexIHU:
## Overview

Implements the entitlements related APIs in TypeSpec

## Notes for reviewer

TypeSpec doesn't cover 100% of OAPI3 which results in some differences in the generated output.
- Some previously unnamed interfaces had to be extracted to models
- TypeSpec doesn't support arbitrary `allOf` types so the `lastReset` prop on `getEntitlement`'s return value isn't on top level anymore


api/spec/src/entitlements/entitlements.tsp
```diff
@@ -0,0 +1,319 @@
...
+
+  usagePeriod?: RecurringPeriod;
+}
+
+/**
+ * Create inpurs for metered entitlement
+ */
+@friendlyName("EntitlementMeteredCreateInputs")
+model EntitlementMeteredCreateInputs {
+  ...OmitProperties<EntitlementCreateSharedFields, "usagePeriod">;
```
tothandras, line 128: If you're looking for `allOf`, you are limited to extend a single model, but it's available:

```suggestion
model EntitlementMeteredCreateInputs extend OmitProperties<EntitlementCreateSharedFields, "usagePeriod"> {
```

api/spec/src/entitlements/entitlements.tsp
```diff
@@ -0,0 +1,319 @@
...
+  isSoftLimit?: boolean = false;
+
+  /**
+   * Deprecated, ignored by the backend. Please use isSoftLimit instead; this field will be removed in the future.
+   */
+  @deprecated("Use isSoftLimit instead")
+  isUnlimited?: boolean = false;
+
+  usagePeriod: RecurringPeriod;
+  measureUsageFrom?: MeasureUsageFrom;
```
tothandras, line 143: Missing documentation.

api/spec/src/entitlements/entitlements.tsp
```diff
@@ -0,0 +1,319 @@
...
+   */
+  @minValue(0)
+  issueAfterReset?: float64;
+
+  /**
+   * Defines the grant priority for the default grant.
+   */
+  @minValue(1)
+  @maxValue(255)
+  issueAfterResetPriority?: uint8 = 1;
```
tothandras, line 159: Take a look at how this will be serialized. I always pick `integer` instead of the more fine-grained alternative for this reason.

api/spec/src/entitlements/entitlements.tsp
```diff
@@ -0,0 +1,319 @@
...
+   * If true, the overage is preserved at reset. If false, the usage is reset to 0.
+   */
+  preserveOverageAtReset?: boolean = false;
+}
+
+/**
+ * Create inputs for static entitlement
+ */
+@friendlyName("EntitlementStaticCreateInputs")
+model EntitlementStaticCreateInputs {
```
tothandras, line 171:
```suggestion
model EntitlementStaticCreateInputs extend EntitlementCreateSharedFields {
```

api/spec/src/entitlements/entitlements.tsp
```diff
@@ -0,0 +1,319 @@
...
+  @encode("json")
+  config: string;
+}
+
+/**
+ * Create inputs for boolean entitlement
+ */
+@friendlyName("EntitlementBooleanCreateInputs")
+model EntitlementBooleanCreateInputs {
+  ...EntitlementCreateSharedFields;
```
tothandras, line 188:
```suggestion
model EntitlementBooleanCreateInputs extend EntitlementCreateSharedFields {
```

api/spec/src/entitlements/entitlements.tsp
```diff
@@ -0,0 +1,319 @@
...
+
+  /**
+   * The identifier key unique to the subject
+   */
+  @example("customer-1")
+  subjectkey: Key;
+
+  /**
+   * The feature the subject is entitled to use.
+   * Either featureKey or featureId is required.
```
tothandras, line 225:
```suggestion
```

api/spec/src/entitlements/entitlements.tsp
```diff
@@ -0,0 +1,319 @@
...
+   */
+  @example("example-feature-key")
+  featureKey: Key;
+
+  /**
+   * The feature the subject is entitled to use.
+   * Either featureKey or featureId is required.
```
tothandras, line 232:
```suggestion
```

api/spec/src/entitlements/entitlements.tsp
```diff
@@ -0,0 +1,319 @@
...
+   */
+  @example("01ARZ3NDEKTSV4RRFFQ69G5FAV")
+  featureId: ULID;
+
+  currentUsagePeriod: Period;
```
tothandras, line 237: Missing docs

api/spec/src/entitlements/entitlements.tsp
```diff
@@ -0,0 +1,319 @@
...
+model EntitlementMeteredCalculatedFields {
+  /**
+   * The time the last reset happened.
+   */
+  @example(DateTime.fromISO("2023-01-01T01:01:01.001Z"))
+  @visibility("read")
+  lastReset: DateTime;
+
+  @visibility("read")
+  currentUsagePeriod: Period;
```
tothandras, line 253: Missing docs

api/spec/src/entitlements/entitlements.tsp
```diff
@@ -0,0 +1,319 @@
...
+@friendlyName("EntitlementMetered")
+model EntitlementMetered {
+  ...OmitProperties<
+    EntitlementMeteredCreateInputs,
+    "type" | "measureUsageFrom" | "metadata" | "usagePeriod" | "featureKey" | "featureId" | "currentUsagePeriod"
+  >;
+  ...OmitProperties<EntitlementMeteredCalculatedFields, "currentUsagePeriod">;
+  ...OmitProperties<EntitlementSharedFields, "type">;
+  type: EntitlementType.metered;
+}
```
tothandras, line 275:
```suggestion
@friendlyName("EntitlementMetered")
model EntitlementMetered extend EntitlementSharedFields {
  ...OmitProperties<
    EntitlementMeteredCreateInputs,
    "type" | "measureUsageFrom" | "metadata" | "usagePeriod" | "featureKey" | "featureId" | "currentUsagePeriod"
  >;
  ...OmitProperties<EntitlementMeteredCalculatedFields, "currentUsagePeriod">;
  type: EntitlementType.metered;
}
```

api/spec/src/entitlements/entitlements.tsp
```diff
@@ -0,0 +1,319 @@
...
+
+/**
+ * Entitlement template of a boolean entitlement.
+ */
+@friendlyName("EntitlementBoolean")
+model EntitlementBoolean extends EntitlementSharedFields {
+  ...OmitProperties<EntitlementBooleanCreateInputs, "type" | "metadata" | "usagePeriod" | "featureKey" | "featureId">;
```
tothandras, line 282: There is also PickProperties

api/spec/src/entitlements/entitlements.tsp
```diff
@@ -0,0 +1,319 @@
...
+  ...OmitProperties<EntitlementSharedFields, "type">;
```
tothandras, line 283:
```suggestion
```

api/spec/src/entitlements/entitlements.tsp
```diff
@@ -0,0 +1,319 @@
...
+  type: EntitlementType.boolean;
+}
+
+/**
+ * A static entitlement.
+ */
+@friendlyName("EntitlementStatic")
+model EntitlementStatic {
+  ...OmitProperties<EntitlementStaticCreateInputs, "type" | "metadata" | "usagePeriod" | "featureKey" | "featureId">;
+  ...OmitProperties<EntitlementSharedFields, "type">;
```
tothandras, line 293:
```suggestion
model EntitlementStatic extend EntitlementSharedFields {
  ...OmitProperties<EntitlementStaticCreateInputs, "type" | "metadata" | "usagePeriod" | "featureKey" | "featureId">;
```

api/spec/src/entitlements/entitlements.tsp
```diff
@@ -0,0 +1,319 @@
...
+@friendlyName("MeasureUsageFromEnum")
+enum MeasureUsageFromEnum {
+  #suppress "@openmeter/api-spec/casing" "Use existing values"
+  CURRENT_PERIOD_START: "CURRENT_PERIOD_START",
+  #suppress "@openmeter/api-spec/casing" "Use existing values"
+  NOW: "NOW",
+}
+
+@friendlyName("MeasureUsageFromTime")
+scalar MeasureUsageFromTime extends DateTime;
```
tothandras, line 309: You can probably just use `DateTime` without creating a new type.

api/spec/src/types.tsp
```diff
@@ -109,39 +132,54 @@ model Resource {
...
-  @summary("Week")
-  week: "week",
-
-  @summary("Month")
-  month: "month",
-
-  @summary("Year")
-  year: "year",
+@friendlyName("RecurringPeriodEnum")
+enum Interval {
```
tothandras, line 136: Please create new types, this'll mess with the new plan API.

api/spec/src/types.tsp
```diff
@@ -109,39 +132,54 @@ model Resource {
...
 /**
- * Time interval with a time unit and count.
+ * Recurring period with an interval and an anchor
  */
-@example(#{ unit: "month", count: 3 })
-@friendlyName("Interval")
-model Interval {
+@example(#{ interval: Interval.Day, anchor: DateTime.fromISO("2023-01-01T01:01:01.001Z") })
+@friendlyName("RecurringPeriod")
+model RecurringPeriod {
```
tothandras, line 152: Same. Also, we'll use `duration` (ISO 8601) in place of current `Interval` object.

api/spec/src/types.tsp
```diff
@@ -44,6 +44,15 @@ scalar DateTime extends utcDateTime;
...
 scalar Money extends string;
 
+/**
+ * Helper types to turn the type information (commonly used at creat and update scenarios) into resources.
+ *
+ * This way `Resource` doesn't have to be extended and the underlying type can be reused freely.
+ */
+model AsResource<T> extends Resource {
+  ...T;
+}
```
tothandras, line 54: It's not used anywhere as far as I see.

```suggestion
/**
 * Helper types to turn the type information (commonly used at creat and update scenarios) into resources.
 *
 * This way `Resource` doesn't have to be extended and the underlying type can be reused freely.
 */
model AsResource<T> extends Resource {
  ...T;
}
```

api/spec/src/entitlements/management.tsp
```diff
@@ -0,0 +1,356 @@
...
+  /**
+   * The period of the segment.
+   */
+  period: Period;
+
+  /**
+   * The total usage of the grant in the period.
+   */
+  @example(100)
+  @visibility("readonly")
```
tothandras, line 270:
```suggestion
  @visibility("read")
```

Please fix the others as well.

tothandras (dismissed):

api/spec/src/entitlements/entitlements.tsp
GAlexIHU: why do you think that's more desirable? TypeSpec (as languages usually are) is single inheritance. I'm using multiple types at multiple places... from my point of view I don't see why one should be more important than the rest

api/spec/src/entitlements/entitlements.tsp
GAlexIHU: The named type has the doc, same as in the current version. I can add docs for these too if you want.

api/spec/src/entitlements/entitlements.tsp
GAlexIHU: I don't understand your issue. This is the current version:
```yaml
issueAfterResetPriority:
  type: integer
  minimum: 1
  maximum: 255
  default: 1
  example: 1
  description: |
    Defines the grant priority for the default grant. If provided, issueAfterReset must have a value.
```

this is the TypeSpec serialized version:
```yaml
issueAfterResetPriority:
  type: integer
  format: uint8
  minimum: 1
  maximum: 255
  description: Defines the grant priority for the default grant.
  default: 1
```
format is the correct keyword to describe that limitation (we could even do away with max 255 as thats also included in the format)

api/spec/src/entitlements/entitlements.tsp
GAlexIHU: same question as above

api/spec/src/entitlements/entitlements.tsp
tothandras: Alright.

api/spec/src/entitlements/entitlements.tsp
tothandras: If it works, then ok, it was just a heads-up that sometimes the format isn't correct and a number could be emitted as a string.

api/spec/src/entitlements/entitlements.tsp
GAlexIHU: The docs say there's something analogous to `keyof` which is what i'd like ideally. (I have to exclude an entire subtype shared by multiple composits)

api/spec/src/entitlements/entitlements.tsp
tothandras: `allOf` and types can be reused in the underlying server, client code.

api/spec/src/entitlements/entitlements.tsp
GAlexIHU: The docs say there's something analogous to `keyof` which is what i'd like ideally. (I have to exclude an entire subtype shared by multiple composits)

api/spec/src/entitlements/entitlements.tsp
GAlexIHU: `EntitlementSharedFields ` and `EntitlementBooleanCreateInputs ` have different signatures, i think the current version is correct

api/spec/src/entitlements/entitlements.tsp
GAlexIHU: I only did it for the `friendlyName` as in the previous version it was also a named type.

api/spec/src/types.tsp
GAlexIHU: Does the new plan api exist in any way currently? I rewrite these as they were also added to the entitlement types which have different signatures.

I suggest we could rename `RecurringPeriodEnum`  to  `Interval`. That's a breaking change in the type name, but it would only affect new versions of the SDKs, and make the naming more consistent. WDYT?

api/spec/src/types.tsp
GAlexIHU: Yes, thanks, i'll delete it

I think we should talk about the Resource model separately: Its not what our resources look like so in my opinion it's quite misleading.

api/spec/src/entitlements/entitlements.tsp
```diff
@@ -116,30 +116,26 @@ model EntitlementCreateSharedFields {
    * Additional metadata for the feature.
    */
   metadata?: Metadata;
-
-  /**
-   * The usage period associated with the entitlement.
-   */
-  usagePeriod?: RecurringPeriod;
```
GAlexIHU: This belongs here

api/spec/src/cloud/main.tsp
```diff
@@ -115,4 +115,16 @@ namespace OpenMeterCloud.Entitlements {
...
+  @tag("Entitlements (Experimental)")
+  interface Features extends OpenMeter.Entitlements.Features {}
+
+  @route("/api/v1/grants")
+  @tag("Entitlements (Experimental)")
+  interface Grants extends OpenMeter.Entitlements.Grants {}
+
+  @route("/api/v1/subjects/{subjectIdOrKey}/entitlements")
+  @tag("Entitlements (Experimental)")
+  interface SubjectManagement extends OpenMeter.Entitlements.SubjectManagement {}
```
sagikazarmark, line 129: How about calling this "Subjects"? It's in a separate namespace, so it wouldn't overlap with the global "Subjects" interface.

(Note: not that it matters much, it won't appear in OpenAPI anyway)

api/spec/src/entitlements/entitlements.tsp
```diff
@@ -0,0 +1,321 @@
...
+   * If page is provided that takes precedence and the paginated response is returned.
+   */
+  @get
+  @operationId("listEntitlements")
+  @summary("List all entitlements")
+  list(
+    /**
+     * Filtering by multiple features.
+     *
+     * Usage: `?feature=feature-1&feature=feature-2`
```
sagikazarmark, line 38: Nit: is this actually used somewhere? Example generation could take care of this in documentation.

api/spec/src/entitlements/entitlements.tsp
```diff
@@ -0,0 +1,321 @@
...
+     * Filtering by multiple entitlement types.
+     *
+     * Usage: `?entitlementType=metered&entitlementType=boolean`
+     */
+    @query entitlementType?: EntitlementType[],
+
+    ...OpenMeter.QueryPagination,
+    ...OpenMeter.QueryLimitOffset,
+    ...OpenMeter.QueryOrdering<EntitlementOrderBy>,
+  ): Entitlement[] | OpenMeter.PaginatedResponse<Entitlement> | OpenMeter.CommonErrors;
```
sagikazarmark, line 59: Is this for backwards compatibility reasons (paginated vs non-paginated response)?

Can you please make a note if so?

api/spec/src/entitlements/entitlements.tsp
```diff
@@ -0,0 +1,337 @@
...
+
+  /**
+   * Get entitlement by id.
+   */
+  @get
+  @operationId("getEntitlementById")
+  get(
+    @path
+    entitlementId: ULID,
+  ): Entitlement | OpenMeter.CommonErrors | OpenMeter.NotFoundError;
```
sagikazarmark, line 69:
```suggestion
  ): Entitlement | OpenMeter.NotFoundError | OpenMeter.CommonErrors;
```


Nit: going from the more specific to the generic responses.

sagikazarmark (commented):
A general comment on naming:

- EntitlementCreateInputs -> CreateEntitlementRequest (or NewEntitlement)
- EntitlementStatic -> StaticEntitlement

These names feel more natural to me.

While I realize these are breaking changes for OpenAPI, due to friendly name we can make these changes in TypeSpec. (Can be delayed, but please create a task somewhere)

It would be interesting to see if we can use TypeSpec for better server side code generation.

api/spec/src/query.tsp
```diff
@@ -0,0 +1,127 @@
...
+using TypeSpec.Rest;
+using TypeSpec.OpenAPI;
+
+namespace OpenMeter;
+
+/**
+ * Pagination query params
+ */
+@friendlyName("Pagination")
+model QueryPagination {
```
sagikazarmark, line 15:
```suggestion
@friendlyName("Pagination")
model QueryPagination {
```

QueryPagination or Pagination?

sagikazarmark, line 15: Why the name difference? Backward compatibility?

api/spec/src/entitlements/entitlements.tsp
GAlexIHU: It is, in general everything tries to be as close to the current version as possible. I figured if I were to refactor/rename anything in this PR it would be terribly hard to validate that everything is as expected.

api/spec/src/entitlements/entitlements.tsp
GAlexIHU: I just copied the current docs everywhere

api/spec/src/query.tsp
sagikazarmark: If this model is solely used in queries, I think it's ok.

sagikazarmark (approved):

sagikazarmark (approved):

PR #1540, opened 2024-09-19
feat: add diff util

chrisgacsal:
## Overview

Add a simple helper for calculating the difference between two slices.

### DIFF
--- a/pkg/slicesx/diff.go
+++ b/pkg/slicesx/diff.go
@@ -1,0 +1,66 @@
+package slicesx
+
+import "github.com/samber/lo"
+
+type Diff[T comparable, S ~[]T] struct {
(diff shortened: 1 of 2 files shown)


pkg/diff/diff.go
```diff
@@ -0,0 +1,66 @@
+package diff
+
+import "github.com/samber/lo"
+
+type Diff[T comparable, S ~[]T] struct {
```
sagikazarmark, line 5: Could this be moved to slicesx?

turip (commented):
<3

pkg/diff/diff.go
chrisgacsal: Done.

turip (approved):

PR #1541, opened 2024-09-19
feat: invoice service

turip:
## Overview

Fixes #(issue)

PR #1544, opened 2024-09-20
feat(customer): crud api

hekike:
- put behind PG FF
- remove `key` from Customer
- separate `Resource` and `UniqueResource` entities
- add CRUD APIs
- fix update
- add repository tests


cmd/server/main.go
```diff
@@ -369,6 +371,32 @@ func main() {
 		})
 	}
 
+	// Initialize Customer
```
hekike, line 374: Do we need to put this behind feature flag?

api/spec/src/pagination.tsp
```diff
@@ -2,6 +2,10 @@ import "@typespec/http";
 
 using TypeSpec.Http;
 
+/**
+ * A paginated query.
+ */
+@friendlyName("queryPaginated")
```
sagikazarmark, line 8:
```suggestion
@friendlyName("PaginatedQuery")
```

api/spec/src/types.tsp
```diff
@@ -49,6 +49,14 @@ scalar Money extends string;
  */
 @friendlyName("Resource")
 model Resource {
+  /**
+   * A unique identifier for the customer.
```
sagikazarmark, line 53: Customer?

openmeter/customer/customer.go
```diff
@@ -23,6 +24,67 @@ type Customer struct {
 	External          *CustomerExternalMapping  `json:"external"`
 }
 
+// AsAPICustomer converts a Customer to an API Customer
+func (c Customer) AsAPICustomer() (api.Customer, error) {
```
sagikazarmark, line 28: Mapping like this belongs to `customer/customerdriver`

openmeter/customer/httpdriver/apimapping.go
```diff
@@ -0,0 +1,88 @@
+package httpdriver
```
sagikazarmark, line 1: customerdriver

openmeter/customer/repository/customer.go
```diff
@@ -4,18 +4,25 @@ import (
 	"context"
```
sagikazarmark, line 4: This should be a customeradapter package, not repository.

openmeter/customer/service.go
```diff
@@ -17,25 +17,26 @@ type CustomerService interface {
...
+func (c *ServiceConfig) Validate() error {
 	if c.Repository == nil {
 		return errors.New("repository is required")
 	}
 
 	return nil
 }
 
-func NewService(c Config) (Service, error) {
+func NewService(c ServiceConfig) (Service, error) {
```
sagikazarmark, line 39: Why not just pass the repository here?

api/spec/src/customer.tsp
```diff
@@ -1,32 +1,103 @@
+import "@typespec/http";
+import "@typespec/rest";
+import "@typespec/openapi3";
+
+import "./errors.tsp";
 import "./billing.tsp";
```
turip, line 6: Please remove the billing providers from the customer object.

api/spec/src/types.tsp
```diff
@@ -171,7 +187,7 @@ scalar CountryCode extends string;
  */
 @friendlyName("Address")
 model Address {
-  country: CountryCode;
+  country?: CountryCode;
```
turip, line 222: I would make at least the country mandatory. That would help a lot with billing-related decisions. (e.g. Tax country,etc)

cmd/server/main.go
```diff
@@ -369,6 +371,35 @@ func main() {
...
+		if err != nil {
+			logger.Error("failed to initialize customer service", "error", err)
+			os.Exit(1)
+		}
+		defer func() {
+			if err = customerService.Close(); err != nil {
+				logger.Error("failed to close customer service", "error", err)
+			}
+		}()
+	}
```
turip, line 400: Given that we are not doing anything in the Close() (notifications required that for it's background queues), I would just get rid of Close() altogether.

openmeter/customer/customer.go
```diff
@@ -23,6 +24,66 @@ type Customer struct {
...
+	}
+
+	if c.External != nil {
+		external := api.CustomerExternalMapping{}
+
+		if c.External.StripeCustomerID != nil {
+			external.StripeCustomerId = c.External.StripeCustomerID
+		}
+
+		customer.External = &external
```
turip, line 61:
```suggestion
		customer.External = &api.CustomerExternalMapping{
		  StripeCustomerId: c.External.StripeCustomerID
		 }
```

openmeter/customer/customer.go
```diff
@@ -23,6 +24,66 @@ type Customer struct {
...
+			State:       c.BillingAddress.State,
+			PostalCode:  c.BillingAddress.PostalCode,
+			Line1:       c.BillingAddress.Line1,
+			Line2:       c.BillingAddress.Line2,
+			PhoneNumber: c.BillingAddress.PhoneNumber,
+		}
+
+		if c.BillingAddress.Country != nil {
+			country := string(*c.BillingAddress.Country)
+			address.Country = &country
```
turip, line 48:
```suggestion
			address.Country = lo.ToPtr(string(*c.BillingAddress.Country))
```

openmeter/customer/httpdriver/apimapping.go
```diff
@@ -0,0 +1,87 @@
...
+			State:       apiCustomer.BillingAddress.State,
+			PostalCode:  apiCustomer.BillingAddress.PostalCode,
+			Line1:       apiCustomer.BillingAddress.Line1,
+			Line2:       apiCustomer.BillingAddress.Line2,
+			PhoneNumber: apiCustomer.BillingAddress.PhoneNumber,
+		}
+
+		if apiCustomer.BillingAddress.Country != nil {
+			country := models.CountryCode(*apiCustomer.BillingAddress.Country)
+			address.Country = &country
```
turip, line 50:
```suggestion
			address.Country = lo.ToPtr(models.CountryCode(*apiCustomer.BillingAddress.Country))
```

openmeter/customer/httpdriver/customer.go
```diff
@@ -0,0 +1,235 @@
...
+	return httptransport.NewHandlerWithArgs(
+		func(ctx context.Context, r *http.Request, params ListCustomersParams) (ListCustomersRequest, error) {
+			ns, err := h.resolveNamespace(ctx)
+			if err != nil {
+				return ListCustomersRequest{}, fmt.Errorf("failed to resolve namespace: %w", err)
+			}
+
+			req := ListCustomersRequest{
+				Namespace:      ns,
+				IncludeDeleted: defaultx.WithDefault(params.IncludeDeleted, customer.IncludeDeleted),
```
turip, line 34:
```suggestion
				IncludeDeleted: lo.FromPtrOr(params.IncludeDeleted, customer.IncludeDeleted),
```

openmeter/customer/repository/customer.go
```diff
@@ -58,13 +69,12 @@ func (r repository) ListCustomers(ctx context.Context, params customer.ListCusto
 	return response, nil
 }
 
+// CreateCustomer creates a new customer
 func (r repository) CreateCustomer(ctx context.Context, params customer.CreateCustomerInput) (*customer.Customer, error) {
-	db := r.client()
-
-	query := db.Customer.Create().
+	// Create the customer in the database
+	query := r.tx.Customer.Create().
```
turip, line 75: r.client() would be better, as this way you get a panic if this is not executed inside a transaction.

test/customer/customer.go
```diff
@@ -0,0 +1,316 @@
...
+	require.ErrorAs(
+		t,
+		err,
+		&customer.SubjectKeyConflictError{Namespace: s.namespace, SubjectKeys: TestSubjectKeys},
+		"Creating a customer with same subject keys must return conflict error",
+	)
+}
+
+// TestUpdate tests the updating of a customer
+func (s *CustomerHandlerTestSuite) TestUpdate(ctx context.Context, t *testing.T) {
```
turip, line 115: This is not invoked from customer_test.go

test suits help with auto discovery as per here:
https://github.com/openmeterio/openmeter/pull/1541/files#diff-7a2c8eb899f11a37561acff33295dc9abe674dd76d818a3ff27348190c88a2c0

api/spec/src/customer.tsp
hekike: Done

openmeter/customer/customer.go
hekike: API response would be different. null vs empty object

openmeter/customer/repository/customer.go
hekike: It's executed in a TX

api/spec/src/types.tsp
hekike: Country is only required for paid plans / subscriptions. Making it required would be challenging to handle freemium customers where in many cases vendor doesn't have the necessary information.

I think alternatively we should add an interface to BillingProfile that validates if customer is eligible to start a paid subscription. This can be various between providers. For example with manual invoicing you may only need an address and business info while with Stripe you need default payment method too. Some invoicing providers supporting manual action required state may allow temporary missing information etc.

I think subscription module should use BillingProfile to check if condition meet for paid plan. cc @GAlexIHU

openmeter/customer/repository/customer.go
turip: But that's not explicit, and we will get a panic if something changes. That's why the client is there.

turip: Or at least add a nil check for tx if you want to mandate tx.

GAlexIHU: We'll have to rewrite how the transaction is handled to work with any other repositories, so i wouldn't worry too much about it for now

turip (approved):

PR #1546, opened 2024-09-20
refactor: id prefixes

turip:
## Overview

This change makes sure that going forward all IDs generated will have ID prefixes, causing easier identification of the entity in question.

The old data is not modified.

### DIFF
--- a/openmeter/ent/schema/idprefix.go
+++ b/openmeter/ent/schema/idprefix.go
@@ -1,0 +1,34 @@
+package schema
+
+// Table of ID prefixes for each entity type.
+// Please make sure you are picking a unique prefix for each entity type.
+// Recommendation:
+// - at least 3 characters of the entity type name.
+// - max length is 8 characters (field size limit)
+
+const (
+	// Billing
+	IDPrefixBillingProfile     = "bi_p_"
+	IDPrefixBillingInvoice     = "bi_i_"
+	IDPrefixBillingInvoiceItem = "bi_ii_"
+	IDPrefixWorkflowConfig     = "bi_wc_"
+
+	// Customer
+	IDPrefixCustomer = "cus_"
+
+	// Entitlements
+	IDPrefixEntitlement           = "en_"
+	IDPrefixEntitlementUsageReset = "en_ur_"
+
+	// Credits
+	IDPrefixGrant = "cr_g_"
+
+	// Notifications
+	IDPrefixNotificationChannel             = "no_c_"
+	IDPrefixNotificationRule                = "no_r_"
+	IDPrefixNotificationEvent               = "no_e_"
+	IDPrefixNotificationEventDeliveryStatus = "no_ed_"
+
+	// Product catalog
+	IDPrefixFeature = "pr_f_"
+)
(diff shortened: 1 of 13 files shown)


pkg/framework/entutils/mixins.go
```diff
@@ -169,3 +196,11 @@ func (c CustomerAddressMixin) Fields() []ent.Field {
 		field.String(fmt.Sprintf("%s_address_phone_number", c.FieldPrefix)).Optional().Nillable(),
 	}
 }
+
+func Must(m ent.Mixin, err error) ent.Mixin {
```
GAlexIHU, line 200: if we're already using it then i suggest maybe `lo.Must`

openmeter/ent/schema/idprefix.go
```diff
@@ -0,0 +1,34 @@
+package schema
+
+// Table of ID prefixes for each entity type.
+// Please make sure you are picking a unique prefix for each entity type.
+// Recommendation:
+// - at least 3 characters of the entity type name.
+// - max length is 8 characters (field size limit)
```
GAlexIHU, line 7: I suppose this is the hard part to get right. I'm not 100% scoping is the way to go, I'd rather define what to do in case of name collisions and use as short names as possible... wdyt?

turip: Yes, scoping is 100% TBD. Initially I was doing stuff like 
- BillingInvoiveItem => bii_
- Customer => c_

But then I didn't like the one-character names, so that's why I ended up with this scoping thing. I am not against dropping that part.

I would just mandate that there are no collisions, as seven characters should be enough.

pkg/framework/entutils/mixins.go
```diff
@@ -56,26 +69,40 @@ func (ResourceMixin) Indexes() []ent.Index {
...
+}
+
+type idMixin struct {
 	mixin.Schema
+
+	IDPrefix string
+}
+
+func (i idMixin) ULIDWithPrefix() string {
+	return fmt.Sprintf("%s_%s", i.IDPrefix, ulid.Make().String())
```
turip, line 87: TODO: With the current IDs the _ is a duplicate.

tools/migrate/migrations/20240920084000_ulid-prefixes.up.sql
```diff
@@ -0,0 +1,34 @@
+-- modify "balance_snapshots" table
+ALTER TABLE "balance_snapshots" ALTER COLUMN "owner_id" TYPE character varying(34);
+-- modify "billing_invoice_items" table
+ALTER TABLE "billing_invoice_items" ALTER COLUMN "id" TYPE character varying(34), ALTER COLUMN "invoice_id" TYPE character varying(34);
+-- modify "billing_invoices" table
+ALTER TABLE "billing_invoices" ALTER COLUMN "id" TYPE character varying(34), ALTER COLUMN "billing_profile_id" TYPE character varying(34), ALTER COLUMN "workflow_config_id" TYPE character varying(34);
```
turip, line 6: TODO: we need to change the schema override too (or remove in case of edge fields as it seems like ent autogenerates the types)

GAlexIHU:
@turip I believe we agreed we'd postpone the id prefixing for the time being, and once we do it would only affect new IDs.

Would you be fine with closing this PR, or is this still relevant?

PR #1549, opened 2024-09-20
feat: add billing related customer fields

turip:
## Overview

This change adds the required fields and the related schema. The strategy is that if an override should be set, the billing_profile_id must be set.


turip:
This is back to draft, as the billing overrides must be handled by the billing service or we will end up in customer being a kitchen-sink service.

turip:
Closing in favor of:
https://github.com/openmeterio/openmeter/pull/1566
https://github.com/openmeterio/openmeter/pull/1572

PR #1561, opened 2024-09-23
Return all validation errors from configuration

sagikazarmark:
## Overview

Instead of returning on error at a time, config now returns all validation errors.

### DIFF
--- a/cmd/balance-worker/main.go
+++ b/cmd/balance-worker/main.go
@@ -61,6 +61,7 @@ func main() {
 
 	flags.String("config", "", "Configuration file")
 	flags.Bool("version", false, "Show version information")
+	flags.Bool("validate", false, "Validate configuration and exit")
 
 	_ = flags.Parse(os.Args[1:])
 
@@ -90,6 +91,10 @@ func main() {
 		println("configuration error:")
 		println(err.Error())
 		os.Exit(1)
+	}
+
+	if v, _ := flags.GetBool("validate"); v {
+		os.Exit(0)
 	}
 
 	extraResources, _ := resource.New(
--- a/cmd/balance-worker/main.go
+++ b/cmd/balance-worker/main.go
@@ -87,7 +87,9 @@ func main() {
 
 	err = conf.Validate()
 	if err != nil {
-		panic(err)
+		println("configuration error:")
+		println(err.Error())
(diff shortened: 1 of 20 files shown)


cmd/balance-worker/main.go
```diff
@@ -87,7 +87,9 @@ func main() {
 
 	err = conf.Validate()
 	if err != nil {
-		panic(err)
+		println("configuration error:")
+		println(err.Error())
+		os.Exit(1)
```
sagikazarmark, line 92: I wonder how this should be reported in logs. If there are more than one errors, the log line is going to be loooong.

chrisgacsal (dismissed):
Nice! :shipit:

cmd/balance-worker/main.go
sagikazarmark: @chrisgacsal any thoughts on this?

chrisgacsal: How about unwrapping the error and print each error separately?

sagikazarmark: It should already print each error on a separate line.

I think I'm gonna leave it like this for now, we can always fix it later.

config/balanceworker.go
```diff
@@ -1,19 +1,25 @@
...
 type BalanceWorkerConfiguration struct {
 	ConsumerConfiguration `mapstructure:",squash"`
 }
 
 func (c BalanceWorkerConfiguration) Validate() error {
+	var errs []error
+
 	if err := c.ConsumerConfiguration.Validate(); err != nil {
-		return err
+		errs = append(errs, errorsx.WithPrefix(err, "consumer"))
```
GAlexIHU, line 19: Nit: I think it would be better if the prefixes used the full name of the configuration object, it would make things 100% unambiguous (e.g. consumer => ConsumerConfiguration)
Wdyt?

GAlexIHU (approved):

config/balanceworker.go
sagikazarmark: Hm...maybe

Right now it follows the name as it appears in the config file which probably makes it easier to debug a problem.

Ultimately, this should be a "path" to the configuration (whether it's file or not), but people rarely look at the code to figure out configuration.

PR #1562, opened 2024-09-23
Add accept header based pagination to listEntitlements

sagikazarmark:
## Overview

Add accept header based pagination to listEntitlements


sagikazarmark:
@tothandras @GAlexIHU AFAIK we agreed this isn't critical for the moment. New APIs are paginated by default and we can live with this one API until v2. Do I remember correctly?

GAlexIHU:
> @tothandras @GAlexIHU AFAIK we agreed this isn't critical for the moment. New APIs are paginated by default and we can live with this one API until v2. Do I remember correctly?

Yes, but for future (before V2) we can reference this if it came up again in other places

GAlexIHU:
Closing in it's current form, added ticket to backlog

PR #1566, opened 2024-09-24
feat: billing profile

turip:
## Overview

This patch adds the logic required for billing profile creation. What's not included in this change:
- Provider-specific configs (will be added when needed)
- Customer overrides (will be a separate PR)
- HTTP API will be a separate PR


openmeter/billing/adapter.go
```diff
@@ -0,0 +1,162 @@
...
+	ProfileAdapter
+
+	WithTx(context.Context) (TxAdapter, error)
+}
+
+type ProfileAdapter interface {
+	CreateProfile(ctx context.Context, input CreateProfileInput) (*Profile, error)
+	GetProfileByKeyOrID(ctx context.Context, input GetProfileByKeyOrIDInput) (*Profile, error)
+	GetProfileByID(ctx context.Context, input GetProfileByIDAdapterInput) (*Profile, error)
+	GetProfileByKey(ctx context.Context, input GetProfileByKeyAdapterInput) (*Profile, error)
```
hekike, line 26: do profiles need a key? Latest discussion was that key mostly makes sense for frontend facing resources.

openmeter/billing/adapter/profile.go
```diff
@@ -0,0 +1,319 @@
+package billingadapter
```
hekike, line 1: I thought we handle DB operation in `repository`

openmeter/billing/adapter/profile.go
```diff
@@ -0,0 +1,319 @@
...
+		return nil, err
+	}
+
+	dbProfiles, err := a.client().BillingProfile.Query().
+		Where(billingprofile.Namespace(input.Namespace)).
+		Where(billingprofile.Or(
+			billingprofile.And(billingprofile.Key(input.IDOrKey), billingprofile.DeletedAtIsNil()),
+			billingprofile.ID(input.IDOrKey),
+		)).
+		WithWorkflowConfig().All(ctx)
```
hekike, line 79: You can do `query.First(ctx)`

openmeter/billing/adapter/profile.go
```diff
@@ -0,0 +1,319 @@
...
+		if dbProfile.ID == input.IDOrKey {
+			return mapProfileFromDB(dbProfile), nil
+		}
+	}
+
+	return nil, nil
+}
+
+func (a adapter) GetProfileByID(ctx context.Context, input billing.GetProfileByIDAdapterInput) (*billing.Profile, error) {
+	if err := input.Validate(); err != nil {
```
hekike, line 104: why not call `GetProfileByKeyOrID` from it?

openmeter/billing/adapter/profile.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+
+	profile, err := a.GetProfileByID(ctx, billing.GetProfileByIDAdapterInput(input))
+	if err != nil {
+		return err
+	}
+
+	c := a.client()
+
+	_, err = c.BillingWorkflowConfig.UpdateOneID(profile.WorkflowConfig.ID).
```
hekike, line 184: namespace where?

hekike (dismissed):

openmeter/billing/adapter/profile.go
turip: https://github.com/openmeterio/openmeter/pull/1544#discussion_r1768339413

openmeter/billing/adapter/profile.go
turip: No, in case of a key ID collusion this prioritizes keys.

openmeter/billing/adapter/profile.go
turip: Because:
- KeyOrID is only used if the service don't know which one to fetch (e.g. when it's coming from the frontend)
- We call ID explicitly in the code if we want to lookup by ID
- We call the getByKey if we want to look up by Key

I think it's cleaner to limit the KeyOrID's ambiguity as much as possible. The only reason this exists is because we will have KeyOrID queries and I wanted to have a single query running in such cases.

Also I am not huge fan of even the slightest possibility to resolve a ULID formatted key by accident ;-)

openmeter/billing/adapter.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	if i.Key == "" {
+		return errors.New("key is required")
+	}
+
+	return nil
+}
+
+func WithTxNoValue(ctx context.Context, repo Adapter, fn func(ctx context.Context, repo TxAdapter) error) error {
```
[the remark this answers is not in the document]
turip: Yes, I didn't wanted to start mandating the TxAdapter interface as a generic type. I think that's a bit too convoluted.

openmeter/billing/adapter.go
turip: I am fine dropping the key, but my understanding was that that's an external ID from the foreign system. If we can drop the key that will help a lot with the logic. WDYT?

hekike (approved):

PR #1567, opened 2024-09-24
feat(api): add Notification API to TypeSpec

chrisgacsal:
## Overview

Add Notification API to TypeSpec.


api/spec/src/notification.tsp
```diff
@@ -8,6 +8,468 @@ using TypeSpec.OpenAPI;
...
+     *
+     * Usage: `?includeDisabled=false`
+     */
+    @query
+    @example(false)
+    includeDisabled?: boolean,
+
+    ...OpenMeter.QueryPagination,
+    ...OpenMeter.QueryOrdering<NotificationChannelOrderBy>,
+  ): OpenMeter.PaginatedResponse<NotificationChannel> | OpenMeter.CommonErrors;
```
sagikazarmark, line 187: Since it's in the same namespace, you don't need `OpenMeter.` I believe.

chrisgacsal: Notification has itw on namespace now.

chrisgacsal:
@sagikazarmark I belive the PR is ready for review. As we discussed offline, we need to move Subject API to OSS from Cloud as Notification API relies on the `Subject` model [here](https://github.com/openmeterio/openmeter/pull/1567/files#diff-4cf643c7d45cd150e9f782f120b957b40a15533929432809a8cb58eddf1f23a4R111).

tothandras (approved):

sagikazarmark (approved):

PR #1572, opened 2024-09-24
feat: billing customer overrides

turip:
## Overview

This implementation allows us to specify the overrides per customer. Right now we are only overriding the workflow configuration part, later if providers are there we can add that to this code.

The API is able to yield a merged (Billing)Profile that we will be using for the invoice creation, where we are snapshotting the current settings.

Additional overrides will be required for the default profile changes when provider overrides are in place OM-927


openmeter/billing/adapter/customeroverride.go
```diff
@@ -0,0 +1,239 @@
...
+			Entity: billing.EntityCustomer,
+			Err:    billing.ErrCustomerNotFound,
+		}
+	}
+
+	return mapCustomerOverrideFromDB(dbCustomerOverride), nil
+}
+
+func (r adapter) DeleteCustomerOverride(ctx context.Context, input billing.DeleteCustomerOverrideInput) error {
+	err := r.client().BillingCustomerOverride.Update().
```
hekike, line 153: should we check if it updates any?

openmeter/ent/schema/billing.go
```diff
@@ -112,6 +114,98 @@ func (BillingWorkflowConfig) Edges() []ent.Edge {
...
+}
+
+func (BillingCustomerOverride) Edges() []ent.Edge {
+	return []ent.Edge{
+		edge.From("customer", Customer.Type).
+			Ref("billing_customer_override").
+			Field("customer_id").
+			Unique().
+			Required().
+			Immutable(),
```
hekike, line 201: do we need cascade?

hekike (approved):

openmeter/billing/adapter/customeroverride.go
turip: No, as the service already checked that and we are in the same transaction, but changed the code.

openmeter/ent/schema/billing.go
turip: That's defined at the To edge.

hekike (approved):

PR #1573, opened 2024-09-24
Fix/update or clear

turip:
## Overview

If the SetNillable* is called with a nil, it will ignore the value instead of setting it to null. This helper can be used to call the Clear/Set based on the value itself.


chrisgacsal (approved):
Nice job! :shipit:

GAlexIHU (approved):
Gosh, I suppose we didn't notice before as most entities are immutable once created... 
Thanks 👍

turip:
> Gosh, I suppose we didn't notice before as most entities are immutable once created... Thanks 👍

Thanks, but the credit goes to @hekike, I was just the hand executing the change (he encountered this issue with the Customer implementation).

PR #1578, opened 2024-09-26
fix(api): query explode

chrisgacsal:
## Overview

Use `explode=true` for query parameters which accept multiple values in order to keep backward compatibility.


GAlexIHU (approved):
👍

sagikazarmark (approved):

PR #1585, opened 2024-09-27
Feat/integration poc

turip:
## Overview

Major differences to the base PR:
- Capabilities are just informational, mostly for user interface
- The actual validation/usage is performed using golang interfaces thus we can enforce some type correctness
- Adds placeholder for db storage
- Adds placeholder for HTTP endpoints
- Integrations are fully self contained, the integration is responsible for data persisting etc


pkg/framework/entutils/mixins.go
```diff
@@ -186,3 +186,14 @@ func (c CustomerAddressMixin) Fields() []ent.Field {
...
+// TODO: this should be inheriting the namespace mixin too probably
+type AppID struct {
+	ent.Schema
+}
+
+func (AppID) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("app_id").Immutable(),
+	}
+}
```
hekike, line 199: Edge is missing

PR #1607, opened 2024-10-02
fix(streaming): alias meter query where

hekike:
Alias meter table and prefix columns are used in the where condition to avoid shadowing by aggregated columns.
This fixes querying hourly and daily windows as well as filters the underlying per-minute windows, not the rounded tumbling window.

### DIFF
--- a/openmeter/streaming/clickhouse_connector/query.go
+++ b/openmeter/streaming/clickhouse_connector/query.go
@@ -291,7 +291,8 @@ type queryMeterView struct {
 }
 
 func (d queryMeterView) toSQL() (string, []interface{}, error) {
-	viewName := GetMeterViewName(d.Database, d.Namespace, d.MeterSlug)
+	viewAlias := "meter"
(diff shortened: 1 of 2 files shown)


openmeter/streaming/clickhouse_connector/query.go
```diff
@@ -291,7 +291,8 @@ type queryMeterView struct {
 }
 
 func (d queryMeterView) toSQL() (string, []interface{}, error) {
-	viewName := GetMeterViewName(d.Database, d.Namespace, d.MeterSlug)
+	viewAlias := "meter"
+	viewName := GetMeterViewName(d.Database, d.Namespace, fmt.Sprintf("%s %s", d.MeterSlug, viewAlias))
```
tothandras, line 295:
```suggestion
	viewName := fmt.Sprintf("%s %s", GetMeterViewName(d.Database, d.Namespace, d.MeterSlug), viewAlias)
```

tothandras (approved):

PR #1613, opened 2024-10-03
Subscription api proposals

GAlexIHU:
## Overview

Based on the new additions to the internal product spec

## Notes for reviewer
- rough sketch, starting point


api/spec/src/subscription.tsp
```diff
@@ -0,0 +1,266 @@
...
+@tag("Subscriptions")
+interface Subscriptions {
+  @get
+  @operationId("getSubscription")
+  get(@path subscriptionId: ULID): Subscription | OpenMeter.NotFoundError | OpenMeter.CommonErrors;
+
+  create(
+    @body body: {
+      customerId: ULID;
+      templatingPlan: {
```
sagikazarmark, line 25: Is `templatingPlan` an industry-standard phrase? It sounds weird to me.

api/spec/src/subscription.tsp
```diff
@@ -0,0 +1,266 @@
...
+        planKey: Key;
+        planVersion: integer;
```
sagikazarmark, line 27: `plan` stutters. How about `key` and `version`?

api/spec/src/subscription.tsp
```diff
@@ -0,0 +1,266 @@
...
+      };
+      currency: CurrencyCode;
+      activeFrom: DateTime;
```
sagikazarmark, line 30: `activeSince` or `startDate`?

api/spec/src/subscription.tsp
```diff
@@ -0,0 +1,266 @@
...
+
+@route("/api/v1/subscriptions")
+@tag("Subscriptions")
+interface Subscriptions {
+  @get
+  @operationId("getSubscription")
+  get(@path subscriptionId: ULID): Subscription | OpenMeter.NotFoundError | OpenMeter.CommonErrors;
+
+  create(
+    @body body: {
```
sagikazarmark, line 23: I'm guessing this should become a model at some point?

api/spec/src/subscription.tsp
```diff
@@ -0,0 +1,266 @@
...
+  /**
+   * Cancels the subscription.
+   */
+  @post
+  @operationId("cancelSubscription")
+  @route("/{subscriptionId}/cancel")
+  cancel(
+    @path subscriptionId: ULID,
+    @body body: {
+      effective: "IMMEDIATE" | DateTime;
```
sagikazarmark, line 57: How about `effectiveDate`, meaning immediate if omitted?

api/spec/src/subscription.tsp
```diff
@@ -0,0 +1,266 @@
...
+    },
+  ): Subscription | OpenMeter.ConflictError | OpenMeter.NotFoundError | OpenMeter.CommonErrors;
+
+  /**
+   * Cancels the subscription.
```
sagikazarmark, line 62: copy pasta

api/spec/src/subscription.tsp
```diff
@@ -0,0 +1,266 @@
...
+  deletePhaseItem(@path subscriptionId: ULID, @path phaseKey: Key, @path itemKey: Key): {
+    @statusCode _: 204;
+  } | OpenMeter.NotFoundError | OpenMeter.CommonErrors;
+}
+
+/**
+ * Subscription is an exact subscription instance.
+ */
+@friendlyName("Subscription")
+model Subscription extends global.Resource {
```
sagikazarmark, line 142: Afaik this should be used with spreading. Am I right @tothandras ?

api/spec/src/subscription.tsp
```diff
@@ -0,0 +1,266 @@
...
+   * The currency code of the subscription.
+   * Will be revised once we add multi currency support.
+   */
+  @summary("Currency")
+  currency: CurrencyCode = "USD";
+
+  /**
+   * If the subscription is canceled or otherwise have to end activeTo denotes the end date.
+   */
+  activeTo?: DateTime;
```
sagikazarmark, line 173: `endDate` or `activeUntil`?

api/spec/src/subscription.tsp
GAlexIHU: they usually call it plan

sagikazarmark (commented):
Couple random thoughts

api/spec/src/subscription.tsp
sagikazarmark: In case I wasn't clear, the `templating` part is what's weird to me. Something like `basePlan` sounds more natural, but I'm not familiar with the lingo.

api/spec/src/subscription.tsp
GAlexIHU: For subscription in particular  `activeFrom` & `activeTo` doesnt sound the best, but i'd like them to be the standard across resources, see `CadencedModel`. (I'm open to other suggestions for the standard but I suppose that's the most general term we can use)

api/spec/src/subscription.tsp
tothandras: Yes

solkimicreb (commented):
I am trying to fully understand the patch language.

- I think you mentioned there can be no rate card with the same feature (a key restriction). I assume the same is true about featureless rate cards with the same price? Is this an issue? Probably not but note to myself: we need to enforce this on the UI.

- Are the above (key, feature, price) restrictions per subscription or per phase? I assume per phrase and the "edit id" for the card also includes the phase id.

- I am not sure I understand the flat patch language for the nested structure of rate cards and phases. How do I edit a phase? For example if I want to extend a phase I send the whole phase with all rate cards and only the duration field changed? Networking is still a none issue imo but storing patches like this may defeat the purpose.

- What if I delete a phase? Do I send a phase delete patch or do I also send all enclosed rate cards delete patches too to make it work?

- Same goes about adding a phase: is it a single big patch or do I need to flatten it out? If so, how?

- What about other interdependencies in the subscription structure. For example if I have a discount applying to a rate card and I delete the rate card do I simply send the rate card delete and let the api make it work or do I also have to send the discount patch without the rate card id?

api/spec/src/plan.tsp
```diff
@@ -75,75 +76,26 @@ model Plan extends UniqueResource {
...
+  ...SubscriptionPhaseCreateInput;
+
   /**
-   * Interval after the subscription starts to transition to the phase.
-   * When null, the phase starts immediately after the subscription starts.
+   * Unique key among the plan phases. Used to reference the phase in the subscription & plan.
    */
-  @summary("Start after")
-  startAfter: Interval | null;
+  key: Key;
```
hekike, line 84: same, consider an `alias PlanKey = Key`

api/spec/src/ratecards.tsp
```diff
@@ -18,10 +18,37 @@ enum RateCardType {
...
+  ...RateCard;
+}
+
+/**
+ * A rate card defines the pricing and entitlement of a feature or service.
+ */
+@friendlyName("RateCard")
+model RateCard {
+  /**
+   * Copied from product doc:
```
hekike, line 35: > Copied from product doc:

this comment doesn't make sense in context of OSS

api/spec/src/ratecards.tsp
```diff
@@ -18,10 +18,37 @@ enum RateCardType {
...
+   *
+   * 1. If a Feature is associated with the SubscriptionItem, it is identified by the Feature
+   *  1.1 It can be an ID reference, for an exact version of the Feature (Features can change across versions)
+   *  1.2 It can be a Key reference, which always refers to the latest (active or inactive) version of a Feature
+   *
+   * 2. If a Feature is not associated with the SubscriptionItem, it is referenced by the Price
+   *
+   * We say “referenced by the Price” regardless of how a price itself is referenced, it colloquially makes sense to say “paying the same price for the same thing”
+   *
+   * In practice this will need an ID being generated
```
hekike, line 48: > In practice this will need an ID being generated

sounds odd in the comment

api/spec/src/ratecards.tsp
```diff
@@ -74,69 +83,12 @@ model RateCardFlatFee extends RateCardBase {
...
- */
-@friendlyName("RateCardOverrideUsageBased")
-model RateCardOverrideUsageBased {
-  ...PickProperties<RateCardUsageBased, "type" | "price" | "discount">;
+  price:
+    | FlatPriceWithPaymentTerm
+    | UnitPriceWithCommitments
+    | TieredPriceWithCommitments
+    | FlatPriceWithPaymentTerm
+    | null;
```
hekike, line 91: why not?

```
  price?:
    | FlatPriceWithPaymentTerm
    | UnitPriceWithCommitments
    | TieredPriceWithCommitments
    | FlatPriceWithPaymentTerm
```

hekike: With `null`, you explicitly have to send `null` on API versus skipping the property.

api/spec/src/subscription.tsp
```diff
@@ -0,0 +1,390 @@
...
+
+@route("/api/v1/subscriptions")
+@tag("Subscriptions")
+interface Subscriptions {
+  @get
+  @operationId("getSubscription")
+  @sharedRoute
+  get(
+    @path subscriptionId: ULID,
+    @query expand?: false,
```
hekike, line 30: Should we use a more detailed expansion?

```
expand: ['customer']
```

See https://docs.stripe.com/expand

api/spec/src/subscription.tsp
```diff
@@ -0,0 +1,390 @@
...
+
+      /**
+       * JSON Patch (RFC6902).
+       *
+       * It is a partial implementation of the JSON Patch standard:
+       * 1. It only respecting the `add` and `remove` operations. All other operations defined in RFC6902 are ignored during execution.
+       * 2. The provided value for ADD operation doesn't match the target type. It is the creation input type for the SubscriptionItem.
+       */
+      @maxItems(100)
+      overrides: SubscriptionItemOverride[];
```
hekike, line 69: Naming: is it an override? It can also extend a plan.
Not sure it's much better but `extensions`, `modifications` etc.

api/spec/src/subscription.tsp
```diff
@@ -0,0 +1,390 @@
...
+    },
+  ): Subscription | OpenMeter.ConflictError | OpenMeter.CommonErrors;
+
+  /**
+   * Applies the JSON Patch (RFC6902) atomically (RFC5789).
```
hekike, line 74: is `atomically` the word? https://dictionary.cambridge.org/us/dictionary/english/atomically

api/spec/src/subscription.tsp
```diff
@@ -0,0 +1,390 @@
...
+   *
+   * It is a partial implementation of the JSON Patch standard:
+   * 1. It only respecting the `add` and `remove` operations. All other operations defined in RFC6902 are ignored during execution.
+   * 2. The provided value for ADD operation doesn't match the target type. It is the creation input type for the SubscriptionItem and SubscriptionPhase.
+   */
+  @patch
+  @operationId("editSubscription")
+  @route("/{subscriptionId}")
+  edit(
+    @header accept: "json-patch+json",
```
hekike, line 84: Is this header required or nice to have? We learned from the ingest endpoint that custom content types are challenging for users.

api/spec/src/subscription.tsp
```diff
@@ -0,0 +1,390 @@
...
+      effectiveDate?: DateTime;
+    },
+  ): Subscription | OpenMeter.ConflictError | OpenMeter.NotFoundError | OpenMeter.CommonErrors;
+
+  /**
+   * Cancels the scheduled cancelation.
+   */
+  @post
+  @operationId("cancelScheduledCancelation")
+  @route("/{subscriptionId}/cancel-scheduled-cancelation")
```
hekike, line 127: mounthful

api/spec/src/subscription.tsp
```diff
@@ -0,0 +1,390 @@
...
+      extendBy: duration;
+    },
+  ): SubscriptionPhase[] | OpenMeter.ConflictError | OpenMeter.CommonErrors;
+
+  // Direct management of a SubscriptionPhase's SubscriptionItems would be possible via these APIs.
+  // Its unclear if we'll need them.
+
+  @get
+  @operationId("getSubscriptionPhaseItemsAt")
+  @route("/{subscriptionId}/phases/{phaseKey}/items")
```
hekike, line 186: Do we have use cases where we need to get multiple phases? wondering if list with phase key filter is more flexible

api/spec/src/subscription.tsp
hekike: +1

api/spec/src/subscription.tsp
```diff
@@ -0,0 +1,390 @@
...
+/**
+ * The actual contents of the Subscription, what the user gets, what they pay, etc...
+ */
+@friendlyName("SubscriptionItem")
+model SubscriptionItem {
+  // Key is overriden to add extra doc. Uses the same type.
+  ...OmitProperties<VirtualResource, "key">;
+
+  /**
+   * Copied from product doc:
```
hekike, line 318: same

api/spec/src/subscription.tsp
```diff
@@ -0,0 +1,390 @@
...
+   *
+   * 1. If a Feature is associated with the SubscriptionItem, it is identified by the Feature
+   *  1.1 It can be an ID reference, for an exact version of the Feature (Features can change across versions)
+   *  1.2 It can be a Key reference, which always refers to the latest (active or inactive) version of a Feature
+   *
+   * 2. If a Feature is not associated with the SubscriptionItem, it is referenced by the Price
+   *
+   * We say “referenced by the Price” regardless of how a price itself is referenced, it colloquially makes sense to say “paying the same price for the same thing”
+   *
+   * In practice this will need an ID being generated
```
hekike, line 331: same

api/spec/src/subscription.tsp
```diff
@@ -0,0 +1,390 @@
...
+   * When null, the feature or service is free.
+   */
+  @summary("Price")
+  @example(#{ type: PriceType.flat, amount: "100", paymentTerm: "in_arrears" })
+  price:
+    | FlatPriceWithPaymentTerm
+    | UnitPriceWithCommitments
+    | TieredPriceWithCommitments
+    | FlatPriceWithPaymentTerm
+    | null;
```
hekike, line 354: same

api/spec/src/subscription.tsp
hekike: +1.

api/spec/src/subscription.tsp
GAlexIHU: "The server MUST apply the entire set of changes atomically and never
   provide (e.g., in response to a GET during this operation) a
   partially modified representation." - RFC5789
   
   yes.

api/spec/src/subscription.tsp
GAlexIHU: it is part of the spec

api/spec/src/subscription.tsp
GAlexIHU: we could do `@route("/{subscriptionId}/unschedule-cancelation")`

api/spec/src/subscription.tsp
GAlexIHU: tbh I'm not sure if these endpoints are necessary at all, I don't really see why someone wouldn't just use `get` with `expand=true`

api/spec/src/subscription.tsp
GAlexIHU: I see little added value, I think the current form is the 90% usecase

api/spec/src/ratecards.tsp
GAlexIHU: I believe it was the preference of @tothandras to use explicit nulls

api/spec/src/subscription.tsp
GAlexIHU: renamed it to `customizations`

GAlexIHU:
> I am trying to fully understand the patch language.
> 
> * I think you mentioned there can be no rate card with the same feature (a key restriction). I assume the same is true about featureless rate cards with the same price? Is this an issue? Probably not but note to myself: we need to enforce this on the UI.
> * Are the above (key, feature, price) restrictions per subscription or per phase? I assume per phrase and the "edit id" for the card also includes the phase id.

There are 2 things here:
1. You can only have access to, and pay for, a given feature once.
2. `RateCards` & `SubscriptionItems` should have a `Key` based on this proposal which is locally unique. For that I suggest we use the `featureKey` if present, if not then a key has to be provided/generated.

> * I am not sure I understand the flat patch language for the nested structure of rate cards and phases. How do I edit a phase? For example if I want to extend a phase I send the whole phase with all rate cards and only the duration field changed? Networking is still a none issue imo but storing patches like this may defeat the purpose.

The patch language is only used for the RateCards (there are 3 extra EPs only for the phase management). We could extend it for the phases as well, my issue is that if you want to add phases, it's really not intuitive to reference them as json document members. It would be much more intuitive to reference them array-like, so you could insert phases, remove phases etc... but then (if we use the same path) the SubscriptionItem references wouldn't be stable.

> * What if I delete a phase? Do I send a phase delete patch or do I also send all enclosed rate cards delete patches too to make it work?

see above, there's a dedicated ep defined

> * Same goes about adding a phase: is it a single big patch or do I need to flatten it out? If so, how?

same

> * What about other interdependencies in the subscription structure. For example if I have a discount applying to a rate card and I delete the rate card do I simply send the rate card delete and let the api make it work or do I also have to send the discount patch without the rate card id?

Discounts are not yet defined in the API, but a fair question. I suppose we could think of a Discount's `appliesTo` field as a filter and not as a reference, it shouldn't be an issue to delete the SubscriptionItem

api/spec/src/productcatalog/plan.tsp
```diff
@@ -71,81 +76,27 @@ model Plan {
  */
 @friendlyName("PlanPhase")
 model PlanPhase {
-  ...OmitProperties<UniqueResource, "id">;
+  ...SubscriptionPhaseCreateInput;
```
chrisgacsal, line 79: Honestly I do not get this. Why the subscription models cannot reuse teh models defiend in plan? The subscription relies on plan and not the other way around. So I would rather put all the plan related fields and adjust subscription models accordingly. I am ok with changing field names, types, but not comfortable with pulling subscription models into plan. It feels super wierd and a good recept for disaster (e.g. circular dependency, etc.).

chrisgacsal: In other words, plan models should be independent from subscription although synergies must be there for sure.

api/spec/src/types.tsp
```diff
@@ -103,6 +114,31 @@ model Resource {
...
+model VirtualResource {
+  ...OmitProperties<Resource, "id">;
+  ...Keyed;
+}
+
+/**
+ * Represents resources that can be cadenced, have scheduled activity changes.
+ */
+@friendlyName("CadencedResource")
+model CadencedResource {
```
chrisgacsal, line 130: Would love to have a settlement on the naming used for describing time period, so I can also update the plan api.

api/spec/src/productcatalog/ratecards.tsp
```diff
@@ -22,7 +22,7 @@ model RateCardBase<T extends RateCardType> {
   /**
    * The type of the RateCard.
    */
-  @visibility("read", "create")
+  @visibility("read", "create", "update")
   @summary("Channel Type")
```
chrisgacsal, line 26: I'll fix this one. Copypasta error.

api/spec/src/productcatalog/subscription.tsp
```diff
@@ -0,0 +1,231 @@
...
+  ...OmitProperties<SubscriptionPhaseCreateInput, "startAfter">;
+
+  /**
+   * The time from which the phase is active.
+   */
+  activeFrom: DateTime;
+}
+
+@friendlyName("SubscriptionPhaseCreateInput")
+model SubscriptionPhaseCreateInput {
```
chrisgacsal, line 94:
```suggestion
model SubscriptionPhaseCreate {
```
So we are aligned with naming used by rest helpers like `TypeSpec.Rest.Resource.ResourceCreateModel` which normally should be used. (This one is an exception).

api/spec/src/types.tsp
GAlexIHU: sure, I believe this is something we talked about around early September, but if you have any preferences lets change accordingly

chrisgacsal (dismissed):

chrisgacsal (approved):

PR #1615, opened 2024-10-04
fix: fix slice init length

cuishuang:
## Overview

The intention here should be to initialize a slice with a capacity of  `len(items)`  rather than initializing the length of this slice. 

The online demo: https://go.dev/play/p/q1BcVCmvidW


chrisgacsal:
Hey @cuishuang! Nice catch. Thanks for your contrib.

sagikazarmark (approved):
Nice catch, thanks!

PR #1617, opened 2024-10-04
Entitlement Scheduling Internals

GAlexIHU:
## Overview

Adds the internals to enable entitlement scheduling. The new entitlement properties are `activeFrom` and `activeTo`.
(these are not externally visible)

Introduces two new methods
- `ScheduleEntitlement`
- `SupersedeEntitlement`
which are wrapped for the public api by `Create` and `Override`.

⚠️ Changes the `clock` package to allow for time freezes ⚠️

## Notes for reviewer

- Not sure if connector/service methods wrapping each other is the way to go, I didn't want to introduce an extra layer for it
- I need to sync with @turip to see if these changes would affect the Edge Worker & Notifications engine in ways I couldn't account for

## Missing

- [x] Value mapping for inactive entitlements
- [x] Effects on Edge Worker


pkg/clock/clock.go
```diff
@@ -1,19 +1,40 @@
 package clock
 
-import "time"
+import (
```
sagikazarmark, line 3: Have you considered using https://github.com/jonboulle/clockwork instead?

openmeter/ent/schema/entitlement.go
```diff
@@ -31,6 +31,8 @@ func (Entitlement) Fields() []ent.Field {
 		field.String("feature_id").Immutable().SchemaType(map[string]string{
 			dialect.Postgres: "char(26)",
 		}),
+		field.Time("active_from").Optional().Nillable().Immutable(),
+		field.Time("active_to").Optional().Nillable(),
```
sagikazarmark, line 35: since and until might be better than form and to

pkg/clock/clock.go
GAlexIHU: I can open open a separate PR for that and then rebase this if you'd like to make that change

openmeter/ent/schema/entitlement.go
```diff
@@ -31,6 +31,8 @@ func (Entitlement) Fields() []ent.Field {
 		field.String("feature_id").Immutable().SchemaType(map[string]string{
 			dialect.Postgres: "char(26)",
 		}),
+		field.Time("active_from").Optional().Nillable().Immutable(),
```
hekike, line 34: when is `active_from` null? For non metered entitlements?

openmeter/entitlement/scheduling.go
```diff
@@ -0,0 +1,195 @@
...
+		})
+		if err != nil {
+			return nil, err
+		}
+
+		return ent, err
+	})
+}
+
+func (c *entitlementConnector) SupersedeEntitlement(ctx context.Context, entitlementId string, input CreateEntitlementInputs) (*Entitlement, error) {
```
hekike, line 129: You mean replace? https://www.britannica.com/dictionary/supersede

openmeter/entitlement/adapter/entitlement.go
```diff
@@ -166,6 +172,39 @@ func (a *entitlementDBAdapter) DeleteEntitlement(ctx context.Context, entitlemen
...
+			c, err := repo.db.Entitlement.Update().
+				Where(db_entitlement.ID(entitlementID.ID), db_entitlement.Namespace(entitlementID.Namespace)).
+				SetActiveTo(at).
+				Save(ctx)
+			if err != nil {
+				return nil, err
+			}
+
+			if c != 1 {
+				return nil, fmt.Errorf("updated %d entitlements, which is not 1", c)
```
turip, line 198: Maybe return a not found error for c == 0

openmeter/entitlement/adapter/entitlement.go
```diff
@@ -529,6 +582,71 @@ func (a *entitlementDBAdapter) ListNamespacesWithActiveEntitlements(ctx context.
...
+		func(ctx context.Context, repo *entitlementDBAdapter) (*[]entitlement.Entitlement, error) {
+			res, err := repo.db.Entitlement.Query().
+				Where(
+					db_entitlement.Or(
+						db_entitlement.ActiveToIsNil(),
+						db_entitlement.ActiveToGT(starting),
+					),
+				).
+				Where(
+					db_entitlement.Or(db_entitlement.DeletedAtIsNil(), db_entitlement.DeletedAtGT(clock.Now())),
```
turip, line 598: Don't we need to also filter for ActiveTo < clock.Now()?

openmeter/entitlement/balanceworker/cachehandler.go
```diff
@@ -0,0 +1,32 @@
...
+func (w *Worker) handleCacheMissEvent(ctx context.Context, event *edge.EntitlementCacheMissEvent, source string) (marshaler.Event, error) {
+	currentTime := clock.Now()
+
+	if event == nil {
+		return nil, fmt.Errorf("event is nil")
+	}
+
+	ent, err := w.entitlement.Entitlement.GetEntitlement(ctx, event.EntitlementNamespace, event.EntitlementIdOrFeatureKey)
+	if _, ok := lo.ErrorsAs[*entitlement.NotFoundError](err); ok {
+		ent, err = w.entitlement.EntitlementRepo.GetActiveEntitlementOfSubjectAt(ctx, event.EntitlementNamespace, event.SubjectKey, event.EntitlementIdOrFeatureKey, event.At)
```
turip, line 24: Why don't we just use this call by default instead?

turip: Also the overal logic is that we should query at currentTime (even if the event happened earlier), so that we include any changes up to current time.

openmeter/entitlement/entitlement.go
```diff
@@ -107,6 +115,37 @@ type Entitlement struct {
 	Config []byte `json:"config,omitempty"`
 }
 
+// ActiveFromTime returns the time the entitlement is active from. Its either the ActiveFrom field or the CreatedAt field
+func (e Entitlement) ActiveFromTime() time.Time {
+	return defaultx.WithDefault(e.ActiveFrom, e.CreatedAt)
```
turip, line 120: nit: let's use lo

chrisgacsal: See: https://github.com/samber/lo?tab=readme-ov-file#fromptror

Also I am thinking about making `defaultx.WithDefault` as an alias for `lo.FromPtrOr`.

openmeter/entitlement/adapter/entitlement.go
```diff
@@ -166,6 +172,39 @@ func (a *entitlementDBAdapter) DeleteEntitlement(ctx context.Context, entitlemen
...
+			ent, err := repo.GetEntitlement(ctx, entitlementID)
+			if err != nil {
+				return nil, err
+			}
+
+			if ent.ActiveTo != nil {
+				return nil, fmt.Errorf("entitlement %s is already deactivated", entitlementID.ID)
+			}
+
+			c, err := repo.db.Entitlement.Update().
```
chrisgacsal, line 189: Why not using `Entitlement.UpdateOneID()` instead?

chrisgacsal, line 189: Why not using `Entitlement.UpdateOneID()` instead?

GAlexIHU: So i can throw an error if it's already deactivated

openmeter/ent/schema/entitlement.go
GAlexIHU: For entitlements that don't use scheduling (including all historical ones)

openmeter/entitlement/scheduling.go
GAlexIHU: I mean supersede

https://www.merriam-webster.com/thesaurus/supersede#:~:text=Some%20common%20synonyms%20of%20supersede,%2C%20obsolete%2C%20or%20otherwise%20inferior.

openmeter/entitlement/adapter/entitlement.go
GAlexIHU: line 182 already returns NotFound if not found
```go
ent, err := repo.GetEntitlement(ctx, entitlementID)
if err != nil {
  return nil, err
}
```

openmeter/entitlement/adapter/entitlement.go
GAlexIHU: I suppose you mean `activeFrom < clock.Now()`

What this query tries to do is simply get all entitlements that are relevant now or will be relevant later
(scheduled to activate later, or are currently active)

There are some data consistency presumptions with this query, being that no two can be active at the same time and `activeFrom` < `activeTo`

openmeter/entitlement/adapter/entitlement.go
GAlexIHU: Sorry, i really miscomprehended that

pkg/clock/clock.go
GAlexIHU: One aspect thats different is that we use the current as a global, changing to clockwork would also need the clock to be injected at each point

turip (approved):

turip (approved):

PR #1629, opened 2024-10-07
refactor: Notification API typespec

chrisgacsal:
## Overview

Refactor typespec definition of Notification API in order to make it aligned with the existing implementation.


chrisgacsal:
Chaned are moved to: https://github.com/openmeterio/openmeter/pull/1632

PR #1633, opened 2024-10-07
Wire refactor

sagikazarmark:
## Overview

Use Wire for dependency injection

Fixes #1618


openmeter/app/clickhouse.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"fmt"
+
+	"github.com/ClickHouse/clickhouse-go/v2"
+	"github.com/ClickHouse/clickhouse-go/v2/lib/driver"
+
+	"github.com/openmeterio/openmeter/config"
+)
+
+// TODO: add closer function?
```
chrisgacsal, line 12: Agreed, we need to gracefully close connections to server. Same way as we do for pgx/ent.

openmeter/app/kafka.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/confluentinc/confluent-kafka-go/v2/kafka"
+	"go.opentelemetry.io/otel/metric"
+
+	"github.com/openmeterio/openmeter/config"
+	pkgkafka "github.com/openmeterio/openmeter/pkg/kafka"
+	kafkametrics "github.com/openmeterio/openmeter/pkg/kafka/metrics"
+)
+
+// TODO: use ingest config directly?
+// TODO: use kafka config directly?
```
chrisgacsal, line 16: Keep it as is for now. I'll refactor the Kafka config init/usage in services in order to adopt the `Producer/ConsumerConfig` sturct available in `pkg/kafka` pkg: https://github.com/openmeterio/openmeter/blob/main/pkg/kafka/config.go#L328

openmeter/app/kafka.go
```diff
@@ -0,0 +1,44 @@
...
+// TODO: add closer function?
```
chrisgacsal, line 17: Yup, we do need closer here as well.

sagikazarmark:
This is just the first batch, needs more work, but it's a solid start.

PR #1634, opened 2024-10-08
fix: quickstart healthcheck

chrisgacsal:
## Overview

Fixes #(issue)


chrisgacsal:
Close as the fix is moved to: https://github.com/openmeterio/openmeter/pull/1632

PR #1637, opened 2024-10-08
feat(app): add with stripe app type

hekike:
- Adds apps
- Adds stripe app type
- Refactor customer pkg to use adapter


openmeter/app/entity/app.go
```diff
@@ -0,0 +1,94 @@
...
+	GetType() appentitybase.AppType
+	GetName() string
+	GetStatus() appentitybase.AppStatus
+	Get() appentitybase.MarketplaceListing
+
+	// ValidateCapabilities validates if the app can run for the given capabilities
+	ValidateCapabilities(capabilities []appentitybase.CapabilityType) error
+
+	// // ValidateCustomer validates if the app can run for the given customer
+	ValidateCustomer(ctx context.Context, customer *customerentity.Customer, capabilities []appentitybase.CapabilityType) error
```
turip, line 26: This should not be part of this interface; this is a generic app, not an app capable of customer sync.

openmeter/appstripe/adapter/stripe.go
```diff
@@ -0,0 +1,344 @@
...
+			Name:        input.Name,
+			Description: input.Description,
+			Type:        appentitybase.AppTypeStripe,
+		})
+		if err != nil {
+			return appstripeentity.App{}, fmt.Errorf("failed to create app: %w", err)
+		}
+
+		// Create the stripe app in the database
+		appStripeCreateQuery := a.db.AppStripe.Create().
```
turip, line 47: Should we ensure that only one application can exist per stripe account?

openmeter/appstripe/adapter/stripe.go
```diff
@@ -0,0 +1,344 @@
...
+		if err != nil {
+			return appstripeentity.App{}, err
+		}
+
+		return app, nil
+	})
+}
+
+// GetStripeApp gets an app
+func (a adapter) GetStripeApp(ctx context.Context, input appstripeentity.GetAppInput) (appstripeentity.App, error) {
```
turip, line 71: This should happen in the service.

openmeter/app/service.go
```diff
@@ -0,0 +1,30 @@
...
+	GetMarketplaceListing(ctx context.Context, input appentity.MarketplaceGetInput) (appentity.RegistryItem, error)
+	ListMarketplaceListings(ctx context.Context, input appentity.MarketplaceListInput) (pagination.PagedResponse[appentity.RegistryItem], error)
+	InstallMarketplaceListingWithAPIKey(ctx context.Context, input appentity.InstallAppWithAPIKeyInput) (appentity.App, error)
+	GetMarketplaceListingOauth2InstallURL(ctx context.Context, input appentity.GetOauth2InstallURLInput) (appentity.GetOauth2InstallURLOutput, error)
+	AuthorizeMarketplaceListingOauth2Install(ctx context.Context, input appentity.AuthorizeOauth2InstallInput) error
+
+	// Installed app
+	CreateApp(ctx context.Context, input appentity.CreateAppInput) (appentitybase.AppBase, error)
+	GetApp(ctx context.Context, input appentity.GetAppInput) (appentity.App, error)
+	GetDefaultApp(ctx context.Context, input appentity.GetDefaultAppInput) (appentity.App, error)
```
turip, line 27: This is more theoretical, but it is a decision how we want to do the defaulting.

Right now we have a default app per type, which allows us to support multiple apps from the same type, while not mandating the partner to store the application id.

I wonder if it makes sense to have a default app per capability (e.g. tax reporting, etc).

If not we should be prudent documenting the default behavior.

config/stripe.go
```diff
@@ -0,0 +1,44 @@
...
+	if c.BaseURL == "" {
+		errs = append(errs, errors.New("webhook base URL is required"))
+	}
+
+	return errors.Join(errs...)
+}
+
+// ConfigureStripe configures the default values for Stripe.
+func ConfigureStripe(v *viper.Viper) {
+	v.SetDefault("stripe.webhook.baseURL", "https://example.com")
```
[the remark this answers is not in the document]
sagikazarmark: I think it makes sense not to mix different kinds of stripe configs.

openmeter/app/adapter.go
```diff
@@ -0,0 +1,32 @@
+package app
```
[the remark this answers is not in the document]
sagikazarmark: Or `apps`?

I have a plan to move the app and config packages out from openmeter to somewhere else, but for now, that package name is taken.

openmeter/app/service.go
sagikazarmark: It's perfectly logical and fine to have multiple "apps" of the same type. AWS/GCP comes to mind for object storage export.

sagikazarmark (changes requested):
Please revert the generated client go change

api/spec/src/app/app.tsp
```diff
@@ -0,0 +1,113 @@
...
+@route("/api/v1/apps")
+@tag("Apps")
+interface Installed {
+  /**
+   * List apps.
+   */
+  @get
+  @route("/")
+  @operationId("listApps")
+  listApps(...ListAppsRequest): Paginated<App>[] | CommonErrors;
```
sagikazarmark, line 32: We tend to use simply `list`, `get`, `uninstall` for these

api/spec/src/app/app.tsp
```diff
@@ -0,0 +1,113 @@
...
+using TypeSpec.OpenAPI;
+
+namespace OpenMeter.App;
+
+/**
+ * Installed Apps API.
+ */
+@route("/api/v1/apps")
+@tag("Apps")
+interface Installed {
```
sagikazarmark, line 25:
```suggestion
interface Apps {
```

?

api/spec/src/app/app.tsp
```diff
@@ -0,0 +1,113 @@
...
+  @operationId("getApp")
+  getApp(@path id: ULID): App | CommonErrors;
+
+  /**
+   * Uninstall an app.
+   */
+  @delete
+  @route("/{id}")
+  @operationId("uninstallApp")
+  uninstallApp(@path id: ULID): App | NotFoundError | CommonErrors;
```
sagikazarmark, line 48: Delete shouldn't return anything. Returning 404 is also unnecessary.

api/spec/src/app/app.tsp
```diff
@@ -0,0 +1,113 @@
...
+}
+
+/**
+ * Query params for listing installed apps
+ */
+@friendlyName("ListAppsRequest")
+model ListAppsRequest extends PaginatedQuery {}
```
sagikazarmark, line 55: Should probably use `is`?

api/spec/src/app/app.tsp
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * App type.
+ */
+enum AppType {
```
sagikazarmark, line 60: Could just be `Type`?

api/spec/src/app/app.tsp
```diff
@@ -0,0 +1,113 @@
...
+}
+
+/**
+ * App.
+ * One of: stripe
+ */
+@friendlyName("App")
+@discriminator("type")
+union App {
+  @summary("Stripe")
```
sagikazarmark, line 71: I don't think summary like this is very useful.

api/spec/src/app/app.tsp
```diff
@@ -0,0 +1,113 @@
...
+   * Status of the app connection.
+   */
+  @visibility("read")
+  status: AppStatus;
+}
+
+/**
+ * App installed status.
+ */
+enum AppStatus {
```
sagikazarmark, line 101: We may need an unknown status for...well, unknown errors.

api/spec/src/app/main.tsp
```diff
@@ -0,0 +1,18 @@
...
+
+// Package Contents
+import "./app.tsp";
+import "./capability.tsp";
+import "./marketplace.tsp";
+import "./stripe.tsp";
+
+using TypeSpec.Http;
+using TypeSpec.Rest;
+using TypeSpec.OpenAPI;
```
sagikazarmark, line 16: This isn't required here IMO

api/spec/src/app/main.tsp
```diff
@@ -0,0 +1,18 @@
+import "@typespec/http";
+import "@typespec/rest";
+import "@typespec/openapi";
+import "@typespec/openapi3";
```
sagikazarmark, line 4: This isn't required here IMO

api/spec/src/app/main.tsp
```diff
@@ -0,0 +1,18 @@
...
+
+import "..";
```
sagikazarmark, line 6: If the app is imported in parent's main, I don't think this is necessary.

api/spec/src/app/marketplace.tsp
```diff
@@ -0,0 +1,164 @@
+import "@typespec/http";
+import "@typespec/rest";
+import "@typespec/openapi3";
+
+import "../errors.tsp";
+import "../oauth.tsp";
+import "../pagination.tsp";
+import "../types.tsp";
```
sagikazarmark, line 8: If the app is imported in parent's main, I don't think this is necessary.

api/spec/src/app/marketplace.tsp
```diff
@@ -0,0 +1,164 @@
+import "@typespec/http";
+import "@typespec/rest";
```
sagikazarmark, line 2: This is unused

api/spec/src/app/marketplace.tsp
```diff
@@ -0,0 +1,164 @@
...
+@route("/api/v1/marketplace")
+@tag("Apps")
+interface Marketplace {
+  /**
+   * List available apps of the app marketplace.
+   */
+  @get
+  @route("/listings")
+  @operationId("marketplaceListListings")
+  marketplaceListings(...PaginatedQuery): Paginated<Listing> | CommonErrors;
```
sagikazarmark, line 27: No need for the marketplace prefix. I would probably call this `listListings`

api/spec/src/app/marketplace.tsp
```diff
@@ -0,0 +1,164 @@
...
+  @query appKey: Key;
+}
+
+/**
+ * Marketplace OAuth2 install response.
+ */
+@friendlyName("MarketplaceOAuth2InstallAuthorizeResponse")
+model MarketplaceOAuth2InstallAuthorizeResponse {
+  @statusCode _: 303;
+}
```
sagikazarmark, line 81: I would just inline this if there isn't anything else in the model.

api/spec/src/app/marketplace.tsp
```diff
@@ -0,0 +1,164 @@
...
+    #{
+      type: CapabilityType.CollectPayments,
+      key: "stripe_collect_payments",
+      name: "Collect Payments",
+      description: "Stripe payments collects outstanding revenue with Stripe customer's default payment method.",
+      requirements: #[Requirements.CustomerExternalStripeCustomerId],
+    }
+  ],
+})
+model Listing {
```
sagikazarmark, line 139: MarketplaceListing?

api/spec/src/app/stripe.tsp
```diff
@@ -0,0 +1,345 @@
+import "@typespec/http";
+import "@typespec/rest";
+import "@typespec/openapi3";
+
+import "./app.tsp";
+import "./capability.tsp";
+import "../errors.tsp";
+import "../oauth.tsp";
+import "../pagination.tsp";
+import "../types.tsp";
```
sagikazarmark, line 10: If the app is imported in parent's main, I don't think this is necessary.

api/spec/src/app/stripe.tsp
```diff
@@ -0,0 +1,345 @@
+import "@typespec/http";
+import "@typespec/rest";
```
sagikazarmark, line 2: Unused

api/spec/src/app/stripe.tsp
```diff
@@ -0,0 +1,345 @@
...
+@route("/api/v1/apps/{id}/stripe")
+@tag("AppsStripe")
+interface AppStripe {
+  /**
+   * Stripe webhook.
+   */
+  @post
+  @route("/webhook")
+  @operationId("appStripeWebhook")
+  appStripeWebhook(@path id: ULID, @body body: StripeWebhookEvent): StripeWebhookResponse | CommonErrors;
```
sagikazarmark, line 31: Does this have to be a public API? Or is it just a webhook we register with Stripe?

api/spec/src/app/marketplace.tsp
```diff
@@ -0,0 +1,164 @@
...
+  /**
+   * Authorize OAuth2 code.
+   * Verifies the OAuth code and exchanges it for a token and refresh token
+   */
+  @get
+  @route("/install/oauth2/authorize")
+  @operationId("marketplaceOAuth2InstallAuthorize")
+  marketplaceOAuth2InstallAuthorize(
+    ...MarketplaceOAuth2InstallAuthorizeRequest,
+  ): MarketplaceOAuth2InstallAuthorizeResponse | CommonErrors;
```
sagikazarmark, line 49: I could be wrong here, but this isn't a public API per se, is it?

api/spec/src/oauth.tsp
```diff
@@ -0,0 +1,130 @@
+import "@typespec/http";
+
+namespace OAuth2;
```
sagikazarmark, line 3: I wonder if this is the best place for this, especially because it directly says "client" and "app" below. Shouldn't it just be in the apps package?

api/spec/src/main.tsp
```diff
@@ -8,6 +8,7 @@ import "./auth.tsp";
 import "./errors.tsp";
 import "./types.tsp";
 
+import "./app";
```
sagikazarmark, line 11: This feels like it should be "apps", not "app".

sagikazarmark (changes requested):
Conceptually, how is this going to work for an integration, like GCP and AWS where you have to setup a trust relationship and potentially have multiple different accounts?

openmeter/appstripe/adapter/stripe.go
hekike: I think it's fine

api/spec/src/app/stripe.tsp
hekike: yes, must be public so Stripe can call it. It handles webhooks.

openmeter/appstripe/adapter/adapter.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	// Register stripe app in marketplace
+	err = config.AppService.RegisterMarketplaceListing(appentity.RegistryItem{
+		Listing: appstripeentity.StripeMarketplaceListing,
+		Factory: adapter,
+	})
+	if err != nil {
+		return adapter, fmt.Errorf("failed to register stripe app: %w", err)
+	}
```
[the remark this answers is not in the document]
hekike: I can't register the app service here as the app service depends on the app adapter.
In that case, registration has to happen explicitly outside of adapter and service.

api/spec/src/app/app.tsp
hekike: Using`is` is helpful to mark the type the exact same. This is true in its current state, but I'm using extend because I am planning to add filters like is deleted, app type, etc.

api/spec/src/app/main.tsp
hekike: It breaks without it

openmeter/appstripe/adapter/adapter.go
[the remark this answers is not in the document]
turip: Ok, I will take a look into this tomorrow more in depth as this violates the usual contract we have.

openmeter/app/adapter.go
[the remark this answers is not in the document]
hekike: Discussed to go with `app` for now and move out existing `app` package related to Wire.
In the future we plan to rename packages to plurals.

openmeter/app/service.go
hekike: We discussed that we keep per app type default to make onboarding a smoother experience.

api/spec/src/main.tsp
hekike: As we keep the Go pkg singular for now, keeping this one too.

api/spec/src/app/marketplace.tsp
hekike: The OAuth provider calls it with the authorization token, we exchange it for refresh token and redirect from it. So it's public.

api/spec/src/app/stripe.tsp
hekike: Correct. It's currently not imported in `app/main.tsp`. Would it be preferred to import them there instead of per file?

api/spec/src/app/app.tsp
hekike: I'll add when we cross that bridge.

turip (approved):

hekike:
Closing in favor of: https://github.com/openmeterio/openmeter/pull/1684

PR #1649, opened 2024-10-10
fix(api): stripeCustomerId to be nullable

chrisgacsal:
## Overview

Allow `stripeCustomerId` to be `nullable: true` in order to ensure API backward compatibility.


api/spec/src/cloud/subjects.tsp
```diff
@@ -80,5 +80,5 @@ model SubjectUpsert {
   currentPeriodEnd?: DateTime;
 
   @example("cus_JMOlctsKV8")
-  stripeCustomerId?: string;
+  stripeCustomerId?: string | null;
```
sagikazarmark, line 83: Probably should be changed in the Subject model as well?

chrisgacsal: I do not think it would be really necessary, but I'll add it there as well in order to be on the safe side.

sagikazarmark (approved):

PR #1687, opened 2024-10-16
feat: name collusion causes prefix drop

turip:
## Overview

Fixes #(issue)

PR #1688, opened 2024-10-16
fix: mandate enum prefixes for generated APIs

turip:
## Overview

This disables the heuristics of oapi codegen thus we get stable names.

context: https://github.com/openmeterio/openmeter/pull/1687


api/api.gen.go
```diff
@@ -126,7 +132,7 @@ const (
 
 // Defines values for NotificationChannelWebhookCreateRequestType.
 const (
-	WEBHOOK NotificationChannelWebhookCreateRequestType = "WEBHOOK"
+	NotificationChannelWebhookCreateRequestTypeWEBHOOK NotificationChannelWebhookCreateRequestType = "WEBHOOK"
```
turip, line 135: This is only used for deserializing/serializing by the API itself, so it should be fine to leave it like this.

PR #1703, opened 2024-10-18
feat: billing profiles

turip:
## Overview

This patch adds support for billing profiles as the first implemented API.

For testing purposes, a new OpenMeter Sandbox is added. The sandbox app gets installed by default if apps are enabled for easier testing.

The change also makes sure that the customeroverrides codebase works as expected, however we are going to implement the APIs at a later point.

## Structural changes

The entities are moved to a separate entities package. Transaction handling added for the billing profiles.

## Notes for reviewer

This PR is longer than usual, as the original codebase was created without the finalized API, so I needed to perform some refactoring on the existing backend code for billing.


app/config/config.go
```diff
@@ -32,6 +32,8 @@ type Configuration struct {
 	Sink          SinkConfiguration
 	BalanceWorker BalanceWorkerConfiguration
 	Notification  NotificationConfiguration
+	Billing       BillingConfiguration
```
hekike, line 35: nit: alphabetical order

openmeter/app/sandbox/helpers.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	if a.AppService == nil {
+		return errors.New("app service is required")
+	}
+
+	return nil
+}
+
+func AutoProvision(ctx context.Context, input AutoProvisionInput) (appentity.App, error) {
```
hekike, line 32: please add comment what auto provision does

openmeter/billing/service/profile.go
```diff
@@ -38,7 +43,35 @@ func (s *Service) CreateProfile(ctx context.Context, input billing.CreateProfile
...
+		if err != nil {
+			return nil, billing.ValidationError{
+				Err: fmt.Errorf("error resolving tax app: %w", err),
+			}
+		}
+
+		invocingApp, err := s.validateAppReference(ctx, input.Namespace, input.Apps.Invoicing, appentitybase.CapabilityTypeInvoiceCustomers)
+		if err != nil {
+			return nil, billing.ValidationError{
+				Err: fmt.Errorf("error resolving tax app: %w", err),
```
hekike, line 57: copy pasta

openmeter/billing/service/profile.go
```diff
@@ -38,7 +43,35 @@ func (s *Service) CreateProfile(ctx context.Context, input billing.CreateProfile
...
+			}
+		}
+
+		paymentsApp, err := s.validateAppReference(ctx, input.Namespace, input.Apps.Payment, appentitybase.CapabilityTypeCollectPayments)
+		if err != nil {
+			return nil, billing.ValidationError{
+				Err: fmt.Errorf("error resolving tax app: %w", err),
```
hekike, line 64: copy pasta

pkg/datex/date.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	return lo.ToPtr(d), nil
+}
+
+type Period struct {
+	period.Period
+}
+
+// FromDuration creates an IMPRECISE Period from a time.Duration
```
hekike, line 38: why is it IMPRECISE?

api/spec/src/billing/profile.tsp
```diff
@@ -78,25 +78,55 @@ interface Profiles {
...
+  apps: ProfileApps;
+
   @summary("Is this the default profile?")
   default: boolean;
 }
 
+/**
+ * ProfileApps represents the applications used by a billing profile
+ */
+@friendlyName("BillingProfileApps")
```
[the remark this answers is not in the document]
turip: Yes, but based on notifications and other parts of the code we are better off adding friendlyName.

For example in this case the auto-generated name would be OpenMeterBilling_ProfileApps, but we don't need the openmeter prefix.

Or in case of the AppType that would become OpenMeterApp_AppType. If we drop the OpenMeter form the namespace we should be able to get aways with some renames and not having friendly name IMHO, but for the time being I would prefer to keep the friendlyNames. cc @tothandras this is something to think about.

api/spec/src/billing/profile.tsp
```diff
@@ -116,67 +146,78 @@ model Workflow {
...
-
-  @summary("The tax app used for this workflow")
-  taxApp: OpenMeter.App.App;
-
-  @summary("The invoicing app used for this workflow")
-  invoicingApp: OpenMeter.App.App;
-
-  @summary("The payment app used for this workflow")
-  paymentApp: OpenMeter.App.App;
+  interval?: string = "PT1H";
```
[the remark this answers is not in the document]
turip: ISO8601 representation of the 1 hour period:
- P: Period
- T: Time part (empty date part)
- 1: amount
- H: hours

app/config/config.go
turip: I can order that in a seperate PR.

pkg/datex/date.go
turip: Right, sorry, with the Normalize(false) this will not be imprecise.

turip:
TODO: If we have any Response object rename it to result.

hekike (approved):

PR #1720, opened 2024-10-21
feat(streaming): raw event connector

hekike:
As of today, OpenMeter uses ClickHouse materialized views to track meters. Events are inserted into a raw events table on which MVs are based. This PR moves the event insert logic out from the sink worker to the streaming connector and splits the connector into two different connectors: 

1. Raw events connector: a low-level connector inserts and queries only raw events. Using this connector is expensive to query meters, as will JSON parse each event.
3. Materialized View connector that extends raw event connector to speed up querying meters. (current default)

PR review guide:

- ./openmeter/streaming/clickhouse/materialized_view/meter_query and test is unchanged just moved around
- ./openmeter/streaming/clickhouse/raw_events/event.query and test is unchanged; just moved around


app/common/openmeter.go
```diff
@@ -21,7 +21,10 @@ import (
 	"github.com/openmeterio/openmeter/openmeter/namespace"
 	"github.com/openmeterio/openmeter/openmeter/sink/flushhandler"
 	"github.com/openmeterio/openmeter/openmeter/sink/flushhandler/ingestnotification"
-	"github.com/openmeterio/openmeter/openmeter/streaming/clickhouse_connector"
+	"github.com/openmeterio/openmeter/openmeter/streaming"
+	"github.com/openmeterio/openmeter/openmeter/streaming/clickhouse_connector_mv"
+	"github.com/openmeterio/openmeter/openmeter/streaming/clickhouse_connector_parse"
+	"github.com/openmeterio/openmeter/openmeter/streaming/clickhouse_connector_raw"
```
sagikazarmark, line 27: While you are at it:

- calling the package `clickhouse`
- and hosting all three implementations in the same package

Would be much nicer.

app/config/aggregation.go
```diff
@@ -4,20 +4,64 @@ import (
...
 type AggregationConfiguration struct {
+	// Engine is the aggregation engine to use
+	Engine     AggregationEngine
 	ClickHouse ClickHouseAggregationConfiguration
+
+	// Set true for ClickHouse first store the incoming inserts into an in-memory buffer
+	// before flushing them regularly to disk.
+	// See https://clickhouse.com/docs/en/cloud/bestpractices/asynchronous-inserts
+	AsyncInsert bool
+	// Set true if you want an insert statement to return with an acknowledgment immediatelyy
```
sagikazarmark, line 45:
```suggestion

	// Set true if you want an insert statement to return with an acknowledgment immediatelyy
```

app/config/aggregation.go
```diff
@@ -4,20 +4,64 @@ import (
...
+	if !slices.Contains(e.Values(), e) {
+		return fmt.Errorf("invalid value")
+	}
+	return nil
+}
+
 type AggregationConfiguration struct {
+	// Engine is the aggregation engine to use
+	Engine     AggregationEngine
 	ClickHouse ClickHouseAggregationConfiguration
```
sagikazarmark, line 39: Long term we should probably move the clickhouse config to top level and better organize engine configs.

This change only increases the chaos as it is with aggregation level, but clickhouse specific configs. :/

cmd/server/main.go
```diff
@@ -333,7 +333,12 @@ func main() {
 	})
 
 	for _, meter := range conf.Meters {
-		err := app.StreamingConnector.CreateMeter(ctx, app.NamespaceManager.GetDefaultNamespace(), meter)
+		if meter == nil {
+			logger.Error("meter configuration is nil")
+			os.Exit(1)
+		}
```
sagikazarmark, line 339: Shouldn't this be caught by config validation?

openmeter/meter/parse.go
```diff
@@ -0,0 +1,93 @@
...
+	"strconv"
+
+	"github.com/cloudevents/sdk-go/v2/event"
+	"github.com/oliveagle/jsonpath"
+
+	"github.com/openmeterio/openmeter/pkg/models"
+)
+
+// ParseEvent validates and parses an event against a meter.
+func ParseEvent(meter Meter, ev event.Event) (*float64, *string, map[string]string, error) {
```
sagikazarmark, line 15: Why is validate being replaced by parsing?

sagikazarmark (commented):
Why did you start passing `meter models.Meter` to some functions and kept `meterSlug` in others? Why start doing it at all? Can't the connector query the meter from the repository?

app/config/aggregation.go
chrisgacsal: `s/immediatelyy/immediately/`

openmeter/meter/parse.go
hekike: The new connector required to know value. It was always parsing, just before didn't return the parsed value. This file looks new but it's mostly renamed.  I removed the new connector from the PR but would like to keep parsing this way as it helps testing.

app/config/aggregation.go
```diff
@@ -100,6 +175,11 @@ func (c ClickHouseAggregationConfiguration) GetClientOptions() *clickhouse.Optio
 
 // ConfigureAggregation configures some defaults in the Viper instance.
 func ConfigureAggregation(v *viper.Viper) {
+	v.SetDefault("aggregation.engine", AggregationEngineClickHouseMV)
```
hekike, line 178: probably we should change default to raw events

turip (approved):

PR #1730, opened 2024-10-21
feat: add otel log support

sagikazarmark:
## Overview

feat: add otel log support


sagikazarmark:
Potentially fixed by #1503

sagikazarmark:
Currently we open a separate connection for each otlp exporter (trace, metrics, log). I will change that in a subsequent PR.

chrisgacsal (approved):

PR #1731, opened 2024-10-21
feat(ingest): adopt topic provisioner

chrisgacsal:
## Overview

Use `TopicProvisioner` to ensure that the target topic for events is available on Kafka cluster in case Kafka driver is used for ingest. Provisioning results are caches i norder to avoid unnecessary roundtrips to between service and Kafka cluster.

### DIFF
--- a/app/common/openmeter.go
+++ b/app/common/openmeter.go
@@ -64,11 +64,18 @@ func NewNamespacedTopicResolver(config config.Configuration) (*topicresolver.Nam
 	return topicResolver, nil
 }
 
-func NewKafkaIngestCollector(producer *kafka.Producer, topicResolver topicresolver.Resolver) (*kafkaingest.Collector, error) {
+func NewKafkaIngestCollector(
+	config config.KafkaIngestConfiguration,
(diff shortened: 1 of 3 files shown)


turip (approved):

app/common/openmeter.go
```diff
@@ -64,11 +64,18 @@ func NewNamespacedTopicResolver(config config.Configuration) (*topicresolver.Nam
 	return topicResolver, nil
 }
 
-func NewKafkaIngestCollector(producer *kafka.Producer, topicResolver topicresolver.Resolver) (*kafkaingest.Collector, error) {
+func NewKafkaIngestCollector(
+	config config.Configuration,
```
sagikazarmark, line 68: This should be limited either to ingest or ingest.kafka.

sagikazarmark (changes requested):

app/common/openmeter.go
chrisgacsal: Agreed. Fixed.

sagikazarmark (approved):

PR #1738, opened 2024-10-22
feat: pending line item creation

turip:
## Overview

This patch adds support for creating new pending line items. This will be the basic flow orchestrated by an adapter to the subscriptions.

The patch contains the invoice mapping as that's needed for troubleshooting purposes.

We are also removing the gobl driver as we are moving away from using gobl as is.


api/spec/src/billing/invoices.tsp
```diff
@@ -398,7 +404,17 @@ enum InvoiceExpand {
 @friendlyName("BillingInvoiceCreateInput")
 model InvoiceCreateInput {
   /**
-   * The pending items to include in the invoice, if not provided all pending items will be included.
+   * The pending line items to include in the invoice, if not provided:
+   * - all line items that have invoice_at < now() will be included
```
hekike, line 408: up to AsOf, right?

api/spec/src/billing/invoices.tsp
```diff
@@ -398,7 +404,17 @@ enum InvoiceExpand {
...
+   */
+  IncludePendingLines?: ULID[];
+
+  /**
+   * The time as of which the invoice is created.
+   *
+   * If not provided, the current time is used.
    */
-  IncludePendingItems?: ULID[];
+  AsOf?: DateTime;
```
hekike, line 419: Defaults to now.

openmeter/billing/customeroverride.go
```diff
@@ -117,3 +117,42 @@ type GetProfileWithCustomerOverrideInput namespacedCustomerID
 func (g GetProfileWithCustomerOverrideInput) Validate() error {
 	return namespacedCustomerID(g).Validate()
 }
+
+type GetCustomerOverrideAdapterInput struct {
+	Namespace  string
+	CustomerID string
```
hekike, line 123: should this be `customerentity.CustomerID`? Which  is basically namespace + customerId and has a Validate method

openmeter/billing/entity/invoice.go
```diff
@@ -82,18 +96,31 @@ func (s InvoiceStatus) Values() []string {
...
+
+	return true
 }
 
 type Invoice struct {
 	Namespace string `json:"namespace"`
 	ID        string `json:"id"`
 
-	InvoiceNumber InvoiceNumber `json:"invoiceNumber"`
+	Number *string `json:"invoiceNumber"`
```
hekike, line 123: nit: we in json it's prefixed with invoice.

openmeter/billing/entity/invoiceline.go
```diff
@@ -0,0 +1,157 @@
...
+	InvoiceLineStatusValid InvoiceLineStatus = "valid"
+	// InvoiceLineStatusDeleted is a deleted invoice line.
+	InvoiceLineStatusDeleted InvoiceLineStatus = "deleted"
+	// InvoiceLineStatusSplit is a split invoice line (the child lines will have this set as parent).
+	InvoiceLineStatusSplit InvoiceLineStatus = "split"
+)
+
+func (InvoiceLineStatus) Values() []string {
+	return []string{
+		string(InvoiceLineStatusValid),
```
hekike, line 45: misses other statuses

openmeter/ent/schema/billing.go
```diff
@@ -315,76 +333,128 @@ type BillingInvoice struct {
...
-		field.String("series").
+		// Customer/supplier
+		field.String("supplier_name").
+			NotEmpty(),
+
+		field.String("supplier_tax_code").
+			Optional().
+			Nillable(),
+
+		field.String("customer_name").
```
hekike, line 360: do we need both supplier and customer name?
Customer name is also confusing as we have customer entity.

api/spec/src/billing/invoices.tsp
turip: Only scalar types can be defaulted, DateTime is not a scalar type.

openmeter/ent/schema/billing.go
turip: By definition the invoice must clone the state of the customer and the supplier latest at the time the invoice is sent to the customer, so that it contains the values at the time of invoice creation and not at current time.

This is why we are cloning the objects.

If this wouldn't be the case a change in the customer entity would be retroactively refracted in the already sent invoices.

api/spec/src/billing/customeroverride.tsp
```diff
@@ -54,24 +54,24 @@ interface CustomerOverrides {
...
    */
   @delete
-  @route("/{customerIdOrKey}")
+  @route("/{customerId}")
   @summary("Delete a customer override")
   @operationId("billingDeleteCustomerOverride")
   delete(
     @path
-    customerIdOrKey: ULIDOrKey,
+    customerId: ULIDOrKey,
```
hekike, line 74: `: ULID`

api/spec/src/billing/invoices/invoice.tsp
```diff
@@ -240,27 +267,57 @@ enum LineTypes {
...
   total: Numeric;
 
   @summary("Period of the line item applies to for revenue recognition pruposes.")
   period: Period;
+
+  @summary("The time this line item should be invoiced.")
+  invoiceAt: DateTime;
+
+  @summary("Quantity of the item being sold.")
+  quantity?: Numeric;
```
hekike, line 310: shouldn't this default to one? Does it make sense selling something without quantity?

openmeter/billing/adapter/invoicelines.go
```diff
@@ -0,0 +1,119 @@
...
+				return nil, err
+			}
+
+			newEnt = newEnt.SetBillingInvoiceManualLines(newManualLineConfig)
+			edges.BillingInvoiceManualLines = newManualLineConfig
+		default:
+			return nil, fmt.Errorf("unsupported type: %s", line.Type)
+		}
+
+		savedLine, err := newEnt.Save(ctx)
```
hekike, line 55:
```
var linesToCreate ent.BillingInvoiceLine[]
for _, line := range input.Lines {
   linesToCreate = append(linesToCreate, r.db.BillingInvoiceLine.Create()
      ...
    )
}

savesLines, err := newEnt.CreateBulk(...linesToCreate)
```
https://entgo.io/docs/crud/#create-many

openmeter/billing/entity/invoice.go
```diff
@@ -117,19 +144,26 @@ type Invoice struct {
 	// Customer is either a snapshot of the contact information of the customer at the time of invoice being sent
 	// or the data from the customer entity (draft state)
 	// This is required so that we are not modifying the invoice after it has been sent to the customer.
-	Profile  Profile         `json:"profile"`
-	Customer InvoiceCustomer `json:"customer"`
+	Customer InvoiceCustomer  `json:"customer"`
+	Supplier SupplierContact  `json:"supplier"`
+	Workflow *InvoiceWorkflow `json:"workflow"`
```
hekike, line 149: omitempty?

openmeter/billing/entity/invoiceline.go
```diff
@@ -0,0 +1,173 @@
...
+	ID        string `json:"id"`
+
+	CreatedAt time.Time  `json:"createdAt"`
+	UpdatedAt time.Time  `json:"updatedAt"`
+	DeletedAt *time.Time `json:"deletedAt,omitempty"`
+
+	Metadata    map[string]string `json:"metadata"`
+	Name        string            `json:"name"`
+	Type        InvoiceLineType   `json:"type"`
+	Description *string           `json:"description"`
```
hekike, line 81: omitempty?

hekike (approved):

api/spec/src/billing/invoices/invoice.tsp
turip: It's not mandated due to the ubp story. We need qty but in that case it doesn't need to be set during creation.

openmeter/billing/adapter/invoicelines.go
turip: I have created a ticket on this, but given that I have to create a lot of additional objects, I wouldn't do that refactor just now. (We are not creating 10000s of records).

hekike (approved):

PR #1739, opened 2024-10-22
Use transacting repos for customer package

GAlexIHU:
## Overview

Transactions were previously ignored by the package due to this.

### DIFF
--- a/openmeter/app/adapter/adapter.go
+++ b/openmeter/app/adapter/adapter.go
@@ -9,6 +9,7 @@ import (
 	"github.com/openmeterio/openmeter/openmeter/app"
 	appentity "github.com/openmeterio/openmeter/openmeter/app/entity"
 	appentitybase "github.com/openmeterio/openmeter/openmeter/app/entity/base"
+	"github.com/openmeterio/openmeter/openmeter/ent/db"
 	entdb "github.com/openmeterio/openmeter/openmeter/ent/db"
 	"github.com/openmeterio/openmeter/pkg/framework/entutils"
 	"github.com/openmeterio/openmeter/pkg/framework/transaction"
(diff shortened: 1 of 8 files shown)


hekike:
Thank you!
It's probably failing because the App uses `transaction.Run` and is not the same TX. When should we use one over other? https://github.com/openmeterio/openmeter/blob/main/openmeter/app/adapter/app.go#L21

GAlexIHU:
`entutils.TransactingRepo` has a simple purpose: it reinstantiates the database client using the transaction passed along (if none is present in the context then it creates a new locally scoped transaction).

https://github.com/openmeterio/openmeter/blob/main/pkg/framework/entutils/transaction.go#L198-L214

The reason this is needed is because otherwise ent couldn't know that we want to execute the queries in this current transaction. An example use of it would be here: https://github.com/openmeterio/openmeter/blob/main/openmeter/entitlement/adapter/entitlement.go#L52-L73

All repository methods should be wrapped in in this (even simple reads) due to transaction isolation: one call might not see the effects of other calls (thats why the tests were failing).

The intended use case would be (as in the example above) that the repository only does DB calls & all of its methods are wrapped in this to ensure everything uses the same transaction if one is present.

In the case of `app`, `customer` and `stripeapp`, the `adapter` package is heavily overburdened while the `service` is mostly empty, so I had to wrap almost every operation is `entutils.TransactingRepo` as db calls were made everywhere, but in general for service layer things `transaction.Run` should be fine, it just expresses that these should happen atomically on the same PG transaction. (In practice if you check the implementation, `entutils.TransactingRepo` also uses `transaction.Run`)

turip (approved):

turip:
Thank you <3

hekike (approved):

PR #1740, opened 2024-10-22
refactor(kafka): topic provisioning

chrisgacsal:
## Overview

* allow defining list of topics which are protected from accidental deletion using TopicProvisioner.
* allowing namespace to be deleted by kafka namespace handler requires the `ingest.kafka.namespaceDeletionEnabled` configuration parameter explcitly set to `true`. It is set to `false` by default.

### DIFF
--- a/pkg/kafka/topicprovisioner.go
+++ b/pkg/kafka/topicprovisioner.go
@@ -37,8 +37,13 @@ type TopicProvisioner interface {
 	DeProvision(ctx context.Context, topics ...string) error
 }
 
+type AdminClient interface {
+	CreateTopics(ctx context.Context, topics []kafka.TopicSpecification, options ...kafka.CreateTopicsAdminOption) ([]kafka.TopicResult, error)
+	DeleteTopics(ctx context.Context, topics []string, options ...kafka.DeleteTopicsAdminOption) ([]kafka.TopicResult, error)
+}
+
 type TopicProvisionerConfig struct {
-	AdminClient *kafka.AdminClient
+	AdminClient AdminClient
 	Logger      *slog.Logger
 	Meter       metric.Meter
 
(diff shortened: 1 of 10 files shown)


pkg/kafka/topicprovisioner.go
```diff
@@ -65,9 +73,15 @@ func NewTopicProvisioner(config TopicProvisionerConfig) (TopicProvisioner, error
 		return nil, errors.New("meter is required")
 	}
 
+	protectedTopics := make(map[string]struct{}, len(config.ProtectedTopics))
```
hekike, line 76: nit: why isn't this bool like? It's a set, right?
`protectedTopicsByTopicName := map[string]bool`

pkg/kafka/topicprovisioner.go
```diff
@@ -233,13 +249,25 @@ func (p *topicProvisioner) DeProvision(ctx context.Context, topics ...string) er
 		return nil
 	}
 
+	topicsToDelete := make([]string, 0, len(topics))
 	for _, topic := range topics {
 		if topic == "" {
-			return fmt.Errorf("invalid topic name %q", topic)
+			p.logger.Warn("skip topic: empty topic name", "topic", topic)
```
hekike, line 255: why can this be empty?

hekike (approved):

pkg/kafka/topicprovisioner.go
chrisgacsal: Yeah, it is a set, but empty `struct{}` does not require memory allocation while `bool` does. Thats why I prefer `map[string]struct{}` over `map[string]bool`.

pkg/kafka/topicprovisioner.go
chrisgacsal: Normally it should not be, but nothing prevents one to make call like:

```go
proivisioner.DeProvision(ctx, []string{"topic-1", "", "topic-3"})
```

I wanted to make sure that empty topic names are at least logged, as it may indicate an issue in the code.

PR #1748, opened 2024-10-24
feat(compose): add ch-ui

hekike:
Adding ch-ui to dev-profile.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -47,6 +47,21 @@ services:
       timeout: 3s
       retries: 30
 
+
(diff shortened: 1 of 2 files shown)


docker-compose.yaml
```diff
@@ -47,6 +47,21 @@ services:
       timeout: 3s
       retries: 30
 
+
+  ch-ui:
+    image: ghcr.io/caioricciuti/ch-ui:latest
```
hekike, line 52: they don't tag their releases

PR #1767, opened 2024-10-29
fix: treat inactive entitlement as deleted

turip:
## Overview

When an entitlement is not active anymore, the recalculator yields this error.

```
error sending event for entitlement [id=XXX]: failed to map entitlement value: unknown entitlement type
```


GAlexIHU (approved):
Nice job

PR #1772, opened 2024-10-31
feat: validation errors api

turip:
## Overview

Given that most of the invoice operations will be done in the background, we need to capture the validation errors. This patch does it in a way that is native to go's error handling: an error can be parsed into multiple validation errors, if an error is not a validationissue typed error or not wrapped into any validation issue modifier it is treated as a normal error.

The state machine honors the errors: if there are critical errors that's considered a failed state.

The validation errors are available on the invoice endpoint.


openmeter/ent/schema/billing.go
```diff
@@ -502,3 +503,62 @@ func (BillingInvoice) Edges() []ent.Edge {
...
+}
+
+func (BillingInvoiceValidationIssue) Edges() []ent.Edge {
+	return []ent.Edge{
+		edge.From("billing_invoice", BillingInvoice.Type).
+			Ref("billing_invoice_validation_issues").
+			Field("invoice_id").
+			Unique().
+			Required(),
+	}
```
hekike, line 563: nit: cascade delete

hekike (approved):

openmeter/ent/schema/billing.go
turip: Thanks, will add it to the next PR.

PR #1787, opened 2024-11-05
feat: support usage based pricing

turip:
## Overview

Add line-splitting capability to support mid-period invoicing. See README.md for more detailed information.

## Algorithm

To achieve the behavior described above, we are using line splitting. By default we would have one line per billing period that would eventually be part of an invoice:

```
                period.start                                              period.end
Line1 [status=valid] |--------------------------------------------------------|
```

When the usage-based line can be billed mid-period, we `split` the line into two:

```
                     period.start              asOf                              period.end
Line1 [status=split]         |--------------------------------------------------------|
SplitLine1 [status=valid]    |------------------|
SplitLine2 [status=valid]                       |-------------------------------------|
```

As visible:
- Line1's status changes from `valid` to `split`: it will be ignored in any calculation, it becomes a grouping line between invoices
- SplitLine1 is created with a period between `period.start` and `asof` (time of invoicing): it will be addedd to the freshly created invoice
- SplitLine2 is created with a period between `asof` and `period.end`: it will be pushed to the gathering invoice

When creating a new invoice between `asof` and `period.end` the same logic continues, but without marking SplitLine2 `split`, instead the new line is added to the original line's parent line:

```
                     period.start              asOf1          asof2                period.end
Line1 [status=split]         |--------------------------------------------------------|
SplitLine1 [status=valid]    |------------------|
SplitLine2 [status=valid]                       |---------------|
SplitLine3 [status=valid]                                       |---------------------|
```

This flattening approach allows us to not to have to recusively traverse lines in the database.


openmeter/billing/service/service.go
```diff
@@ -50,6 +56,15 @@ func (c Config) Validate() error {
 		return errors.New("logger cannot be null")
 	}
 
+	// TODO: breaking, we need to update main and cloud accordingly
```
turip, line 59: Let's add this to main.

openmeter/billing/adapter/invoicelines.go
```diff
@@ -63,50 +73,197 @@ func (r *adapter) CreateInvoiceLines(ctx context.Context, input billing.CreateIn
...
 	}
 
 	return result, nil
 }
 
+func (r *adapter) createManualUsageBasedLine(ctx context.Context, ns string, line billingentity.Line) (*db.BillingInvoiceManualUsageBasedLineConfig, error) {
+	lineConfig, err := r.db.BillingInvoiceManualUsageBasedLineConfig.Create().
+		SetNamespace(ns).
+		SetPriceType(line.ManualUsageBased.Price.Type()).
+		SetPrice(&line.ManualUsageBased.Price).
```
hekike, line 91: This should be called `Amount`, no?

openmeter/billing/entity/defaults.go
```diff
@@ -0,0 +1,7 @@
+package billingentity
+
+import "time"
+
+const (
+	DefaultMeterResolution = time.Minute
```
turip, line 6: This will be removed as soon as we have better granularity.

openmeter/billing/README.md
```diff
@@ -0,0 +1,131 @@
...
+SplitLine1 [status=valid]    |------------------|
+SplitLine2 [status=valid]                       |---------------|
+SplitLine3 [status=valid]                                       |---------------------|
+```
+
+This flattening approach allows us to not to have to recusively traverse lines in the database.
+
+### Usage-based quantity
+
+When a line is created for an invoice, the quantity of the underlying merter is captured into the line's qty field. This information is never updated, so late events will have to create new invoice lines when needed.
```
hekike, line 89: typo /s/merter/meter

openmeter/billing/adapter/invoice.go
```diff
@@ -246,6 +246,7 @@ func (r *adapter) CreateInvoice(ctx context.Context, input billing.CreateInvoice
 		SetNillableDueAt(input.DueAt).
 		SetNillableCustomerTimezone(customer.Timezone).
 		SetNillableIssuedAt(lo.EmptyableToPtr(input.IssuedAt)).
+		SetCustomerSubjectKeys(input.Customer.UsageAttribution.SubjectKeys).
```
hekike, line 249: Just a comment: in the future, we should be able to attribute usage to customers based on different parameters.
Maybe in the future this field would be better to be json store usage attribution like: `
```
.SetCustomerUsageAttribution(input.Customer.UsageAttribution)
```

openmeter/billing/service/lineservice/linebase.go
```diff
@@ -0,0 +1,229 @@
...
+			ParentLine: &parentLine,
+			Status:     billingentity.InvoiceLineStatusValid,
+			PeriodEnd:  splitAt,
+			InvoiceAt:  splitAt,
+		})
+
+		postSplitAtLine := l.CloneForCreate(UpdateInput{
+			ParentLine:  &parentLine,
+			Status:      billingentity.InvoiceLineStatusValid,
+			PeriodStart: splitAt,
```
hekike, line 190: just double checking: is this is also splitAt and and previous ends at splitAt one should be exclusive when we query usage

openmeter/billing/service/lineservice/manualusagebasedline.go
```diff
@@ -0,0 +1,137 @@
...
+		return nil, err
+	}
+
+	meter := meterAndFactory.meter
+
+	asOfTruncated := asof.Truncate(billingentity.DefaultMeterResolution)
+
+	switch meter.Aggregation {
+	case models.MeterAggregationSum, models.MeterAggregationCount,
+		models.MeterAggregationMax, models.MeterAggregationUniqueCount:
```
hekike, line 78: I'm not sure Max works here. Max of period chunk is different then max of whole period

openmeter/billing/service/lineservice/meters.go
```diff
@@ -0,0 +1,131 @@
...
+	// Validation
+	if err := in.Validate(); err != nil {
+		return nil, err
+	}
+
+	meterQueryParams := &streaming.QueryParams{
+		Aggregation:   in.Meter.Aggregation,
+		FilterSubject: in.Subjects,
+		From:          &in.Line.Period.Start,
+		To:            &in.Line.Period.End,
```
hekike, line 76: feature has group by filters that should be applied here too

openmeter/billing/service/lineservice/manualusagebasedline.go
turip: Yes, sorry, I was in a bit of rush for the meters, I have updated the docs and the algorithm.

openmeter/billing/service/lineservice/meters.go
turip: It's already included here: https://github.com/openmeterio/openmeter/pull/1787/files#diff-12d9d74ba62a6275562925d58ede10dfb9708c723129ef2b114e2053efc63473R79-R84

openmeter/billing/service/lineservice/linebase.go
turip: The plan was that:

```
// Period represents a time period, in billing the time period is always interpreted as
// [start, end) (i.e. start is inclusive, end is exclusive).
```

hekike (approved):

PR #1803, opened 2024-11-08
Update Helm Chart with new services & features

GAlexIHU:
## Overview

TLDR: Updates the Helm Chart adding the new capabilities.

- adds balance worker and notification services
- adds new dependencies (pg, redis, svix)
- initContainers to wait for dependency services
- fixes previous OpenMeter config merging (beforehand a lot of provided config wasn't honoured when using self hosted services)
- update docs

## Notes for reviewer
- I changed the Kafka used to a simpler version via bitnami. The reason for this is not ideal, I simply couldn't get it to work for esoteric reasons. This has the user impact that a migration path from the old chart version to the new should be defined by us (wouldn't be an issue for fresh installs). I also added further warnings that using the included dependencies in production is strongly discouraged. I think this naturally folds into a larger conversation about whether we want to provide an "everything included" experience with this helm chart even for production deployment, so on this topic I'd ask for some guidance from  @tothandras and @hekike


hekike:
Our strong recommendation is not to run your databases and Kafka in production but instead BYO.

tothandras (approved):

PR #1806, opened 2024-11-08
chore: disable dagger cloud

turip:
## Overview

Seems like we have an integration issue with Dagger Cloud. Temporarily disabling reporting until the issue has been sorted out.

Reverting this PR reproduces the issue, we can use that to validate any solution.


grouville:
Thanks, checking atm 👀

PR #1808, opened 2024-11-08
refactor: add TX support to Plan API

chrisgacsal:
## Overview

Add database transaction support to Plan API.

### DIFF
--- a/openmeter/productcatalog/plan/service/phase.go
+++ b/openmeter/productcatalog/plan/service/phase.go
@@ -5,115 +5,34 @@ import (
 	"fmt"
 
 	"github.com/openmeterio/openmeter/openmeter/productcatalog/plan"
+	"github.com/openmeterio/openmeter/pkg/framework/transaction"
 	"github.com/openmeterio/openmeter/pkg/pagination"
 )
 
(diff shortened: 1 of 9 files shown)


openmeter/productcatalog/plan/service/phase.go
```diff
@@ -5,140 +5,162 @@ import (
...
+	fn := func(ctx context.Context) (pagination.PagedResponse[plan.Phase], error) {
+		if err := params.Validate(); err != nil {
+			return pagination.PagedResponse[plan.Phase]{}, fmt.Errorf("invalid list PlanPhases params: %w", err)
+		}
+
+		return s.adapter.ListPhases(ctx, params)
 	}
 
-	return s.adapter.ListPhases(ctx, params)
+	return transaction.Run(ctx, s.adapter, fn)
```
hekike, line 21: I think you need to use `entutils.TransactingRepo` no? cc @GAlexIHU

turip (approved):

openmeter/productcatalog/plan/service/phase.go
turip: Nope, the original idea is that each function is wrapped in adapter in the transacting repo, while in service you define the transactions. In billing I am using the transacting repo, as it achieves the same as long as the adapter is not called directly.

I might or might not change billing to have this format later.

PR #1823, opened 2024-11-11
Refactor plan types 

GAlexIHU:
## Overview

As we're pushing for shared type use between different parts of `productcatalog` it makes life easier to separate the types to be used.


chrisgacsal (changes requested):
Temprary -1 until I managed to review this change.

GAlexIHU:
We most likely wont be doing this.

PR #1834, opened 2024-11-12
feat: invoice line details calculations

turip:
## Overview

Each (`valid`) line can have one or more detailed lines (children). These lines represent the actual sub-charges that are caused by the parent line.

Example:
> If a line has:
> - Usage of 200 units
> - Tiered pricing:
> - Tier1: 1 - 50 units cost flat $300
> - Tier2: 51 - 100 units cost flat $400
> - Tier3: 100 - 150 units cost flat $400 + $1/unit
> - Tier4: more than 150 units cost $15/unit

This would yield the following lines:

- Line with quantity=200
  - Line quantity=1 per_unit_amount=300 total=300 (Tier1)
  - Line quantity=1 per_unit_amount=400 total=400 (Tier2)
  - Line quantity=1 per_unit_amount=400 total=400 (Tier3, flat component)
  - Line quantity=50 per_unit_amount=1 total=50 (Tier3, per unit price)
  - Line quantity=50 per_unit_amount=15 total=759 (Tier4)

Apps can choose to syncronize the original line (if the upstream system understands our pricing model) or can use the sublines to syncronize individual lines without having to understand billing details.

### Detailed Lines vs Splitting

When we are dealing with a split line, the calculation of the quantity is by taking the meter's quantity for the whole line period ([`parent.period.start`, `splitline.period.end`]) and the amount before the period (`parent.period.start`, `splitline.period.start`).

When substracting the two we get the delta for the period (this gets the delta for all supported meter types except of Min and Avg).

We execute the pricing logic (e.g. tiered pricing) for the line qty, while considering the before usage, as it reflects the already billed for items.

Corner cases:
- Graduating tiered prices cannot be billed mid-billing period (always arrears, as the calculation cannot be split into multiple items)
- Min, Avg meters are always billed arrears as we cannot calculate the delta.

### Detailed line persisting

In order for the calculation logic, to not to have to deal with the contents of the database, it is (mostly) the adapter layer's responsibility to understand what have changed and persist only that data to the database.

In practice the high level rules are the following (see [adapter/invoicelinediff_test.go](./adapter/invoicelinediff_test.go) for examples):
- If an entity has an ID then it will be updated
- If an entity has changed compared to the database fetch, it will be updated
- If a child line, discount gets removed, it will be removed from the database (in case of lines with all sub-entities)
- If an entity doesn't have an ID a new entity will be generated by the database

For idempotent entity sources (detailed lines and discounts for now), we have also added a field called `ChildUniqueReferenceID` which can be used to detect entities serving the same purpose.

#### ChildUniqueReferenceID example

Let's say we have an usage-based line whose detailed lines are persisted to the database, but then we would want to change the quantity of the line.

First we load the existing detailed lines from the database, and save the database versions of the entities in memory.

We execute the calculation for the new quantity that yields new detailed lines without database IDs.

The entity's `ChildrenWithIDReuse` call can be used to facilitate the line reuse by assigning the known IDs to the yielded lines where the `ChildUniqueReferenceID` is set.

Then the adapter layer will use those IDs to make decisions if they want to persist or recreate the records.

We could do the same logic in the adapter layer, but this approach makes it more flexible on the calculation layer if we want to generate new lines or not. If this becomes a burden we can do the same matching logic as part of the upsert logic in adapter.


openmeter/billing/entity/errors.go
```diff
@@ -28,6 +28,8 @@ var (
 	ErrInvoiceActionNotAvailable = NewValidationError("invoice_action_not_available", "invoice action not available")
 
 	ErrInvoiceLineFeatureHasNoMeters             = NewValidationError("invoice_line_feature_has_no_meters", "usage based invoice line: feature has no meters")
+	ErrInvoiceLineGraduatedSplitNotSupported     = NewValidationError("invoice_line_graduated_split_not_supported", "graduated tiered pricing is not supported for split periods")
```
hekike, line 31: should be registered in httpdriver

openmeter/billing/service/lineservice/usagebasedline.go
```diff
@@ -135,3 +153,621 @@ func (l usageBasedLine) SnapshotQuantity(ctx context.Context, invoice *billingen
...
+				PaymentTerm:            plan.InArrearsPaymentTerm,
+			},
+		}, nil
+	}
+
+	return nil, nil
+}
+
+func (l usageBasedLine) calculateUnitPriceDetailedLines(usage *featureUsageResponse, unitPrice plan.UnitPrice) (newDetailedLinesInput, error) {
+	out := make(newDetailedLinesInput, 0, 3)
```
hekike, line 222: why three? I see two appends

openmeter/billing/service/lineservice/usagebasedline.go
```diff
@@ -135,3 +153,621 @@ func (l usageBasedLine) SnapshotQuantity(ctx context.Context, invoice *billingen
...
+				Quantity:               alpacadecimal.NewFromInt(1),
+				Amount:                 tier.FlatPrice.Amount,
+				ChildUniqueReferenceID: GraduatedFlatPriceChildUniqueReferenceID,
+				PaymentTerm:            plan.InArrearsPaymentTerm,
+			}
+
+			if price.MaximumAmount != nil {
+				line = line.AddDiscountForOverage(addDiscountInput{
+					BilledAmountBeforeLine: out.Sum(),
+					MaxSpend:               *price.MaximumAmount,
```
hekike, line 299: Should we check if it's not nil?

openmeter/billing/service/lineservice/usagebasedline_test.go
```diff
@@ -0,0 +1,1366 @@
...
+				MaximumAmount: lo.ToPtr(alpacadecimal.NewFromFloat(100)),
+			}),
+			lineMode: midPeriodSplitLineMode,
+			usage: featureUsageResponse{
+				LinePeriodQty: alpacadecimal.NewFromFloat(10),
+			},
+			expect: newDetailedLinesInput{
+				{
+					Name:                   "feature: usage in period",
+					Amount:                 alpacadecimal.NewFromFloat(10),
```
hekike, line 356: shouldn't this be 100?

openmeter/billing/service/lineservice/usagebasedline_test.go
```diff
@@ -0,0 +1,1366 @@
...
+					Quantity:               alpacadecimal.NewFromFloat(10),
+					ChildUniqueReferenceID: UnitPriceUsageChildUniqueReferenceID,
+					PaymentTerm:            plan.InArrearsPaymentTerm,
+				},
+			},
+		})
+	})
+
+	t.Run("usage present, max spend set, but not hit", func(t *testing.T) {
```
hekike, line 365: has the same test name as previous

openmeter/billing/service/lineservice/usagebasedline_test.go
```diff
@@ -0,0 +1,1366 @@
...
+	})
+}
+
+func TestTieredGraduatedCalculation(t *testing.T) {
+	testTiers := []plan.PriceTier{
+		{
+			UpToAmount: lo.ToPtr(alpacadecimal.NewFromFloat(5)),
+			FlatPrice: &plan.PriceTierFlatPrice{
+				// 20/unit
+				Amount: alpacadecimal.NewFromFloat(100),
```
hekike, line 403: Tiered pricing means units prices on amount see:
https://openmeter.io/blog/tiered-pricing-model#example-comparison

So the unit price will be $100 for for the first 5 items not $20

openmeter/billing/service/lineservice/usagebasedline_test.go
```diff
@@ -0,0 +1,1366 @@
...
+
+	t.Run("tiered graduated, last price, usage present, tier4", func(t *testing.T) {
+		runUBPTest(t, ubpCalculationTestCase{
+			price: plan.NewPriceFrom(plan.TieredPrice{
+				Mode:  plan.GraduatedTieredPrice,
+				Tiers: testTiers,
+			}),
+			lineMode: lastInPeriodSplitLineMode,
+			usage: featureUsageResponse{
+				LinePeriodQty: alpacadecimal.NewFromFloat(100),
```
hekike, line 506: I think a test with 14 and 15 qty would be more useful

openmeter/billing/service/lineservice/usagebasedline_test.go
```diff
@@ -0,0 +1,1366 @@
...
+						},
+					},
+				},
+			},
+		})
+	})
+}
+
+func TestTieredVolumeCalculation(t *testing.T) {
+	testTiers := []plan.PriceTier{
```
hekike, line 701: On volume based each unit is billed on the highest achieved discount, see:
https://openmeter.io/blog/tiered-pricing-model#example-comparison

openmeter/billing/service/lineservice/usagebasedline_test.go
```diff
@@ -0,0 +1,1366 @@
...
+		runUBPTest(t, ubpCalculationTestCase{
+			price: plan.NewPriceFrom(plan.TieredPrice{
+				Mode:  plan.VolumeTieredPrice,
+				Tiers: testTiers,
+			}),
+			lineMode: singlePerPeriodLineMode,
+			usage: featureUsageResponse{
+				LinePeriodQty: alpacadecimal.NewFromFloat(22),
+			},
+			expect: newDetailedLinesInput{
```
hekike, line 768: on qty 22 we should have only one line item on the highest discount, see:
https://openmeter.io/blog/tiered-pricing-model#example-comparison

hekike (changes requested):

openmeter/billing/service/lineservice/usagebasedline.go
turip: It's checked on 296.

openmeter/billing/service/lineservice/usagebasedline_test.go
turip: No, there was 10 usage, at 10/unit price. The amount here reflects the unit price. We can rename it perUnitAmount. But on a line level this is always per unit amount.

openmeter/billing/entity/errors.go
turip: All of these are wrapped into ValidationErrors or whatever error it should be reported as, and that is the thing that the HTTP driver handles.

I can only add a fallback, but regardless the previous logic should ensure proper error codes.

openmeter/billing/service/lineservice/usagebasedline_test.go
turip: To match the plan's notion we are using the following terminology:
Tiered/Volume based prices are what the blog names Volume-based pricing
Tiered/Graduating based prices are what the blog names Tiered-based pricing

openmeter/billing/service/lineservice/usagebasedline_test.go
turip: This was a missmatch in naming, fixed.

openmeter/ent/schema/billing.go
```diff
@@ -285,13 +285,24 @@ func (BillingInvoiceLine) Fields() []ent.Field {
...
 			Optional(),
+
+		// child_unique_reference_id is uniqe per parent line, can be used for upserting
+		// and identifying lines created for the same reason (e.g. tiered price tier)
+		// between different invoices.
+		//
+		// As entgo doesn't support conditional unique indexes, defaults to ID of the
+		// line.
+		// TODO: add hooks
+		field.String("child_unique_reference_id").
```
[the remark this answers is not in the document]
turip: Actually with the current setup of child_unique_reference_id being optional, and having a where condition on it not being null for the index this was working fine during the testing.

With the recent changes to the adapter, this is not strictly required OTOH, but it's just a safeguard for db consistency.

hekike (approved):

PR #1838, opened 2024-11-13
fix(logger): http context on logs

hekike:
Use contextx to collect log attributes and enrich logs.
Use semantic convention for common log attributes.


app/common/telemetry.go
```diff
@@ -89,25 +88,33 @@ func NewLoggerProvider(ctx context.Context, conf config.LogTelemetryConfig, res
...
-				otelslog.NewHandler,
-			).Handler(conf.NewHandler(os.Stdout)),
-
-			// Otel logger
-			NewLevelHandler(
-				realotelslog.NewHandler(metadata.OpenTelemetryName, realotelslog.WithLoggerProvider(loggerProvider)),
-				conf.Level,
-			),
-		)),
+	// Stdout logger
```
hekike, line 91: should be the same just refactor

chrisgacsal (approved):

PR #1732, opened 2024-10-21
Subscriptions Internal APIs

GAlexIHU:
## Overview

Includes internal APIs for subscription Management

1. DB Entities
- Subscription + SubscriptionPatch (with value type normalized)
- Price
- SubscriptionEntitlement

2. Interfaces
- Subscription Package Connector with
  - Create
  - Get
  - Expand
  - Edit
  - Cancel (no-op / panics)
- "Patch" Commands for (name TBD) to be used with Create & Edit
  - AddItem
  - RemoveItem
  - AddPhase
  - RemovePhase
  - ExtendPhase
- `SubscriptionXSpec` & `SubscriptionXView` interfaces
  - define create specifications and defacto state for Subscriptions & its subresources

3. Misc
- Testutils for Subscriptions
- Entitlement Annotations <= probably wrong implementation, annotation based Access Control checks for entitlements

## TODO

- [ ] Documentation
- [ ] Self Review

PR #1851, opened 2024-11-14
Subscription Internals V2

GAlexIHU:
## Overview

Supersedes https://github.com/openmeterio/openmeter/pull/1732

Includes internal APIs for subscription Management

1. DB Entities
- Subscription
- SubscriptionPhase
- SubscriptionItem
- SubscriptionEntitlement

2. Interfaces
- Subscription Service (*mostly* package internal API) with
  - Create
  - Get
  - Expand
  - Edit
  - Cancel
  - Continue (a canceled)
- Workflow Service (external API), with
  - CreateFromPlan
  - EditRunning (via patches)
- "Patch" Commands for (name TBD) to be used with Create & Edit
  - AddItem
  - RemoveItem
  - AddPhase
  - RemovePhase
  - StretchPhase
- `SubscriptionXSpec` & `SubscriptionXView` interfaces
  - define create specifications and defacto state for Subscriptions & its subresources

3. Misc
- Testutils for Subscriptions
- ULID feature keys are no longer supported

## Notes for reviewer
- None of `openmeter/subscriptions` is included in application startup.


openmeter/ent/schema/price.go
```diff
@@ -0,0 +1,54 @@
...
+	"entgo.io/ent/dialect"
+	"entgo.io/ent/schema/edge"
+	"entgo.io/ent/schema/field"
+	"entgo.io/ent/schema/index"
+
+	"github.com/openmeterio/openmeter/openmeter/productcatalog/plan"
+	"github.com/openmeterio/openmeter/pkg/framework/entutils"
+)
+
+type Price struct {
```
hekike, line 14: Let's prefix these like `SubscriptionPrice`

openmeter/ent/schema/subscription.go
```diff
@@ -0,0 +1,135 @@
...
+		entutils.TimeMixin{},
+		entutils.MetadataAnnotationsMixin{},
+		entutils.CadencedMixin{},
+	}
+}
+
+func (SubscriptionItem) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("phase_id").NotEmpty().Immutable(),
+		field.String("key").NotEmpty().Immutable(),
```
hekike, line 109: Do we need a key for it, isn't ID enaough? Do we expect users to name subscription items manually?

openmeter/ent/schema/subscription_entitlement.go
```diff
@@ -0,0 +1,47 @@
...
+		entutils.NamespaceMixin{},
+		entutils.TimeMixin{},
+	}
+}
+
+func (SubscriptionEntitlement) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("entitlement_id").NotEmpty().Immutable(),
+		field.String("subscription_id").NotEmpty().Immutable(),
+		field.String("subscription_phase_key").NotEmpty().Immutable(),
```
hekike, line 28: any reason we don't reference `subscription_phase` by ID?

openmeter/ent/schema/subscription_entitlement.go
```diff
@@ -0,0 +1,47 @@
...
+		field.String("subscription_item_key").NotEmpty().Immutable(),
```
hekike, line 29: any reason we don't reference `subscription_item` by ID?

openmeter/ent/schema/subscription.go
```diff
@@ -0,0 +1,135 @@
...
+		// TODO: pending PR
+		field.String("rate_card").
```
hekike, line 111: why isn't this a reference to `plan_rate_card` table?

openmeter/ent/schema/subscription.go
```diff
@@ -0,0 +1,135 @@
...
+type SubscriptionItem struct {
+	ent.Schema
+}
+
+func (SubscriptionItem) Mixin() []ent.Mixin {
+	return []ent.Mixin{
+		entutils.IDMixin{},
+		entutils.NamespaceMixin{},
+		entutils.TimeMixin{},
+		entutils.MetadataAnnotationsMixin{},
```
hekike, line 101: why do we need annotations instead of just adding a  readonly or editable flag like: `field.Bool("readonly").NotEmpty().Immutable().Default(false)` ?

openmeter/entitlement/scheduling.go
```diff
@@ -190,6 +190,8 @@ func (c *entitlementConnector) SupersedeEntitlement(ctx context.Context, entitle
 		}
 
 		// Create new entitlement
+		//
+		// The Unique Constraint durint Scheduling catches the InconsistencyError where the new entitltment would be scheduled active longer then any later entitlement would start.
```
hekike, line 194: typo durint

openmeter/subscription/adapters/entitlement/adapter.go
```diff
@@ -0,0 +1,177 @@
...
+) *EntitlementSubscriptionAdapter {
+	return &EntitlementSubscriptionAdapter{
+		entitlementConnector: entitlementConnector,
+		repo:                 repo,
+		txCreator:            txCreator,
+	}
+}
+
+func (a *EntitlementSubscriptionAdapter) ScheduleEntitlement(ctx context.Context, input subscription.ScheduleSubscriptionEntitlementInput) (*subscription.SubscriptionEntitlement, error) {
+	if input.EntitlementInputs.ActiveFrom == nil {
```
hekike, line 40: how about `if err := input.Validation(); err != nil {`

openmeter/ent/schema/subscription.go
GAlexIHU: To avoid linking I'll copy from relevant docs

SubscriptionItem can be identified, it has a reference:
- If a Feature is associated with the SubscriptionItem, it is identified by the Feature
  - It can be an ID reference, for an exact version of the Feature (Features can change across versions)
  - It can be a Key reference, which always refers to the latest (active or inactive) version of a Feature
- If a Feature is not associated with the SubscriptionItem, it is referenced by the Price
  - We say “referenced by the Price” regardless of how a price itself is referenced, it colloquially makes sense to say “paying the same price for the same thing”
In practice this will need an ID _(key)_ being generated

Keys are used for identifying the "same thing" across different instances (with different IDs)

openmeter/ent/schema/subscription.go
GAlexIHU: It's not a direct reference to an actual plan's actual ratecard, it will only hold the same contents as a ratecard does.

openmeter/subscription/adapters/entitlement/adapter.go
GAlexIHU: good catch

openmeter/ent/schema/price.go
GAlexIHU: deleted it

openmeter/ent/schema/subscription_entitlement.go
GAlexIHU: there were, but currently there isnt

openmeter/ent/schema/subscription_entitlement.go
GAlexIHU: there were, but currently there isn't

openmeter/ent/schema/subscription.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+}
+
+func (SubscriptionItem) Fields() []ent.Field {
+	return []ent.Field{
+		// Note: we cannot use CadencedMixin as these are mutable
+		field.Time("active_from"),
+		field.Time("active_to").Optional().Nillable(),
+		field.String("phase_id").NotEmpty().Immutable(),
+		field.String("start_after_phase").
```
hekike, line 111: What is `start_after_phase` for?

openmeter/entitlement/adapter/entitlement.go
```diff
@@ -412,14 +413,18 @@ func mapEntitlementEntity(e *db.Entitlement) *entitlement.Entitlement {
 				UpdatedAt: e.UpdatedAt.UTC(),
 				DeletedAt: convert.SafeToUTC(e.DeletedAt),
 			},
+			AnnotatedModel: models.AnnotatedModel{
```
hekike, line 416: does it still need to be Annotated?

openmeter/entitlement/entitlement.go
```diff
@@ -92,6 +97,100 @@ type CreateEntitlementInputs struct {
...
+func (c CreateEntitlementInputs) Equal(other CreateEntitlementInputs) bool {
+	if c.Namespace != other.Namespace {
+		return false
+	}
+
+	if !reflect.DeepEqual(c.FeatureID, other.FeatureID) {
+		return false
+	}
+
+	if !reflect.DeepEqual(c.FeatureKey, other.FeatureKey) {
```
hekike, line 113: why do we need deep equal for a string?

GAlexIHU: deepEqual is not ideal, its used because they're `*string` not `string`

openmeter/entitlement/adapter/entitlement.go
GAlexIHU: I think the direction we agreed in general is that everything's supposed to have metadata, so yes?

openmeter/ent/schema/subscription.go
GAlexIHU: stores relative start time compared to phase start time if it differs. It's needed for more complex scenarios like cancellation & continuing

openmeter/ent/schema/subscription.go
```diff
@@ -0,0 +1,141 @@
...
+		entutils.NamespaceMixin{},
+		entutils.TimeMixin{},
+		entutils.MetadataAnnotationsMixin{},
+	}
+}
+
+func (SubscriptionPhase) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("subscription_id").NotEmpty().Immutable(),
+		field.String("key").NotEmpty().Immutable(),
```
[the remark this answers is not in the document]
GAlexIHU: no, it should be a stable locally unique reference, which an index is not

openmeter/ent/schema/subscription.go
```diff
@@ -0,0 +1,141 @@
...
+		entutils.NamespaceMixin{},
+		entutils.TimeMixin{},
+		entutils.MetadataAnnotationsMixin{},
+		entutils.CadencedMixin{},
+	}
+}
+
+func (Subscription) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("plan_key").NotEmpty().Immutable(),
```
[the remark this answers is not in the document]
GAlexIHU: Yes, that will become optional in a future PR

openmeter/ent/schema/subscription.go
```diff
@@ -0,0 +1,141 @@
...
+		field.Int("plan_version").Min(1).Immutable(),
```
hekike, line 33: Are we using `plan_key` + `plan_version` instead of `plan_id` to know for version migration which plan it is?

openmeter/ent/schema/subscription.go
```diff
@@ -0,0 +1,141 @@
...
+		field.String("customer_id").NotEmpty().Immutable(),
+		field.String("currency").GoType(currencyx.Code("")).MinLen(3).MaxLen(3).NotEmpty().Immutable(),
+	}
+}
+
+func (Subscription) Indexes() []ent.Index {
+	return []ent.Index{
+		index.Fields("namespace", "id"),
+		index.Fields("namespace", "customer_id"),
```
hekike, line 42: in the future probably we need index on plan too so we can list subs for a given plan version

openmeter/ent/schema/subscription.go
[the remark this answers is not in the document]
hekike: I'm not sure I agree with the statement, as you can put a unique index on subscription_id + key both if it's string or integer.

My question is more if this should be a string if it's going to be number in product catalog phase. But I also see that change is not merged yet: https://github.com/openmeterio/openmeter/blob/main/openmeter/productcatalog/plan/phase.go#L19 cc @chrisgacsal  Maybe worth discussing with @chrisgacsal and @tothandras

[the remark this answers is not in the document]
hekike: Or this is not a reference to plan_phase? If not do we need one?

openmeter/ent/schema/subscription.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+}
+
+func (SubscriptionPhase) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("subscription_id").NotEmpty().Immutable(),
+		field.String("key").NotEmpty().Immutable(),
+		field.String("name").NotEmpty(),
+		field.String("description").Optional().Nillable(),
+		field.Time("active_from").Immutable(),
```
hekike, line 73: don't we need an optional active_to? Or that coming from rate cards somehow? If yes why do we need active_from?

openmeter/ent/schema/subscription.go
```diff
@@ -0,0 +1,141 @@
...
+		index.Fields("namespace", "subscription_id"),
+		index.Fields("namespace", "subscription_id", "key"),
+	}
+}
+
+func (SubscriptionPhase) Edges() []ent.Edge {
+	return []ent.Edge{
+		edge.From("subscription", Subscription.Type).Field("subscription_id").Ref("phases").Unique().Immutable().Required(),
+		edge.To("items", SubscriptionItem.Type),
+	}
```
hekike, line 89: we should add a unqiue index for subscription_id + phase key

openmeter/ent/schema/subscription.go
```diff
@@ -0,0 +1,141 @@
...
+func (SubscriptionItem) Indexes() []ent.Index {
+	return []ent.Index{
+		index.Fields("namespace", "id"),
+		index.Fields("namespace", "phase_id", "key"),
+	}
+}
+
+func (SubscriptionItem) Edges() []ent.Edge {
+	return []ent.Edge{
+		edge.From("phase", SubscriptionPhase.Type).Field("phase_id").Ref("items").Unique().Immutable().Required(),
```
hekike, line 136: should we call field `subscription_phase_id`?

openmeter/ent/schema/subscription.go
```diff
@@ -0,0 +1,141 @@
...
+		edge.To("entitlements", SubscriptionEntitlement.Type),
```
hekike, line 137: should we define db field  to be `entitlement_id`?

openmeter/ent/schema/subscription.go
```diff
@@ -0,0 +1,141 @@
...
+		field.Time("active_from"),
+		field.Time("active_to").Optional().Nillable(),
+		field.String("phase_id").NotEmpty().Immutable(),
+		field.String("start_after_phase").
+			GoType(datex.ISOString("")).Nillable().Optional(),
+		field.String("end_after_phase").
+			GoType(datex.ISOString("")).Nillable().Optional(),
+		field.String("key").NotEmpty().Immutable(),
+		field.String("rate_card").
+			GoType(&subscription.RateCard{}).
```
hekike, line 117: any reason it's not materialized? In product catalog rate card is materialized:
https://github.com/openmeterio/openmeter/blob/main/openmeter/ent/schema/productcatalog.go#L119

openmeter/ent/schema/subscription.go
GAlexIHU: Just that it conveys more visible meaning than an opaque id.

openmeter/ent/schema/subscription.go
GAlexIHU: By the current spec, a subscription cannot have a state where there isn't an active phase, unless the subscription itself is inactive (cancelled etc...) it's just a simple way of guaranteeing it

openmeter/ent/schema/subscription.go
GAlexIHU: We can add a unique index for subscription_id + phase_key + deletedAt (taking into account that deletedAt is nillable)

openmeter/ent/schema/subscription.go
GAlexIHU: Right now mainly for simplicity and time saving. ProductCatalog models (like RateCard) should be made available for use outside of those packages which is currently not possible, the `subscription.RateCard` should disappear once that is resolved. I'd be much happier to reuse the exact code used in Plan for much of this, but you're right that's not a requirement, I'm happy to normalize it if you'd like, it's just a bit of double work, but there's already quite a bit of double work happening so that doesn't have to be an issue.

hekike (approved):

PR #1869, opened 2024-11-18
fix: updating PlanPhase to have no RateCards

chrisgacsal:
## Overview

Fix issue where updating a Plan Phase to have no RateCards is not properly applied.

### DIFF
--- a/openmeter/productcatalog/plan/adapter/phase.go
+++ b/openmeter/productcatalog/plan/adapter/phase.go
@@ -386,7 +386,7 @@ func (a *adapter) UpdatePhase(ctx context.Context, params plan.UpdatePhaseInput)
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


GAlexIHU (commented):
AFAIK it should still be disallowed when publishing, is that handled separately?

chrisgacsal:
@GAlexIHU It is handled for `Plan` but not for `PlanPhase` which is a bug and I already have a fix for that, just wanted to push these chagnes separately. So  you are right, that `Plans` `Status` must be checked when `PlanPhase` is udpated/deleted and only allow for draft `Plans`.

GAlexIHU (approved):
I'm happy to approve to get things going, please request my review on the next PR too (one with the Plan API fix)

PR #1888, opened 2024-11-19
fix(customer): delete customer subjects

hekike:
Delete customer subject relations so another customer can reuse the subject.

### DIFF
--- a/openmeter/ent/schema/customer.go
+++ b/openmeter/ent/schema/customer.go
@@ -85,14 +85,23 @@ func (CustomerSubjects) Fields() []ent.Field {
 		field.Time("created_at").
 			Default(clock.Now).
 			Immutable(),
+		field.Time("deleted_at").
+			Optional().
+			Nillable(),
+		// We use boolean for soft delete instead of time.Time
+		// because we can only add unique indexes on fields that are not nullable.
+		field.Bool("is_deleted").Default(false),
 	}
 }
 
 func (CustomerSubjects) Indexes() []ent.Index {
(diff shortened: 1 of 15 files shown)


solkimicreb (approved):

openmeter/ent/schema/customer.go
```diff
@@ -85,14 +85,23 @@ func (CustomerSubjects) Fields() []ent.Field {
...
 	return []ent.Index{
-		index.Fields("customer_id", "subject_key").
-			Unique(),
-		index.Fields("namespace", "subject_key").
+		index.Fields("namespace", "customer_id", "is_deleted"),
+		index.Fields("namespace", "subject_key", "is_deleted").
+			Annotations(
+				// Partial index: We skip the index on active non deleted customer subjects.
+				entsql.IndexWhere("is_deleted = false"),
+			).
```
tothandras, line 104: Can we avoid somehow the use of `is_deleted`? Like:
```suggestion
		index.Fields("namespace", "customer_id"),
		index.Fields("namespace", "subject_key", "is_deleted").
			Annotations(
				// Partial index: We skip the index on active non deleted customer subjects.
				entsql.IndexWhere("deleted_at IS NULL"),
			).
```

hekike: This doesn't work. You need Postgres 17 and using `NULLS NOT DISTINCT `.
https://www.postgresql.org/docs/current/indexes-unique.html#INDEXES-UNIQUE

PR #1889, opened 2024-11-19
fix: listing deleted PlanPhases

chrisgacsal:
## Overview

* fix returning deleted PlanPhases which is not desired
* fix unique index so repalcing PlanPhase does not result in SQL contraint violation as long as the `deleted_at` field is unique

### DIFF
--- a/tools/migrate/migrations/20241119154436_plan.up.sql
+++ b/tools/migrate/migrations/20241119154436_plan.up.sql
@@ -1,0 +1,5 @@
+-- drop index "planphase_plan_id_key" from table: "plan_phases"
+DROP INDEX "planphase_plan_id_key";
+-- create index "planphase_plan_id_key_deleted_at" to table: "plan_phases"
+-- atlas:nolint
+CREATE UNIQUE INDEX "planphase_plan_id_key_deleted_at" ON "plan_phases" ("plan_id", "key", "deleted_at");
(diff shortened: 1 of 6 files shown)


tools/migrate/migrations/20241119154436_plan.up.sql
```diff
@@ -0,0 +1,5 @@
+-- drop index "planphase_plan_id_key" from table: "plan_phases"
+DROP INDEX "planphase_plan_id_key";
+-- create index "planphase_plan_id_key_deleted_at" to table: "plan_phases"
+-- atlas:nolint
+CREATE UNIQUE INDEX "planphase_plan_id_key_deleted_at" ON "plan_phases" ("plan_id", "key", "deleted_at");
```
GAlexIHU, line 5: We discussed offline that this allows multiple plans with same id & key as long as they all have deletedAt NULL

GAlexIHU (dismissed):

tools/migrate/migrations/20241119154436_plan.up.sql
chrisgacsal: Yup. We need to upgrade to Postgres 15+ which supports `NULLS NOT DISTINCT` which is the proper fix for this issue.

GAlexIHU (approved):

tothandras (approved):

PR #1905, opened 2024-11-22
feat: line updating

turip:
## Overview

This patch adds support for updating lines. This paves the way for line deletion.

Mayjor changes:
- Calculator can recalculate the whole invoice
- WithTransactingRepo is moved to adapter level so that we have less dependencies on service level
- All changes advance the invoice as much as possible
- InvoiceLine update operation becomes PUT as we cannot properly support PATCH on union types for now
- Add entity level support for ValidationError
- Add support for persisting the prePeriodQTY for Usage based prices
- Add support for validating updates in terms of billable period


openmeter/billing/invoiceline.go
```diff
@@ -101,28 +106,220 @@ func (i AssociateLinesToInvoiceAdapterInput) Validate() error {
...
+		outErr = errors.Join(outErr, billingentity.ValidationWithFieldPrefix("id", billingentity.ErrFieldRequired))
 	}
 
-	if g.InvoiceID == "" {
-		return errors.New("invoice id is required")
+	if u.Invoice.Namespace == "" {
+		outErr = errors.Join(outErr, billingentity.ValidationWithFieldPrefix("namespace", billingentity.ErrFieldRequired))
+	}
+
+	if u.UpdatedAt.IsZero() {
```
hekike, line 132: isn't updated at set at create?

turip: This is for conflict detection. I will remove this conflict detection in the next PR or make it optional backend logic only.

Right now each update mandates that you specify the current db version of UpdatedAt, so that we are not overwriting anything with stale data. This helps a lot in tests, but can be a bit daunting for API usage, so discussing with @tothandras I will drop this as part of the API refactor.

hekike (approved):

PR #1931, opened 2024-12-07
feat: invoicing app support

turip:
## Overview

This patch adds support for invoicing apps, allowing to integrate with external parties such as Stripe.

The invoicing app is required to implement this interface or the invoicing flow will fail:
```

type InvoicingApp interface {
	// ValidateInvoice validates if the app can run for the given invoice
	ValidateInvoice(ctx context.Context, invoice Invoice) error

	// UpsertInvoice upserts the invoice on the remote system, the invoice is read-only, the app should not modify it
	// the recommended behavior is that the invoices FlattenLinesByID is used to get all lines, then the app should
	// syncronize all the fee lines and store the external IDs in the result.
	UpsertInvoice(ctx context.Context, invoice Invoice) (*UpsertInvoiceResult, error)

	// FinalizeInvoice finalizes the invoice on the remote system, starts the payment flow. It is safe to assume
	// that the state machine have already performed an upsert as part of this state transition.
	//
	// If the payment is handled by a decoupled implementation (different app or app has strict separation of concerns)
	// then the payment app will be called with FinalizePayment and that should return the external ID of the payment. (later)
	FinalizeInvoice(ctx context.Context, invoice Invoice) (*FinalizeInvoiceResult, error)

	// DeleteInvoice deletes the invoice on the remote system, the invoice is read-only, the app should not modify it
	// the invoice deletion is only invoked for non-finalized invoices.
	DeleteInvoice(ctx context.Context, invoice Invoice) error
}

```

### DIFF
--- a/openmeter/ent/schema/billing.go
+++ b/openmeter/ent/schema/billing.go
@@ -325,6 +325,11 @@ func (BillingInvoiceLine) Fields() []ent.Field {
 			}).
 			Optional(),
 
+		// App IDs
+		field.String("invoicing_app_external_id").
+			Optional().
+			Nillable(),
+
 		// child_unique_reference_id is uniqe per parent line, can be used for upserting
 		// and identifying lines created for the same reason (e.g. tiered price tier)
 		// between different invoices.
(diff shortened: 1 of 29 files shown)


openmeter/ent/schema/billing.go
```diff
@@ -597,6 +602,15 @@ func (BillingInvoice) Fields() []ent.Field {
 				dialect.Postgres: "char(26)",
 			}),
 
+		// App IDs
```
hekike, line 605: don't we need a `tax_app_external_id` too?

hekike (approved):

openmeter/ent/schema/billing.go
turip: Eventually we will but tax is a heavier lift, as we need to implement calculations, tax db tables and recalculation avoidance, so that's better kept seperate

PR #1959, opened 2024-12-13
feat: billing add subscription metadata to lines

turip:
## Overview

Add subscription metadata for lines.

We only support setting it via internal APIs, externally the user cannot update/create lines with subscription references.


api/spec/src/billing/invoices/invoice.tsp
```diff
@@ -580,6 +580,12 @@ model InvoiceLineBase {
    */
   @visibility("read")
   externalIDs?: InvoiceLineAppExternalIDs;
+
+  /**
+   * Subscription are the references to the subscritpions that this line is related to.
+   */
+  @visibility("read")
+  subscriptions?: InvoiceLineSubscriptionReference;
```
hekike, line 588: Should this be singular? Looks like reference is for one.

api/spec/src/types.tsp
```diff
@@ -78,6 +78,22 @@ model UniqueResource {
   key: Key;
 }
 
+/**
+ * IDResource is a resouce with an ID.
+ */
+// NOTE: this can be used to have a type, that we can later replace with the expanded type if needed without
+// breaking api compatibility
+@friendlyName("IDResource")
+model IDResource {
```
hekike, line 87: What's the value of doing IDResource? Is the resource type changing for the fields where it's used?

hekike: This btw reminds me of GraphQL's global IDs where every resource's ID is globally unique. Usually it's a concatenation of the type and ID .

hekike (approved):

api/spec/src/types.tsp
turip: This is just an API compatibility issue. This allows the response to be like this:

```
{
   "subscription": {
      "subscription": {
         "id": ...
      },
      "phase": {
          "id": ...
      }
...
}
```

This way if we want to extend the API with extending the objects later, we can do that without needing to break the API.

PR #1974, opened 2024-12-16
Subscription Change

GAlexIHU:
## Overview

Slightly restructures subscription package

Implements
- Plan Change (from one key to another)
- Version Migration (from one version to another)

Changes the API signature for the above two as it makes more sense to return both the current and the new version of the subscriptions


api/spec/src/productcatalog/subscription.tsp
```diff
@@ -233,6 +233,18 @@ model SubscriptionEdit {
   customizations: SubscriptionEditOperation[];
 }
 
+/**
+ * Response body for subscription change.
+ */
+@friendlyName("SubscriptionChangeResponseBody")
+model SubscriptionChangeResponseBody {
+  @summary("Current subscription")
+  current: Subscription;
```
hekike, line 242: Should we call this `previous` and `next` instead?
Current, to me, is ambiguous as we don't know when the previous ends.

cmd/server/wire.go
```diff
@@ -37,31 +37,31 @@ type Application struct {
...
+	MeterRepository           meter.Repository
+	NamespaceHandlers         []namespace.Handler
+	NamespaceManager          *namespace.Manager
+	Notification              notification.Service
+	Meter                     metric.Meter
+	Plan                      plan.Service
+	RouterHook                func(chi.Router)
+	Secret                    secret.Service
+	Subscription              common.SubscriptionServiceWithWorkflow
+	SubscriptionChangeService plansubscription.PlanSubscriptionService
```
hekike, line 62: Can we initialize this in the Subscription service? 
I don't think we need this granularity to be able to run one of them independently.

e2e/productcatalog_test.go
```diff
@@ -370,4 +370,91 @@ func TestPlan(t *testing.T) {
...
+		require.NotNil(t, planAPIRes.JSON201)
+		require.NotNil(t, planAPIRes.JSON201.Id)
+		require.NotNil(t, planAPIRes.JSON201.Version)
+		require.NotNil(t, planAPIRes.JSON201.Key)
+
+		assert.NotEqual(t, planId, *planAPIRes.JSON201.Id)
+		assert.Equal(t, PlanKey, planAPIRes.JSON201.Key)
+		assert.Equal(t, 2, *planAPIRes.JSON201.Version)
+
+		// Let's publis the new version
```
hekike, line 402: /s/publish/publish

e2e/productcatalog_test.go
```diff
@@ -370,4 +370,91 @@ func TestPlan(t *testing.T) {
...
+	})
+
+	var migratedSubscriptionId string
+	var migratedSubView api.SubscriptionExpanded
+
+	t.Run("Should migrate the subscription to a newer version", func(t *testing.T) {
+		require.NotNil(t, subscriptionId)
+
+		apiRes, err := client.MigrateSubscriptionWithResponse(ctx, subscriptionId, api.MigrateSubscriptionJSONRequestBody{
+			TargetVersion: 2,
```
hekike, line 416: HTTP API should default to the latest version.

openmeter/productcatalog/subscription/http/change.go
```diff
@@ -0,0 +1,133 @@
...
+
+			// Any transformation function generated by the API will succeed if the body is serializable, so we have to check for the presence of
+			// fields to determine what body type we're dealing with
+			type testForCustomPlan struct {
+				CustomPlan any `json:"customPlan"`
+			}
+
+			var t testForCustomPlan
+
+			bodyBytes, err := json.Marshal(body)
```
hekike, line 50: why do we need to double parse body?

openmeter/productcatalog/subscription/http/migrate.go
```diff
@@ -0,0 +1,65 @@
...
+			if err != nil {
+				return MigrateSubscriptionRequest{}, err
+			}
+
+			return MigrateSubscriptionRequest{
+				ID: models.NamespacedID{
+					Namespace: ns,
+					ID:        params.ID,
+				},
+				TargetVersion: body.TargetVersion,
```
hekike, line 42: should accept undefined and then use latest version

openmeter/productcatalog/subscription/service/change.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (s *service) Change(ctx context.Context, request plansubscription.ChangeSubscriptionRequest) (plansubscription.SubscriptionChangeResponse, error) {
+	var def plansubscription.SubscriptionChangeResponse
+	var plan subscription.Plan
+
+	if err := request.PlanInput.Validate(); err != nil {
+		return def, &models.GenericUserError{Message: err.Error()}
+	}
+
+	// TODO: let's validate the plan can be changed to!
```
hekike, line 21: pls create ticket and add as `// TODO (OM-XX)`

openmeter/productcatalog/subscription/service/change_test.go
```diff
@@ -0,0 +1,258 @@
...
+				ID: sub.NamespacedID,
+				WorkflowInput: subscription.ChangeSubscriptionWorkflowInput{
+					ActiveFrom: clock.Now(),
+					Name:       sub.Name,
+				},
+				PlanInput: pInp,
+			})
+
+			require.NotNil(t, err)
+			require.ErrorAs(t, err, lo.ToPtr(&models.GenericUserError{}))
```
hekike, line 255: Should we use typed error? Otherwise this will return 500

openmeter/subscription/testutils/compare.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/openmeterio/openmeter/openmeter/entitlement"
+	"github.com/openmeterio/openmeter/openmeter/subscription"
+	"github.com/openmeterio/openmeter/pkg/recurrence"
+)
+
+// Ensures the created view matches the input spec
+// FIXME: there are quire a few things not asserted here, let's be methodical and add them
```
hekike, line 16: reference ticket pls + list out the few things

api/spec/src/productcatalog/subscription.tsp
GAlexIHU: I suppose thats fair, previous is weird for me as if the change is scheduled later previous would point to a still active one that might deactivate in days or later (thats why i chose current). New vs Next should definitely be Next though (IMO)

openmeter/productcatalog/subscription/service/change_test.go
GAlexIHU: what do you mean? this is a typed error

openmeter/productcatalog/subscription/http/change.go
GAlexIHU: To be able to determine which request body we received. It's commented just above, but the body.AsXY functions will always succeed if the body is deserializable (there isn't a discriminator for this type that we could check for). We can do it without the extra parsing, it was just less awkward to do it like this, as the values body.AsXY returns cannot easily be checked for being empty (they aren't pointers but structs)

openmeter/productcatalog/subscription/service/change.go
GAlexIHU: thanks, was left there by mistake, removed it

openmeter/productcatalog/subscription/http/migrate.go
GAlexIHU: updated

hekike (dismissed):

turip (approved):

PR #1978, opened 2024-12-17
feat(invoice): stripe app invoice sync

hekike:
Synchronize OpenMeter invoices with Stripe Invoicing through apps.

## How it works

The invoice state machine calls the interfaces of compatible apps like Stripe App to Upsert and Finalize the invoice.
Driven by the state machine, the Stripe App creates, updates, and finalizes a Stripe Invoice object via the Stripe SDK.

If the invoice doesn't exist yet (determined by checking its external invoicing app ID), it is created; otherwise, it is updated in Stripe. Every OpenMeter invoice line and discount is translated to flat InvoiceLineItem(s) in Stripe. Discounts are represented as negative amounts. The update diffs the line items between Stripe and OpenMeter and decides which lines need to be updated, added, or removed in Stripe.

We prioritize synchronizing lines with quantity and per-unit price for greater visibility when they fit Stripe amount cents. When any Stripe line doesn't fit cents without floating points, we synchronize all lines with the total amount (the product of quantity and unit price).

Currently, automatic tax collection is enabled.

## Todos:

- [x] Add discount external id (https://github.com/openmeterio/openmeter/pull/1989)
- [x] Use external IDs to map Stripe line item to OM Line item instead of metadata
- [x] Minimum spend Total amount is zero


openmeter/app/stripe/entity/app/app_invoice.go
```diff
@@ -0,0 +1,182 @@
...
+
+	invoice.Lines.ForEach(func(lines []*billing.Line) {
+		for _, line := range lines {
+			switch line.Type {
+			case billing.InvoiceLineTypeFee:
+				lineParams.Lines = append(lineParams.Lines, &stripe.InvoiceAddLinesLineParams{
+					InvoiceItem: lo.ToPtr(line.Name),
+					Description: line.Description,
+					Amount:      lo.ToPtr(line.FlatFee.PerUnitAmount.GetFixed()),
+					Quantity:    lo.ToPtr(line.FlatFee.Quantity.GetFixed()),
```
turip, line 132: Discounts should be added too (it's working on a line level)

openmeter/app/stripe/entity/app/app_invoice.go
```diff
@@ -0,0 +1,426 @@
...
+
+				result.AddLineExternalID(line.ID, line.ExternalIDs.Invoicing)
+
+				updateStripeLines.Lines = append(updateStripeLines.Lines, &stripe.InvoiceUpdateLinesLineParams{
+					ID:          lo.ToPtr(line.ExternalIDs.Invoicing),
+					Description: line.Description,
+					Amount:      lo.ToPtr(line.Totals.Amount.GetFixed()),
+					Quantity:    lo.ToPtr(line.UsageBased.Quantity.GetFixed()),
+					// period is not updatable
+				})
```
turip, line 351: If you are syncronizing the UsageBasedLine and it's children the total will be doubled. Either synchronize the children only or the UBP line's totals but both both.

openmeter/app/stripe/entity/app/app_invoice.go
```diff
@@ -0,0 +1,426 @@
...
+		})
+
+		// Add line item
+		switch line.Type {
+		case billing.InvoiceLineTypeFee:
+			lineIDs = append(lineIDs, line.ID)
+
+			stripeInvoiceLineParams.Lines = append(stripeInvoiceLineParams.Lines, &stripe.InvoiceAddLinesLineParams{
+				Description: line.Description,
+				Amount:      lo.ToPtr(line.Totals.Amount.GetFixed()),
```
turip, line 174: This is incorrect, due to number representation.

GetFixed gets the fixed point representation of the number if available (I think it will use ~10 binary digits).

This only exists for optimized serialization of the alpacadecimal package.

The stripe API states that:
```
	// The integer amount in cents (or local equivalent) of the charge to be applied to the upcoming invoice. If you want to apply a credit to the customer's account, pass a negative amount.
	Amount *int64 `form:"amount"`
```

So here you should use the currencyx/gobl currency package to determine the subunits of the given currency and get an integer representation rounded to the base 10 decimal places.

```

File: openmeter/app/stripe/client/invoice.go [hunk 4]
```diff
@@ -0,0 +1,175 @@
...
+		Customer: lo.ToPtr(input.StripeCustomerID),
+		// FinalizeInvoice will advance the invoice
+		AutoAdvance: lo.ToPtr(false),
+		// When charging automatically, Stripe will attempt to pay this invoice using the default source attached to the customer.
+		// When sending an invoice, Stripe will email this invoice to the customer with payment instructions. Defaults to charge_automatically.
+		CollectionMethod: lo.ToPtr(string(stripe.InvoiceCollectionMethodChargeAutomatically)),
+		// If not set, defaults to the default payment method in the customer’s invoice settings.
+		DefaultPaymentMethod: input.StripeDefaultPaymentMethodID,
+		StatementDescriptor:  lo.ToPtr(input.StatementDescriptor),
+	}
```
tothandras, line 32:
```suggestion
		StatementDescriptor:  lo.ToPtr(input.StatementDescriptor),
		Number:               input.Number,
	}
```

openmeter/app/stripe/client/invoice.go
```diff
@@ -0,0 +1,175 @@
...
+
+	if input.DueDate != nil {
+		params.DueDate = lo.ToPtr(input.DueDate.Unix())
+	}
+
+	if input.Number != nil {
+		params.Number = input.Number
+	}
```
tothandras, line 40:
```suggestion
```

openmeter/app/stripe/client/invoice.go
```diff
@@ -0,0 +1,175 @@
...
+
+	params := &stripe.InvoiceParams{
+		StatementDescriptor: lo.ToPtr(input.StatementDescriptor),
+	}
+
+	if input.DueDate != nil {
+		params.DueDate = lo.ToPtr(input.DueDate.Unix())
+	}
+
+	if input.Number != nil {
```
tothandras, line 59: same here, can be directly included in params

openmeter/app/stripe/client/invoice.go
```diff
@@ -0,0 +1,175 @@
...
+	if i.Currency == "" {
+		return errors.New("currency is required")
+	}
+
+	if i.DueDate != nil && i.DueDate.IsZero() {
+		return errors.New("due date cannot be zero")
+	}
+
+	if i.Number != nil && *i.Number == "" {
+		return errors.New("invoice number cannot be empty")
```
tothandras, line 112: validate that it's a number instead?

openmeter/app/stripe/client/invoice.go
```diff
@@ -0,0 +1,175 @@
...
+	if i.StripeInvoiceID == "" {
+		return errors.New("stripe invoice id is required")
+	}
+
+	if i.DueDate != nil && i.DueDate.IsZero() {
+		return errors.New("due date cannot be zero")
+	}
+
+	if i.Number != nil && *i.Number == "" {
+		return errors.New("invoice number cannot be empty")
```
tothandras, line 140: same

openmeter/app/stripe/entity/app/customer.go
```diff
@@ -0,0 +1,171 @@
...
+						CustomerID: customerID,
+						Condition:  fmt.Sprintf("default payment method %s not found in stripe account %s", *stripeCustomerData.StripeDefaultPaymentMethodID, stripeAppData.StripeAccountID),
+					}
+				}
+
+				return fmt.Errorf("failed to get default payment method: %w", err)
+			}
+		} else {
+			// Check if the customer has a default payment method
+			if stripeCustomer.DefaultPaymentMethod != nil {
```
tothandras, line 84: optional: check error case first to avoid nesting

tothandras (changes requested):

openmeter/app/stripe/client/invoice.go
hekike: It's not a number. It's the invoice sequence like INV-001.

tothandras (approved):

PR #1996, opened 2024-12-28
feat: use plan phase duration and index instead of start after

tothandras:
- update API and DB to use duration + index instead of startAfter for plan phases
- remove unused next plan operation
- remove phase level API endpoints
- remove phase level internal operations

TODO:
- [ ] remove commented out code
- [ ] fix tests


openmeter/ent/schema/productcatalog.go
```diff
@@ -83,6 +83,8 @@ func (PlanPhase) Fields() []ent.Field {
 		field.String("plan_id").
 			NotEmpty().
 			Comment("The plan identifier the phase is assigned to."),
+		field.Int("index").
```
hekike, line 86: `uint8`

openmeter/productcatalog/plan.go
```diff
@@ -68,58 +76,70 @@ func (p Plan) Validate() error {
...
+	return true
 }
 
-var _ models.Validator = (*PlanMeta)(nil)
+var (
+	_ models.Validator         = (*PlanMeta)(nil)
+	_ models.Equaler[PlanMeta] = (*PlanMeta)(nil)
+)
 
 type PlanMeta struct {
```
hekike, line 103: Why is this called PlanMeta? Is this the internal model of Plan?

openmeter/productcatalog/plan/service.go
```diff
@@ -328,242 +322,249 @@ func (i ArchivePlanInput) Validate() error {
 	return nil
 }
 
-type NextPlanInput struct {
-	// NamespacedID
-	models.NamespacedID
+// type NextPlanInput struct {
```
hekike, line 325: why commented out?

hekike (changes requested):

tothandras:
I'm splitting this PR into multiple commits

openmeter/productcatalog/plan/service.go
chrisgacsal: The functionality provided by `NextPlan` API is also available via `CreatePlan` API. The only difference is that the former allows creating a new version with optinal plan payload while the latter requires a plan to be sent as part of the request. But I tend to agree that it does not worth to keep/maintain 2 API endpoints for the same operation.

PR #2010, opened 2025-01-04
feat: use plan phase duration and index instead of start after

tothandras:
Update API and DB to use duration + index instead of startAfter for plan phases.


api/spec/src/productcatalog/plan.tsp
```diff
@@ -147,20 +147,20 @@ model PlanPhase {
...
   /**
-   * The rate cards of the plan.
+   * The duration of the phase.
    */
   @visibility("read", "create", "update")
-  @summary("Rate cards")
-  rateCards: RateCard[];
+  @summary("Duration")
+  @encode(DurationKnownEncoding.ISO8601)
+  @example(duration.fromISO("P1Y1D"))
```
GAlexIHU, line 155: nit: a realistic example could be better

tools/migrate/migrations/20250103121359_plan-phase-duration.down.sql
```diff
@@ -0,0 +1,8 @@
+-- delete all "plan_phases", phases are incompatible with the old plan phase schema
+-- atlas:nolint destructive
+DELETE FROM "plan_phases";
```
GAlexIHU, line 3: I think there are (at least) 4 options here:
1. Delete everything from productcatalog so nothing is inconsistent
2. Delete just phases + dependent tables (what you did here)
3. Leave in the previous phases in a broken state
4. Write a routine that pessimistically maps start_after to duration

Why did you choose n.o.2?

pkg/datex/date.go
```diff
@@ -125,6 +125,19 @@ func (d *Period) ISOStringPtrOrNil() *ISOString {
 	return lo.ToPtr(d.ISOString())
 }
 
+// Equal returns true if the two periods are equal
+func (p *Period) Equal(v *Period) bool {
```
GAlexIHU, line 129: IMO this is a bit misleading, equal here means their component values are equal, not that their period / length is equal. Do you think normalization / simplification should be part of the comparison or not? (IMO probably not, just thinking)

chrisgacsal (approved):

GAlexIHU (approved):

PR #2022, opened 2025-01-06
Attribute based errors

GAlexIHU:
## Overview

Introduces a simple Trait system for better error handling. Errors can have arbitrary number of traits, you can assign traits and check for the presence of a trait. Two example cases where this was used are:
- Allowing some errors during incremental validation but not final validation (`subscriptionspec.go`)
- HTTP status mapping with traits

If you're approving of the approach i'll invest more time, e.g. migrate & dedupe the current error mappings, etc...

## Notes for reviewer

Please comment and advise

Considered alternatives:
- https://github.com/joomcode/errorx?tab=readme-ov-file (decided against it to save on complexity)


tothandras:
Let's postpone this for now. It can be a working concept, but I'm afraid of introducing another error format while we don't have a clear plan standardizing them throughout the codebase.

GAlexIHU:
> Let's postpone this for now. It can be a working concept, but I'm afraid of introducing another error format while we don't have a clear plan standardizing them throughout the codebase.

As an intermediate solution, what would you say of separating the trait concept from errors? 

Instead of `pkgs/erorrsx` we could have 
```go
// pkg/traits
type Trait struct {
  id uint64
  label string
}

type WithTraits interface {
  Traits() []Trait
}

// ...
```

The reason I'm asking is because there are some further changes I want to make to subscription pkg's error handling and the trait concept makes them a lot easier that's why it came up. 

Thoughts?

PR #2068, opened 2025-01-13
Subscription custom state

GAlexIHU:
## Overview

If we want to store the `PlanReference` for custom subscription as well then we need a different way of determining what counts as a 'custom' subscription. This PR adds a simple bool flag in the state to server that purpose.


GAlexIHU:
We agreed offline that this doesn't match the current direction so this PR will be closed

PR #2071, opened 2025-01-13
fix(app): default

hekike:
- Handle deleted apps in auto defaulting logic.
- HTTP error for app default not found
- Update the app with defaulting logic
- Return app `default` property


tothandras (approved):

api/spec/src/app/app.tsp
```diff
@@ -40,6 +40,18 @@ interface Apps {
...
+  /**
+   * Update an app.
+   */
+  @patch
+  @route("/{id}")
+  @operationId("updateApp")
+  @summary("Update app")
+  update(
+    @path id: ULID,
+    @body app: Rest.Resource.ResourceReplaceModel<AppBase>,
```
tothandras, line 52: Use `CreateOrUpdateModel` instead?

tothandras (approved):

PR #2102, opened 2025-01-17
fix: disable pro rating

turip:
## Overview

We want to give the users more control about when to pro-rate, so we are disabling this feature for now.


openmeter/billing/worker/subscription/sync.go
```diff
@@ -24,10 +24,16 @@ const (
 	SubscriptionSyncComponentName billing.ComponentName = "subscription-sync"
 )
 
+type FeatureFlags struct {
```
GAlexIHU, line 27: unrelated to this PR: lets intergrate posthog FFs to the backend as well (through cloud + deps)

GAlexIHU (approved):

PR #2104, opened 2025-01-17
Add Customer filtering by subscription plan key

GAlexIHU:
## Overview

Via `planKey`


tothandras (approved):
@solkimicreb do we need filtering by version as well?

PR #2126, opened 2025-01-21
feat: currencies endpoint

turip:
## Overview

To be used by clients to understand the list of supported currencies.


GAlexIHU (approved):

api/spec/src/static/routes.tsp
```diff
@@ -0,0 +1,19 @@
+import "../rest.tsp";
+
+namespace OpenMeter.Static;
+
+using TypeSpec.Http;
+using TypeSpec.OpenAPI;
+
+@route("/api/v1/static")
+@tag("Static Data")
```
tothandras, line 9: Missing `tagMetadata`. Can we reuse an existing tag instead?

PR #2142, opened 2025-01-23
feat: implement invoice number generation

turip:
## Overview

All invoices get temporary invoice numbers, so that it's easier to reference them.

Format:
- Gathering `GATHER-{{.CustomerPrefix}}-{{.Currency}}-{{.NextSequenceNumber}}`
- Draft (before stripe provides a number) `DRAFT-{{.CustomerPrefix}}-{{.NextSequenceNumber}}`
- Sandbox finalized invoices: `"OM-SANDBOX-{{.CustomerPrefix}}-{{.NextSequenceNumber}}"`

Invoice numbers on invoices are mandatory going forward. Old invoices will have `INV-1`, `INV-2` ...


openmeter/ent/schema/billing.go
```diff
@@ -828,3 +826,29 @@ func (BillingInvoiceValidationIssue) Edges() []ent.Edge {
...
+		entutils.NamespaceMixin{},
+	}
+}
+
+func (BillingSequenceNumbers) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("scope"),
+		field.Other("last", alpacadecimal.Decimal{}).
+			SchemaType(map[string]string{
+				"postgres": "numeric",
```
GAlexIHU, line 845: this is just an int, right?

openmeter/billing/adapter/seq.go
```diff
@@ -0,0 +1,70 @@
...
+					billingsequencenumbers.Scope(input.Scope),
+				).
+				ForUpdate().
+				First(ctx)
+			if err != nil {
+				return alpacadecimal.Zero, err
+			}
+		}
+
+		next := existingRecord.Last.Add(alpacadecimal.NewFromInt(1))
```
GAlexIHU, line 57: cant we do something like 
```sql
UPDATE xy SET counter = counter + 1 WHERE ns = ... RETURNING counter;
```
with ent?

GAlexIHU (approved):

openmeter/ent/schema/billing.go
turip: Yes, but as postgres' numeric is arbritary large, I don't like running into surprises. Call me overly zealous. ;-)

openmeter/billing/adapter/seq.go
turip: As far as I have seen ent doesn't support that (but I have only spent a few minutes checking).

PR #2148, opened 2025-01-24
feat: add support for tax included and tax code settings on invoices

turip:
## Overview

This patch makes sure that our stripe app handles taxBehavior and taxCode properly, by moving to the InvoiceItem API of stripe.

Besides that it also fixes issues with invoice edit/update.

### DIFF
--- a/openmeter/app/stripe/entity/app/invoice.go
+++ b/openmeter/app/stripe/entity/app/invoice.go
@@ -182,11 +182,11 @@ func (a App) createInvoice(ctx context.Context, invoice billing.Invoice) (*billi
 	for _, line := range leafLines {
 		// Add discounts for line if any
 		for _, discount := range line.FlattenDiscountsByID() {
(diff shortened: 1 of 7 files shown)


openmeter/app/stripe/entity/app/invoice.go
```diff
@@ -439,6 +442,61 @@ func getDiscountStripeAddLinesLineParams(calculator StripeCalculator, line *bill
...
+			}
+
+			add.PriceData.ProductData.Name = lo.ToPtr(line.Name)
+		}
+	}
+
+	return add
+}
+
+func mapAddPriceDataToUpdatePriceData(add *stripe.InvoiceAddLinesLinePriceDataParams) *stripe.InvoiceUpdateLinesLinePriceDataParams {
```
turip, line 484: This is not invoked for invoice creation, but does not work.

hekike (approved):

PR #2158, opened 2025-01-27
feat(app): stripe app validate customer data

hekike:
Validate if the Stripe app customer and payment method exist before saving it.

### DIFF
--- a/openmeter/app/stripe/adapter/stripe.go
+++ b/openmeter/app/stripe/adapter/stripe.go
@@ -525,7 +525,7 @@ func (a adapter) GetSupplierContact(ctx context.Context, input appstripeentity.G
 	}
 
 	// Get Stripe App client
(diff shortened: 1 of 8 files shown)


openmeter/app/stripe/adapter/stripe.go
```diff
@@ -582,24 +582,24 @@ func (a adapter) GetMaskedSecretAPIKey(secretAPIKeyID secretentity.SecretID) (st
 }
 
 // getStripeAppClient returns a Stripe App Client based on App ID
-func (a adapter) getStripeAppClient(ctx context.Context, appID appentitybase.AppID) (stripeclient.StripeAppClient, error) {
+func (a adapter) getStripeAppClient(ctx context.Context, appID appentitybase.AppID) (appstripeentity.AppData, stripeclient.StripeAppClient, error) {
```
turip, line 585: Nit maybe return a single type instead of returning three values

turip (approved):

PR #2159, opened 2025-01-27
fix: stripe sync issues

turip:
## Overview

This patch fixes the issues found around executing the flowing scenario:
- Have a draft invoice backed by stripe app
- Replace a line in the draft invoice by another line
- Sync fails

Fixes included:
- Make sure we are assigning the new line's IDs during state machine transitions by calling the updateInvoice
- Make sure that we are always expanding the Invoice object with the service parts by using the service.updateInvoice call
- Make sure that we are always saving the invoice line stripe ID to external ID instead of the invoice item ID
- Support lookups in sync code by invoice item ID as a fallback
- [misc] Do not double-call status details calculations

### DIFF
--- a/openmeter/app/stripe/client/invoice_line.go
+++ b/openmeter/app/stripe/client/invoice_line.go
@@ -12,15 +12,51 @@ import (
 )
 
 // AddInvoiceLines is the input for adding invoice lines to a Stripe invoice.
-func (c *stripeAppClient) AddInvoiceLines(ctx context.Context, input AddInvoiceLinesInput) ([]*stripe.InvoiceItem, error) {
+func (c *stripeAppClient) AddInvoiceLines(ctx context.Context, input AddInvoiceLinesInput) ([]StripeInvoiceItemWithLineID, error) {
 	if err := input.Validate(); err != nil {
 		return nil, fmt.Errorf("stripe add invoice lines: invalid input: %w", err)
 	}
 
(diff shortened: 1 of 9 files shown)


openmeter/billing/service/invoice.go
```diff
@@ -828,6 +834,27 @@ func (s *Service) UpdateInvoice(ctx context.Context, input billing.UpdateInvoice
...
+	if err != nil {
+		return billing.Invoice{}, err
+	}
+
+	invoice, err = s.resolveWorkflowApps(ctx, invoice)
+	if err != nil {
+		return billing.Invoice{}, fmt.Errorf("error adding fields to invoice [%s]: %w", invoice.ID, err)
+	}
+
+	invoice, err = s.resolveStatusDetails(ctx, invoice)
```
turip, line 850: TODO: in most cases this will be resolved once more, so let's optimize it if possible

openmeter/app/stripe/client/invoice_line.go
```diff
@@ -12,15 +12,51 @@ import (
...
+			return nil, fmt.Errorf("stripe add invoice lines: line not found: %s", item.ID)
+		}
+
+		lines = append(lines, StripeInvoiceItemWithLineID{
+			InvoiceItem: item,
+			LineID:      lineID,
+		})
+	}
+
+	return lines, nil
```
turip, line 59: TODO: use compare for update so that retries due to rate limit will cause less calls.

chrisgacsal (approved):

PR #2163, opened 2025-01-28
Possibility to enforce billing alignment for plans

GAlexIHU:
## Overview

Defines alignment object for Plans (and later Subscriptions). Old values are defaulted as false so it's an opt-in behavior. If enabled, for a plan to be publishable all billable items in the phases have to align which is enforced by a lexically identical BillingCadence (i.e. P1M == P1M but P12M != P1Y). (this comparison would be easy to change as its localized)

## Notes for reviewer

- It will need frontend changes `openmeter.cloud` because of how subscription create is handled (the FE will need to send this new property for custom subscriptions)


api/spec/src/productcatalog/plan.tsp
```diff
@@ -62,6 +62,11 @@ model PlanReference {
 model Plan {
   ...UniqueResource;
 
+  /**
+   * Alignment configuration for the plan.
+   */
+  Alignment: Alignment;
```
hekike, line 68: should we make this optional and provide default?

openmeter/productcatalog/plan.go
```diff
@@ -85,6 +86,27 @@ func (p Plan) ValidForCreatingSubscriptions() error {
...
+	// Let's check Alignment
+	if p.Alignment.BillablesMustAlign {
+		for i, phase := range p.Phases {
+			periods := make(map[datex.ISOString]bool)
+
+			// For each phase, all RateCards that have a price associated must align
+			for _, rc := range phase.RateCards.Billables() {
+				// 1 time prices are excluded
+				if d := rc.GetBillingCadence(); d != nil {
+					periods[d.ISOString()] = true
```
hekike, line 98: why not using duration as key so P12M == P1Y works too?

hekike: Also I think you can simplify this by just taking the first period and checking if all consecutive ==

openmeter/ent/schema/productcatalog.go
```diff
@@ -198,3 +200,14 @@ var (
 	PriceValueScanner               = entutils.JSONStringValueScanner[*productcatalog.Price]()
 	DiscountsValueScanner           = entutils.JSONStringValueScanner[[]productcatalog.Discount]()
 )
+
+// AlignmentMixin for Alignment config
+type AlignmentMixin struct {
```
hekike, line 205: This mixin is only used at one place, why not just a column?

hekike (changes requested):

openmeter/ent/schema/productcatalog.go
GAlexIHU: I intend to reuse it in the next PR

openmeter/productcatalog/plan.go
GAlexIHU: We can do `d.Normalise(true)`

api/spec/src/productcatalog/plan.tsp
GAlexIHU: Would you prefer it to be on by default? (That would make more sense to me as a default)

tothandras: should be lowercase
```suggestion
  alignment: Alignment;
```

openmeter/productcatalog/alignment.go
```diff
@@ -0,0 +1,10 @@
+package productcatalog
+
+type Alignment struct {
```
tothandras, line 3: Do we need a type for this? I'd just add it to the plan.

openmeter/ent/schema/productcatalog.go
tothandras: It's just an extra hop, copying this field is fine.

api/spec/src/productcatalog/plan.tsp
hekike: Yes, default sounds good! Thanks

openmeter/productcatalog/alignment.go
GAlexIHU: Will be reused in upcoming PRs, I'll keep it as is.

api/spec/src/productcatalog/alignment.tsp
```diff
@@ -0,0 +1,21 @@
...
+ * Alignment configuration for a plan or subscription.
+ */
+@friendlyName("Alignment")
+model Alignment {
+  /**
+   * Whether all Billable items and RateCards must align.
+   * Alignment means the Price's BillingCadence must align for both duration and anchor time.
+   */
+  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
+  BillablesMustAlign?: boolean;
```
tothandras, line 20:
```suggestion
  billablesMustAlign?: boolean;
```

tothandras (changes requested):

api/spec/src/productcatalog/alignment.tsp
GAlexIHU: We should add a lint rule for these

GAlexIHU: Added one

hekike (dismissed):

tothandras (approved):

PR #2168, opened 2025-01-29
Disallow items with periods less than 1 hour

GAlexIHU:
## Overview

We didn't have any validations for how long or short the UsagePeriod can be.

## Notes for reviewer

The migration will soft-delete any such items & ratecards in the DB as otherwise they would be bricked when parsing

### DIFF
--- a/tools/migrate/migrations/20250129143720_minimum-entitlement-usageperiod.up.sql
+++ b/tools/migrate/migrations/20250129143720_minimum-entitlement-usageperiod.up.sql
@@ -1,0 +1,15 @@
+-- let's delete all subscription items where the usagePeriod is less than 1 hour
+UPDATE
+    subscription_items
+SET
+    deleted_at = NOW()
+WHERE
+    (entitlement_template ->> 'usagePeriod') :: INTERVAL < 'PT1H' :: INTERVAL;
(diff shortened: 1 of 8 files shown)


tools/migrate/migrations/20250129143720_minimum-entitlement-usageperiod.up.sql
```diff
@@ -0,0 +1,15 @@
+-- let's delete all subscription items where the usagePeriod is less than 1 hour
+UPDATE
+    subscription_items
+SET
+    deleted_at = NOW()
+WHERE
+    (entitlement_template ->> 'usagePeriod') :: INTERVAL < 'PT1H' :: INTERVAL;
```
tothandras, line 7: I think comparisons won't work with ISO durations.

tothandras (changes requested):

tools/migrate/migrations/20250129143720_minimum-entitlement-usageperiod.up.sql
GAlexIHU: They do

tothandras (approved):

PR #2171, opened 2025-01-29
feat: support stripe webhooks for payment statuses

turip:
## Overview

This patch adds support for receiving payment status updates from stripe including the void state.

The docs folder contains more details on validating and local setup.


openmeter/app/stripe/service/billing.go
```diff
@@ -0,0 +1,192 @@
...
+			PageNumber: 1,
+			PageSize:   5,
+		},
+		Expand: billing.InvoiceExpand{},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	if len(invoices.Items) == 0 {
```
hekike, line 172: shouldn't we return NotFound and handle it upstream? Like we do with DB

openmeter/billing/adapter/invoice.go
```diff
@@ -216,6 +216,17 @@ func (a *adapter) ListInvoices(ctx context.Context, input billing.ListInvoicesIn
...
 
+		if input.ExternalIDs != nil {
+			switch input.ExternalIDs.Type {
+			case billing.InvoicingExternalIDType:
+				query = query.Where(billinginvoice.InvoicingAppExternalIDIn(input.ExternalIDs.IDs...))
+			case billing.PaymentExternalIDType:
+				query = query.Where(billinginvoice.PaymentAppExternalIDIn(input.ExternalIDs.IDs...))
+			case billing.TaxExternalIDType:
+				query = query.Where(billinginvoice.TaxAppExternalIDIn(input.ExternalIDs.IDs...))
+			}
```
hekike, line 227: error if unknown type?

docs/stripe-dev.md
```diff
@@ -0,0 +1,81 @@
+# Developing against a stripe test account
```
hekike, line 1: Awesome docs!

openmeter/app/stripe/service/billing.go
turip: I want to ignore any invoices that we don't know of. E.g. if there are invoices generated on the stripe side, let's ignore them for now.

openmeter/billing/adapter/invoice.go
turip: It's not needed as it's part of the validation: https://github.com/openmeterio/openmeter/blob/667b205098558e4fb0b5cc433892b00219ac7af4/openmeter/billing/invoice.go#L641

PR #2210, opened 2025-02-04
Subscription Alignment

GAlexIHU:
## Overview

Fixes #(issue)


openmeter/subscription/item.go
```diff
@@ -18,6 +18,8 @@ type SubscriptionItem struct {
 	// The defacto cadence of the item is calculated and persisted after each change.
 	models.CadencedModel `json:",inline"`
 
+	BillingBehaviorOverride BillingBehaviorOverride `json:"billingBehaviorOverride"`
```
turip, line 21: nit: as far as I see we only use the Subscription level override, but not this one.

GAlexIHU: This is part of the reset-anchor logic (that is partially implemented in this PR)

PR #2218, opened 2025-02-05
fix: only publish events after successful commits

turip:
## Overview

This prevents invalid events being generated for transactions already rolled back.

The current implementation ignores hook errors (just calls log) as we are using this for event publishing only, where we have countermeasures for lost events (e.g. periodic jobs).


pkg/framework/transaction/transaction.go
```diff
@@ -27,6 +29,30 @@ func RunWithNoValue(ctx context.Context, creator Creator, cb func(ctx context.Co
 	return err
 }
 
+func AddPostCommitHook(ctx context.Context, logger *slog.Logger, callback func(ctx context.Context) error) {
```
GAlexIHU, line 32: I don't like that adding a hook requires providing a logger instance. The manager can have the logger. If you want specificity you can always add a stacktrace...

pkg/framework/transaction/transaction.go
```diff
@@ -67,27 +96,39 @@ func getTx(ctx context.Context, creator Creator) (context.Context, Driver, error
 // Manages the transaction based on the behavior of the callback
 func manage[R any](ctx context.Context, tx Driver, cb func(ctx context.Context, tx Driver) (R, error)) (R, error) {
 	var def R
+
+	hookMgr, err := GetHookManagerFromContext(ctx)
```
GAlexIHU, line 100: could be input if you need it to be present anyways (?)

pkg/framework/transaction/transaction.go
```diff
@@ -67,27 +96,39 @@ func getTx(ctx context.Context, creator Creator) (context.Context, Driver, error
...
 		}
 	}()
 
-	err := tx.SavePoint()
+	err = tx.SavePoint()
 	if err != nil {
 		return def, err
 	}
 
+	hookMgr.PostSavePoint()
```
GAlexIHU, line 121: i don't love the fact that `hookmanager` does everything in parallel to the txDriver, they could get out of sync, be buggy in different ways, etc... I'd much preffer if the txDriver implemented some interface through which it can manage hooks, i'd find that a lot less error prone

pkg/framework/transaction/hookmanager.go
```diff
@@ -0,0 +1,162 @@
...
+	*h = append(*h, []hook{})
+}
+
+// GetLayerCount returns the number of layers
+func (h HookLayers) GetLayerCount() int {
+	return len(h)
+}
+
+type TransactionHooks struct {
+	PostCommitHooks HookLayers
```
GAlexIHU, line 67: txDriver handles concurrency correctly, this should too in my opinion

GAlexIHU (changes requested):
If you add some locking or equivalent to hookmanager I'll approve

pkg/framework/transaction/transaction.go
turip: That can be done as an improvement.

GAlexIHU (approved):

PR #2220, opened 2025-02-05
fix: reuse gathering invoices

turip:
## Overview

If a gathering invoice (even deleted) exists for a specific line let's reuse that invoice instead of generating a new one.


chrisgacsal (approved):
Awesome! Ty :shipit:

PR #2221, opened 2025-02-05
feat: store status details in DB

turip:
## Overview

This allows for checking the DB for stuck invoices that would otherwise need to fetch data from state machine.

As a sideeffect we are also reducing the load on the DB due to some locks being unnecessary with this setup.

### DIFF
--- a/openmeter/billing/service/invoicestate.go
+++ b/openmeter/billing/service/invoicestate.go
@@ -28,6 +28,18 @@ var invoiceStateMachineCache = sync.Pool{
 	},
 }
 
+var gatheringInvoiceStatusDetails = billing.InvoiceStatusDetails{
+	Immutable: false,
+	// The invoicable state is calculated in the services recalculateGatheringInvoice for now, as the
+	// line data is available there. On the long run we need to cache this information.
+	//
+	// For now, as a safety measure we lie here, as the recalculation will be performed either ways
+	// and the CreateInvoice method will validate this once more.
+	AvailableActions: billing.InvoiceAvailableActions{
+		Invoice: &billing.InvoiceAvailableActionInvoiceDetails{},
+	},
+}
+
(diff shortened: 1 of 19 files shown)


openmeter/ent/schema/billing.go
```diff
@@ -680,6 +680,9 @@ func (BillingInvoice) Fields() []ent.Field {
 		field.Enum("status").
 			GoType(billing.InvoiceStatus("")),
 
+		field.JSON("status_details_cache", billing.InvoiceStatusDetails{}).
```
hekike, line 683: is it called cache because it's the last synced status?
If yes, how about calling it: `external_status` or `invoice_provider_status`

openmeter/billing/service/invoicestate.go
```diff
@@ -28,6 +28,18 @@ var invoiceStateMachineCache = sync.Pool{
 	},
 }
 
+var gatheringInvoiceStatusDetails = billing.InvoiceStatusDetails{
+	Immutable: false,
+	// The invoicable state is calculated in the services recalculateGatheringInvoice for now, as the
+	// line data is available there. On the long run we need to cache this information.
+	//
+	// For now, as a safety measure we lie here, as the recalculation will be performed either ways
```
hekike, line 36: is this a placeholder basically that will be updated?  If yes how about doing something like:

`var gatheringInvoiceStatusDetails = billing.NewInvoiceStatusDetails()` or
`var gatheringInvoiceStatusDetails = billing.NewEmptyInvoiceStatusDetails()`

I'm not a fan of word lie. It feels it would fake a status while it's just not calculated yet.

openmeter/ent/schema/billing.go
turip: The UpdatedAt reflects that information. Whenever we need to touch the invoice, we need to recalculate the status details either ways.

chrisgacsal (approved):

PR #2224, opened 2025-02-06
Add JS client generation to `just generate`

GAlexIHU:
## Overview

Also generate new types

### DIFF
--- a/justfile
+++ b/justfile
@@ -4,6 +4,7 @@ default:
 
 generate:
     dagger call generate server export --path api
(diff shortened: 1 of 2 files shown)


justfile
```diff
@@ -4,6 +4,7 @@ default:
 
 generate:
     dagger call generate server export --path api
+    dagger call generate javascript-sdk export --path api/client/javascript
```
tothandras, line 7:
```
dagger call generate javascript-sdk -o api/client/javascript
```

tothandras (changes requested):

PR #2239, opened 2025-02-10
Add CreateSubscription defaults and CustomerKey support 

GAlexIHU:
## Overview

As the name suggests...

Fixes OM-1179


pkg/framework/commonhttp/encoder.go
```diff
@@ -65,3 +65,11 @@ func EmptyResponseEncoder[Response any](statusCode int) httptransport.ResponseEn
 		return nil
 	}
 }
+
+// DummyErrorEncoder is a dummy error encoder that always returns a 400 status code with the received error.
+func DummyErrorEncoder() httptransport.ErrorEncoder {
```
tothandras, line 70: Is it being use anywhere?

pkg/ref/ref.go
```diff
@@ -0,0 +1,16 @@
+package ref
+
+import "fmt"
+
+type IDOrKey struct {
```
tothandras, line 5: I think we already have an `IDOrKey` type. Let's consolidate if possible.

api/spec/src/productcatalog/subscription.tsp
```diff
@@ -306,9 +306,14 @@ model CustomSubscriptionCreate {
...
-   * The ID of the customer.
+   * The ID of the customer. Provide either the key or ID. Has presedence over the key.
    */
-  customerId: ULID;
+  customerId?: ULID;
+
+  /**
+   * The key of the customer. Provide either the key or ID.
+   */
+  customerKey?: Key;
```
tothandras, line 316: The `key` on the `Customer` is string: https://github.com/openmeterio/openmeter/blob/0669fec3a5aa6bdc0d3b57ae6f1895b747eff9a6/api/spec/src/customer.tsp#L224

```suggestion
  customerKey?: string;
```

api/spec/src/productcatalog/subscription.tsp
```diff
@@ -347,9 +352,14 @@ model PlanSubscriptionCreate {
...
```
tothandras, line 362:
```suggestion
  customerKey?: string;
```

openmeter/productcatalog/subscription/http/create.go
```diff
@@ -72,6 +75,13 @@ func (h *handler) CreateSubscription() CreateSubscriptionHandler {
 					return CreateSubscriptionRequest{}, fmt.Errorf("failed to map timing: %w", err)
 				}
 
+				ref := ref.IDOrKey{
+					ID: lo.FromPtrOr(parsedBody.CustomerId, ""),
+				}
+				if ref.ID == "" {
+					ref.Key = lo.FromPtrOr(parsedBody.CustomerKey, "")
+				}
```
tothandras, line 83: We could just create the ref here.

```suggestion
				ref := ref.IDOrKey{
					ID: lo.FromPtrOr(parsedBody.CustomerId, ""),
					Key: lo.FromPtrOr(parsedBody.CustomerKey, ""),
				}
```

openmeter/productcatalog/subscription/http/create.go
```diff
@@ -104,18 +114,25 @@ func (h *handler) CreateSubscription() CreateSubscriptionHandler {
...
```
tothandras, line 122: Same as above

tothandras (dismissed):

api/spec/src/productcatalog/subscription.tsp
GAlexIHU: why?

openmeter/productcatalog/subscription/http/create.go
GAlexIHU: invalid if both are provided

api/spec/src/productcatalog/subscription.tsp
tothandras: it's an external key / id, which can have different formats

openmeter/productcatalog/subscription/http/create.go
tothandras: They both default to empty strings either way.

pkg/framework/commonhttp/encoder.go
GAlexIHU: Not in the codebase. It's hard to debug the e2e tests, I use it there if something breaks I replace the actual errorEncoder with this so I can retrieve the error. Its a 3 lines of code I write then delete 2-3 times a week, figured might as well commit it here

api/spec/src/productcatalog/subscription.tsp
GAlexIHU: Thats what `Key` is for

If we allow any string that causes a lot of issues

openmeter/productcatalog/subscription/http/create.go
GAlexIHU: nvm, i can't read

api/spec/src/productcatalog/subscription.tsp
tothandras: We allow any strings, it's customer specific and by design. cc. @hekike

GAlexIHU: I added some (mild) extra validations around it, wdyt?

openmeter/productcatalog/subscription/http/create.go
GAlexIHU: nvm, I can read, it's not equivalent as API supports providing both fields with documented priority for ID while `ref.IDOrKey` only accepts one at a time

GAlexIHU: We talked with @hekike that it might be better if we don't introduce the IDOrKey concept to the package internals so it was changed so that the handler resolves key => id

api/spec/src/productcatalog/subscription.tsp
```diff
@@ -303,12 +303,23 @@ union SubscriptionChange {
...
-  ...CustomSubscriptionChange;
+  ...OmitProperties<CustomSubscriptionChange, "timing">;
 
   /**
-   * The ID of the customer.
+   * Timing configuration for the change, when the change should take effect.
+   * The default is immediate.
    */
-  customerId: ULID;
+  timing?: SubscriptionTiming;
```
hekike, line 312: please set it to default

pkg/framework/commonhttp/encoder.go
hekike: nit: how about calling it `TestErrorEncoder` to signify it's for testing

hekike (dismissed):

openmeter/productcatalog/subscription/http/create.go
```diff
@@ -99,30 +121,75 @@ func (h *handler) CreateSubscription() CreateSubscriptionHandler {
...
+			}
+
+			customerID := request.CustomerRef.ID
+			if customerID == "" {
+				cust, err := h.CustomerService.ListCustomers(ctx, customer.ListCustomersInput{
+					Key:            lo.ToPtr(request.CustomerRef.Key),
+					Namespace:      request.WorkflowInput.Namespace,
+					Page:           pagination.NewPage(1, 1),
+					IncludeDeleted: false,
```
turip, line 175: Why not do the validation in subscription service?

turip (approved):

PR #2243, opened 2025-02-10
fix(sink): deadlock in flushhandler

chrisgacsal:
## Overview

Fix deadlock during draining events channel in flushhadler. This should fix the `failed to shutdown flush success handlers` error messages reported by `sink-worker` during shutdown.

### DIFF
--- a/openmeter/sink/flushhandler/handler.go
+++ b/openmeter/sink/flushhandler/handler.go
@@ -30,10 +30,19 @@ type FlushEventHandlerOptions struct {
 	CallbackTimeout time.Duration
 }
 
+var _ FlushEventHandler = (*flushEventHandler)(nil)
+
 type flushEventHandler struct {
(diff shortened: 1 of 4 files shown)


turip (dismissed):

openmeter/sink/flushhandler/handler.go
```diff
@@ -80,48 +89,89 @@ func NewFlushEventHandler(opts FlushEventHandlerOptions) (FlushEventHandler, err
...
 		logger:          opts.Logger,
 	}, nil
 }
 
+func (f *flushEventHandler) Close() error {
+	if f.isShutdown.Load() {
+		return nil
+	}
+
+	f.isShutdown.Store(true)
```
turip, line 128:
```suggestion
	if f.isShutdown.Swap(true) {
		return nil
	}
```

PR #2249, opened 2025-02-11
Add current cycle info to subs response

GAlexIHU:
## Overview

If the `Sub` is aligned returns the current billing period (when makes sense)

Adding this property could be a bit confusing as it might imply stronger billing guarantees than there are


solkimicreb (dismissed):
Looks good to me on a schema / api level but can't comment on go code 👀

turip (dismissed):

turip (approved):

PR #2262, opened 2025-02-13
Block Customer deletion if they have active Subscription or Entitlements

GAlexIHU:
## Overview

As the name suggests

## Notes for Reviewer

Fetching all this from the DB is not ideal but deleting customers is a fairly infrequent operation


turip:
I have implemented a minimal validation support for this yesterday here: https://github.com/openmeterio/openmeter/pull/2343

I would love if we use that here too.

GAlexIHU:
ent validation implemented here: https://github.com/openmeterio/openmeter/pull/2537
sub validation implemented already elsewhere

PR #2268, opened 2025-02-14
fix(billing): dedupe invoice validation issues

hekike:
Deduplicate validation issues to avoid:

`ERROR: ON CONFLICT DO UPDATE command cannot affect row a second time (SQLSTATE 21000)`


tothandras:
approved ✅

PR #2275, opened 2025-02-14
Hotfix sync exclusion rules

GAlexIHU:
## Overview

Fixes syncing panic happening due to excluding de-facto 0 prices from line generation

### DIFF
--- a/openmeter/billing/worker/subscription/sync.go
+++ b/openmeter/billing/worker/subscription/sync.go
@@ -488,15 +488,19 @@ func (h *Handler) provisionPendingLines(ctx context.Context, subs subscription.S
 		return fmt.Errorf("creating new lines: %w", err)
 	}
 
-	newLines = lo.Filter(newLines, func(l *billing.LineWithCustomer, _ int) bool {
-		return l != nil
-	})
-
 	_, err = h.billingService.CreatePendingInvoiceLines(ctx, billing.CreateInvoiceLinesInput{
 		Namespace: subs.Subscription.Namespace,
-		Lines: lo.Map(newLines, func(l *billing.LineWithCustomer, _ int) billing.LineWithCustomer {
-			return *l
-		}),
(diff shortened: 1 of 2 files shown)


openmeter/billing/worker/subscription/sync.go
```diff
@@ -482,9 +488,15 @@ func (h *Handler) provisionPendingLines(ctx context.Context, subs subscription.S
 		return fmt.Errorf("creating new lines: %w", err)
 	}
 
+	newLines = lo.Filter(newLines, func(l *billing.LineWithCustomer, _ int) bool {
+		return l != nil
+	})
+
```
chrisgacsal, line 494:
```suggestion
	newLines = func () {
		result := make([]billing.LineWithCustomer, 0, len(newlines))
		
		for _, line := range newLines {
			if line == nil {
				continue
			}
			
			result = append(result, *line)
		}
		
		return result
	}

```

openmeter/billing/worker/subscription/sync.go
```diff
@@ -482,9 +488,15 @@ func (h *Handler) provisionPendingLines(ctx context.Context, subs subscription.S
...
 	_, err = h.billingService.CreatePendingInvoiceLines(ctx, billing.CreateInvoiceLinesInput{
 		Namespace: subs.Subscription.Namespace,
-		Lines:     newLines,
+		Lines: lo.Map(newLines, func(l *billing.LineWithCustomer, _ int) billing.LineWithCustomer {
+			return *l
+		}),
```
chrisgacsal, line 499: This change is not needed if the suggestion from above is applied.

openmeter/billing/worker/subscription/sync.go
chrisgacsal: We can do this in one iteration.

chrisgacsal (changes requested):

chrisgacsal (approved):

PR #2305, opened 2025-02-17
Collect runtime metrics

GAlexIHU:
## Overview

Using https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/README.md


app/common/telemetry.go
```diff
@@ -227,6 +237,13 @@ func NewTelemetryHandler(metricsConf config.MetricsTelemetryConfig, healthChecke
 		router.Handle("/healthz/ready", handler)
 	}
 
+	// Start runtime metrics collector
+	{
```
turip, line 241: nit We don't need the closure here.

turip (approved):

PR #2312, opened 2025-02-18
Entitlement Reset Calculation (Entitlement Reset Part II) 

GAlexIHU:
## Overview

Part II of the Entitlement Reset
Part I: #2339

Rewrite engine calculations so they don't rely on  triggered reset actions. We build a timeline from all the manual reset times and the programmatic ones (based on entitlement's `UsagePeriod`, these were previously done via the auto reset job) and calculate grant balances seamlessly.

## Changes

- Enable engine to run across resets (localize reset calculation logic to engine)
- Introduce time abstractions to help with period and timeline calculations
- Rewrite UsageReset objects to store current anchors at time of reset (historically we overwrote this field on the entitlement)
- Snapshotting behavior is free from engine logic. We can snapshot whenever we want.
- Some of our tests were testing impossible scenarios that should not have been possible but were due to bugs. Those have been changed.

## NOTES
- balance and entitlement events published but not yet processed will fail due to entitlement schema changing

## Follow Ups
1. Disable calling `ResetEntitlementsWithExpiredUsagePeriod` in cloud (i.e. auto-reset job)
2. We could get rid of the concept of `CurrentUsagePeriod` in the DB after the job is disabled. CurrentUsagePeriod, OriginalPeriodAnchor, etc... are error prone to work with and as soon as auto reset is gone we can get rid of them.
3. Currently there's no programmatic way of creating Snapshots. This could have a performance impact as time passes.
4. Handling Snapshots could be simplified. Currently, reset always creates a snapshot. Apart from that, GetBalance will create snapshots from history no later than the current UsagePeriod start & 7 days ago (this somewhat mimics current behavior, as eventually it will result in a snapshot at every programmatic reset time). It's not clear what the snapshotting behavior should be: as long as we _eventually_ create snapshots, the system should be fine (unless we see a need for optimization). 

## TODO

- [x] test manual resets conflicting with programmatic resets (when resetting to past at time of programmatic reset)

### DIFF
--- a/openmeter/credit/balance.go
+++ b/openmeter/credit/balance.go
@@ -91,10 +91,12 @@ func (m *connector) GetBalanceOfOwner(ctx context.Context, owner grant.Namespace
 		return nil, fmt.Errorf("failed to calculate balance for owner %s at %s: %w", owner.ID, at, err)
 	}
 
-	history, err := engine.NewGrantBurnDownHistory(result.History)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create grant burn down history: %w", err)
-	}
+	// TODO: add back saving snapshots
+
+	// history, err := engine.NewGrantBurnDownHistory(result.History)
+	// if err != nil {
+	// 	return nil, fmt.Errorf("failed to create grant burn down history: %w", err)
+	// }
 
 	// FIXME: It can be the case that we never actually save anything if the history has a single segment.
 	// In practice what we can save is the balance at the last activation or recurrence event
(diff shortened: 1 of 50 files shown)


openmeter/credit/balance.go
```diff
@@ -83,43 +83,24 @@ func (m *connector) GetBalanceOfOwner(ctx context.Context, owner grant.Namespace
...
-		})
-		if err != nil {
-			return nil, fmt.Errorf("failed to save balance for owner %s at %s: %w", owner.ID, at, err)
-		}
+	// Let's see if a snapshot should be saved
+	if err := m.snapshotEngineResult(ctx, snapshotParams{
+		grants: grants,
+		owner:  owner,
+		runRes: result,
+		before: clock.Now().AddDate(0, 0, -7), // 7 days ago
```
turip, line 101: nit: I would love to make this a const with proper comments on what it means.

openmeter/credit/balance.go
```diff
@@ -37,44 +36,45 @@ type BalanceHistoryParams struct {
 var _ BalanceConnector = &connector{}
 
 func (m *connector) GetBalanceOfOwner(ctx context.Context, owner grant.NamespacedOwner, at time.Time) (*balance.Snapshot, error) {
```
turip, line 38: As far as I see the `GetBalanceOfOwner` and `GetBalanceHistoryOfOwner` is the same up to the point where we execute the engine. Can we merge that part into a function and just invoke different transformations on the engine results?

openmeter/credit/engine/grant.go
```diff
@@ -120,6 +121,21 @@ func (e *engine) getGrantRecurrenceTimes(period timeutil.Period) ([]struct {
 	return deduped, nil
 }
 
+// A grant is relevant if its active at any point during the period, both limits inclusive
+// A grant is also relevant if it is mentioned in the balance map
```
turip, line 125: nit: Please elaborate why it's needed.

openmeter/credit/engine/reset.go
```diff
@@ -0,0 +1,60 @@
...
+	if behavior.PreserveOverage {
+		startingOverage = snap.Overage
+	}
+
+	prioritizedGrants := slices.Clone(grants)
+	if err := PrioritizeGrants(prioritizedGrants); err != nil {
+		return balance.Snapshot{}, fmt.Errorf("failed to prioritize grants: %w", err)
+	}
+
+	rolledOver, _, startingOverage, err := BurnDownGrants(rolledOver, prioritizedGrants, startingOverage)
```
turip, line 50: nit: Why not return a struct, this is getting crowded in here ;-)

openmeter/entitlement/metered/grant_owner_adapter.go
```diff
@@ -107,19 +108,52 @@ func (e *entitlementGrantOwner) GetStartOfMeasurement(ctx context.Context, owner
 	return metered.MeasureUsageFrom, nil
 }
 
+// The current usage period start time is either the current period start time, or if this is the first period then the start of measurement
 func (e *entitlementGrantOwner) GetUsagePeriodStartAt(ctx context.Context, owner grant.NamespacedOwner, at time.Time) (time.Time, error) {
-	// If this is the first period then return start of measurement, otherwise calculate based on anchor.
-	// To know if this is the first period check if usage has been reset.
+	ent, err := e.entitlementRepo.GetEntitlement(ctx, owner.NamespacedID())
```
turip, line 113: Maybe not for this specific PR, but these should be refactored to not to fetch the entitlement multiple times during engine building. E.g. this one is called 2-3 times during a balance query, having to refetch entitlements multiple times from the db.

turip (approved):
What I am missing is the emulation of the CurrentUsagePeriod (so until that is solved the API will return previous usage periods for entitlements whose current period expired).

Or am I missing something? I suppose that will be one of the next PRs to get rid of that field and send the calculated data our customers.

turip (dismissed):

turip (approved):

PR #2322, opened 2025-02-22
refactor(meter): service

hekike:
- Move meter model to meter package
- Meter httphandler
- Use API generated types in meter response


openmeter/meter/adapter/adapter.go
```diff
@@ -0,0 +1,42 @@
+package adapter
```
GAlexIHU, line 1: i think inmemory was a more descriptive name for this

GAlexIHU (approved):

PR #2352, opened 2025-02-27
Add validation for Continuing

GAlexIHU:
## Overview

Fixes [OM-1226: Subscription disallow "Continue" if a future sub is scheduled](https://linear.app/openmeter/issue/OM-1226/subscription-disallow-continue-if-a-future-sub-is-scheduled)


GAlexIHU:
Implemented as part of rewrite here https://github.com/openmeterio/openmeter/pull/2424

PR #2376, opened 2025-03-04
chore: date calculation research

turip:
## Overview

Fixes #(issue)

### DIFF
--- a/pkg/timeutil/out.md
+++ b/pkg/timeutil/out.md
@@ -1,0 +1,128 @@
+# intro
+
+Dump of test case outputs.
+
+We are generating period starting from an anchor.
+
+Example output:
+```
+iteration[0]:
+	exp: [2024-12-31T00:00:00Z..2025-01-31T00:00:00Z]
+	got: [2024-12-31T00:00:00Z..2025-01-31T00:00:00Z]
+	v2:  [2024-12-31T00:00:00Z..2025-01-31T00:00:00Z]
+```
+
+iteration: which period's index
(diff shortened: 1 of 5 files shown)


turip (commented):
DO NOT MERGE!

pkg/timeutil/out.md
```diff
@@ -0,0 +1,128 @@
+# intro
+
+Dump of test case outputs.
```
turip, line 3: @hekike please validate if this is the expected output

pkg/timeutil/out.md
```diff
@@ -0,0 +1,128 @@
...
+	v2:  [2024-02-29T00:00:00Z..2025-02-28T00:00:00Z]
+iteration[1]:
+	exp: [2025-02-28T00:00:00Z..2026-02-28T00:00:00Z]
+	got: [2025-03-01T00:00:00Z..2026-03-01T00:00:00Z]
+	v2:  [2025-02-28T00:00:00Z..2026-02-28T00:00:00Z]
+
+
+## Daylight savings changes - anchor has timezone information
+
+Anchor: `2025-02-01T13:00:00+01:00` (BP Time)
```
turip, line 67: Right now we are storing all times in UTC. @hekike don't we want to make an exception in case of anchors to store the tz information in the DB.

The anchor's tz information is crucial to preserve timestamps.

hekike: I'm okay with storign TZ

PR #2392, opened 2025-03-08
Feat/query cache

hekike:
## Overview

Fixes #(issue)

### DIFF
--- a/etc/seed/seed.yaml
+++ b/etc/seed/seed.yaml
@@ -5,7 +5,7 @@ input:
     # batch_size: 1
     mapping: |
       let max_subjects = ${SEEDER_MAX_SUBJECTS:20}
-      let event_type = "prompt"
+      let event_type = "tokens"
       let source = "api-gateway"
       let models = ["gpt-3", "gpt-4"]
       let types = ["input", "output"]
--- a/etc/seed/seed.yaml
+++ b/etc/seed/seed.yaml
@@ -5,12 +5,12 @@ input:
     # batch_size: 1
     mapping: |
       let max_subjects = ${SEEDER_MAX_SUBJECTS:20}
-      let event_type = "tokens"
+      let event_type = "prompt"
       let source = "api-gateway"
       let models = ["gpt-3", "gpt-4"]
       let types = ["input", "output"]
       let subject = "customer-%d".format(random_int(seed: timestamp_unix_nano()) % $max_subjects)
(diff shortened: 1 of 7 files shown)


openmeter/streaming/clickhouse/cache.go
```diff
@@ -0,0 +1,276 @@
...
+func (c *Connector) fetchCachedMeterRows(ctx context.Context, hash string, queryMeter queryMeter) ([]meterpkg.MeterQueryRow, error) {
+	var cachedValues []meterpkg.MeterQueryRow
+
+	cacheQuery := getMeterQueryRowsFromCache{
+		Database:  queryMeter.Database,
+		TableName: meterQueryRowCacheTable,
+		Hash:      hash,
+		Namespace: queryMeter.Namespace,
+		From:      queryMeter.From,
+		To:        queryMeter.To,
```
turip, line 177: I would mandate To to be not nil (maybe from too, but that's less problematic), so that we are not storing any cache item that's "open ended"

openmeter/streaming/clickhouse/cache.go
```diff
@@ -0,0 +1,276 @@
...
+	// Step 1: Look up cached rows
+	cachedValues, err := c.fetchCachedMeterRows(ctx, hash, cacheableQueryMeter)
+	if err != nil {
+		return originalQueryMeter, values, fmt.Errorf("failed to lookup cached meter rows: %w", err)
+	}
+
+	// If we have cached values, add them to the results
+	// Also, update the query range to query uncached periods
+	if len(cachedValues) > 0 {
+		c.config.Logger.Debug("cached rows found", "from", cacheableQueryMeter.From, "to", cacheableQueryMeter.To, "count", len(cachedValues))
```
turip, line 90: I would consider adding some metrics that would allow us to validate the hit rate of the cache, as frequent writes might make the cache entirely disabled.

turip (approved):

openmeter/streaming/clickhouse/cache.go
hekike: We discussed it separately. With open-ended `to` we don't cache the last 3 days (default, see `QueryCacheMinimumCacheableUsageAge`)
See: https://github.com/openmeterio/openmeter/pull/2392/files#diff-42194278b2abea7e30a5260024bfbac0a8c925e4fa207230bb2d31016a22cfafR133-R157

PR #2426, opened 2025-03-12
Add Restore Functionality

GAlexIHU:
## Overview

- Add restore API
- Fix bug so we don't returned deleted Subs internally

## Summary by CodeRabbit

- **New Features**
	- Introduced a new option allowing users to restore a canceled subscription. This enhancement reinstates the subscription immediately—cancelling any scheduled future subscriptions—and addresses potential scheduling conflicts.
- **Tests**
	- Expanded test coverage to ensure the restoration process performs reliably across various scenarios.


openmeter/subscription/service/workflowservice.go
```diff
@@ -167,3 +168,39 @@ func (s *workflowService) ChangeToPlan(ctx context.Context, subscriptionID model
...
+
+	// Let's filter out the current sub if present
+	scheduled = lo.Filter(scheduled, func(s subscription.Subscription, _ int) bool {
+		return s.NamespacedID != subscriptionID
+	})
+
+	return transaction.Run(ctx, s.TransactionManager, func(ctx context.Context) (subscription.Subscription, error) {
+		// Let's delete all scheduled subs
+		for _, sch := range scheduled {
+			if err := s.Service.Delete(ctx, sch.NamespacedID); err != nil {
```
turip, line 198: Shouldn't we instead ask the users to cancel the scheduled subscriptions? It feels strange that a continue might interfere with scheduled subs.

turip (approved):

openmeter/subscription/service/workflowservice.go
GAlexIHU: Yeah, we added this aggregate operation to match user intent. There might be some edge case scenarios where you'd want to separately manage all scheduled subscriptions and you can do that with the existing APIs (cancel, unschedule-cancel). This `restore` is the main path where you just want to go back to the current one. I also think its a bit awkward but this was the internal agreement, I can send you the docs in private.

PR #2440, opened 2025-03-14
Add Expand to customer subscription listing

GAlexIHU:
## Overview

- Adds expand flag to customerSubscriptions API
- Expanding is done in memory (as always), upper pagesize limit is added to guard against abuse
- return types were converted to use `mo.Either`

## Summary by CodeRabbit

- **New Features**
  - Introduced an optional query parameter that allows customers to request an expanded view of subscription details.
  - Public API endpoints and client integrations now support retrieving additional subscription information when desired.
  - Updated API documentation reflects the new parameter.

- **Refactor**
  - Improved response handling to support both standard and expanded subscription views.
  - Enforced a maximum page size for requests using the expanded view option to ensure efficient processing.


api/spec/src/customer.tsp
```diff
@@ -113,6 +113,7 @@ interface Customers {
   @summary("List customer subscriptions")
   listCustomerSubscriptions(
     @path customerId: ULID,
+    @query expand?: boolean = false,
```
turip, line 116: It might be better to use an enum here, just so that we can add more granual expanding later.

turip (approved):

api/spec/src/customer.tsp
GAlexIHU: to be honest i think "expand" is on of the worst things you can incorporate into a REST API... i'll rename it to expandToView (and the we can deprecate it later if we need to)

turip (approved):

api/spec/src/customer.tsp
turip: Rename is fine for now. I don't like expand either, but we should stick to the conventions we have.

api/spec/src/customer.tsp
```diff
@@ -113,6 +113,7 @@ interface Customers {
   @summary("List customer subscriptions")
   listCustomerSubscriptions(
     @path customerId: ULID,
+    @query expandToView?: boolean = false,
```
tothandras, line 116: Please follow the patterns used at other endpoints with `expand` and an enum. If not specified only the `id` is returned in the object.

tothandras (changes requested):

api/spec/src/customer.tsp
```diff
@@ -113,6 +113,7 @@ interface Customers {
...
     ...PaginatedQuery,
   ): PaginatedResponse<ProductCatalog.Subscription> | NotFoundError | CommonErrors;
```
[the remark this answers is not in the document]
tothandras: @GAlexIHU are these two fundamentally different types, or one is an extension of the other? If the former, let's create another API endpoint, the clients can't decide automatically based on query param which type will be returned.

openmeter/subscription/list.go
```diff
@@ -24,4 +29,8 @@ func (i ListSubscriptionsInput) Validate() error {
 	return nil
 }
 
-type SubscriptionList = pagination.PagedResponse[Subscription]
+type (
+	PagedSubscriptions     = pagination.PagedResponse[Subscription]
+	PagedSubscriptionViews = pagination.PagedResponse[SubscriptionView]
+	SubscriptionList       = mo.Either[PagedSubscriptions, PagedSubscriptionViews]
```
turip, line 35: I am concerned about this. In essence this will mean that wherever in the code we are using the List call from the service we will have to do something like this:

```
subsView, ok := res.Right()
if !ok {
   return error.New("list did not return what we wanted")
}
```

In other parts of the system we can provide a unified struct, with expandable fields. I feel like that `SubscriptionView` is an extension of `Subscription` (as it's properly captured in the typespec).

Can we just make this change to the golang part, and not use mo?

GAlexIHU: I also don't particularly like it, i just found it to be the lesser evil. I wouldn't rewrite the structure of `SubscriptionView` at his point as that would take some time. 

SubscriptionView has validations & consistency checks built in all around, as that type is used across the subscription logic. Making its parts optional, "expandable" would break every validation, and definitely some parts of the logic.

We could create _another_ type that SubscriptionView can be transformed to, that is a partial, which managed the "expanding" (really rather removing parts in this case), but that i feel like is another unnecessary type & complications -- alternatively we could write expanding for subscriptions, again on a new type, where the validations wouldn't be present

Probably the simplest option is to just split the service api in two, or do the mapping on another layer of abstraction that we don't expect to reused (e.g. something handler / http specific) - which i would most likely go with after all this discussion as noone seems really happy with the current solution

turip: From the backend perspective, if we rule out the common type (which I can understand from the time perspective and knowing the module), I would prefer to have a `List` service call, and a `ListView` service call, which calls the `List` and wastefully calculates the view. (This is aligned with the Get/GetView which can be the same call if the same type is used in the future).

For the typespec perspective I would defer it to @tothandras (and that will drive the how the HTTP driver interfaces the service).

As I see the Get endpoint always returns the expanded version, and I understand why we don't want to do it for the list, but if we change the return type of the list that's a breaking change for existing customers using the subscription.

PR #2445, opened 2025-03-14
fix: checkout session event serialization

turip:
## Overview

Let's not expose non-discriminated union fields via events.

## Summary by CodeRabbit

- **Refactor**
	- Streamlined payment setup events by removing extra application-specific details and associated validations, and updated the event version to v2.
- **New Features**
	- Enhanced data output by exposing the namespace attribute in API responses.


pkg/models/model.go
```diff
@@ -136,7 +136,7 @@ func (m ManagedModel) IsDeletedAt(t time.Time) bool {
 }
 
 type NamespacedModel struct {
-	Namespace string `json:"-" yaml:"-"`
```
turip, line 139: This is an ancient heritage from the old times where this was used in the API layer. This makes event sending quite hard, so let's get rid of this now that API layer does not depend on this anymore.

hekike (approved):

chrisgacsal (approved):

PR #2478, opened 2025-03-18
Entitlement Batch Update methods

GAlexIHU:
## Overview

To fix generating reset events we'll do batch updates of entitlements updating their new UsagePeriod. This PR includes the methods enabling this
- New method for batch updates
- Change listing method to not overwrite `CurrentUsagePeriod` with calculated values

## Notes for reviewer

- Let's wait with merging this until the relevant changes in cloud are implemented, otherwise update would break

## Summary by CodeRabbit

- **New Features**
	- Enhanced performance by introducing a unique index for improved entitlement retrieval.
	- Enabled efficient batch updating of current entitlement usage periods.
  
- **Refactor**
	- Streamlined and standardized entitlement management logic for more consistent data handling.
  
- **Tests**
	- Expanded automated tests to verify new update and query functionalities.

### DIFF
--- a/openmeter/ent/schema/entitlement.go
+++ b/openmeter/ent/schema/entitlement.go
@@ -74,7 +74,7 @@ func (Entitlement) Indexes() []ent.Index {
 		index.Fields("namespace", "id", "subject_key"),
 		index.Fields("namespace", "feature_id", "id"),
 		index.Fields("namespace", "current_usage_period_end"),
-		// Index for collecting entitlements with due rsets
+		// Index for collecting entitlements with due resets
 		index.Fields("current_usage_period_end", "deleted_at"),
 		index.Fields("created_at", "id").Unique(),
 	}
--- a/openmeter/ent/schema/entitlement.go
+++ b/openmeter/ent/schema/entitlement.go
@@ -74,8 +74,9 @@ func (Entitlement) Indexes() []ent.Index {
 		index.Fields("namespace", "id", "subject_key"),
 		index.Fields("namespace", "feature_id", "id"),
 		index.Fields("namespace", "current_usage_period_end"),
-		// index for collecting all the entitlements with due resets
+		// Index for collecting entitlements with due rsets
 		index.Fields("current_usage_period_end", "deleted_at"),
+		index.Fields("created_at", "id").Unique(),
 	}
 }
 
--- a/openmeter/entitlement/adapter/entitlement.go
+++ b/openmeter/entitlement/adapter/entitlement.go
@@ -559,16 +559,30 @@ func (a *entitlementDBAdapter) UpsertEntitlementCurrentPeriods(ctx context.Conte
 				dbUpdates = append(dbUpdates, create)
 			}
 
-			// Let's do a batch insert with on conflict do update
-			err = repo.db.Entitlement.CreateBulk(dbUpdates...).
-				OnConflict(
-					sql.ConflictColumns(db_entitlement.FieldID),
-					sql.ResolveWith(func(u *sql.UpdateSet) {
-						u.SetExcluded(db_entitlement.FieldCurrentUsagePeriodStart).
-							SetExcluded(db_entitlement.FieldCurrentUsagePeriodEnd)
-					})).Exec(ctx)
+			// Let's try to come up with a sensible limiting to avoid hitting PG's limit on max number of parameters
+			// We'll assume each upsert contributing len(Columns) parameters
+			// Let's also subtract 4 for the ON CONFLICT columns
+			dbUpdatesChunks := lo.Chunk(dbUpdates, (MAX_POSTGRES_QUERY_PARAMS-4)/len(db_entitlement.Columns))
 
-			return err
+			// Let's preserve the atomic nature of the operation by running inside a transaction
+			return transaction.RunWithNoValue(ctx, repo, func(ctx context.Context) error {
+				for _, chunk := range dbUpdatesChunks {
+					// Let's do a batch insert with on conflict do update
+					// Let's do a batch insert with on conflict do update
+					err = repo.db.Entitlement.CreateBulk(chunk...).
+						OnConflict(
+							sql.ConflictColumns(db_entitlement.FieldID),
+							sql.ResolveWith(func(u *sql.UpdateSet) {
+								u.SetExcluded(db_entitlement.FieldCurrentUsagePeriodStart).
+									SetExcluded(db_entitlement.FieldCurrentUsagePeriodEnd)
+							})).Exec(ctx)
+					if err != nil {
+						return err
+					}
+				}
+
+				return nil
+			})
 		},
 	)
 }
@@ -790,3 +804,7 @@ func withLatestUsageReset(q *db.EntitlementQuery, namespaces []string) *db.Entit
 		}
 	})
 }
+
+const (
+	MAX_POSTGRES_QUERY_PARAMS = 65535
+)
--- a/openmeter/entitlement/adapter/entitlement.go
+++ b/openmeter/entitlement/adapter/entitlement.go
@@ -68,7 +68,7 @@ func (a *entitlementDBAdapter) GetEntitlement(ctx context.Context, entitlementID
 				return nil, err
 			}
 
-			return mapEntitlementEntity(res, true), nil
+			return mapEntitlementEntity(res), nil
 		},
 	)
 }
@@ -99,7 +99,7 @@ func (a *entitlementDBAdapter) GetActiveEntitlementOfSubjectAt(ctx context.Conte
 				return nil, err
 			}
 
-			return mapEntitlementEntity(res, true), nil
+			return mapEntitlementEntity(res), nil
 		},
 	)
 }
@@ -147,7 +147,7 @@ func (a *entitlementDBAdapter) CreateEntitlement(ctx context.Context, ent entitl
 				return nil, err
 			}
 
-			return mapEntitlementEntity(res, true), nil
+			return mapEntitlementEntity(res), nil
 		},
 	)
 }
@@ -261,7 +261,7 @@ func (a *entitlementDBAdapter) GetActiveEntitlementsOfSubject(ctx context.Contex
 
 			result := make([]entitlement.Entitlement, 0, len(res))
 			for _, e := range res {
-				result = append(result, *mapEntitlementEntity(e, true))
+				result = append(result, *mapEntitlementEntity(e))
 			}
 
 			return result, nil
@@ -383,7 +383,7 @@ func (a *entitlementDBAdapter) ListEntitlements(ctx context.Context, params enti
 
 				mapped := make([]entitlement.Entitlement, 0, len(entities))
 				for _, entity := range entities {
-					mapped = append(mapped, *mapEntitlementEntity(entity, true))
+					mapped = append(mapped, *mapEntitlementEntity(entity))
 				}
 
 				response.Items = mapped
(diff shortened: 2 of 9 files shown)


openmeter/entitlement/adapter/entitlement.go
```diff
@@ -499,21 +499,133 @@ func (a *entitlementDBAdapter) UpdateEntitlementUsagePeriod(ctx context.Context,
...
+				Where(db_entitlement.IDIn(slicesx.Map(updates, func(u entitlement.UpsertEntitlementCurrentPeriodElement) string {
+					return u.ID
+				})...),
+				).
+				All(ctx)
+			if err != nil {
+				return err
+			}
+
+			if len(entitlements) != len(updates) {
```
GAlexIHU, line 528: This looks like a check but in reality between these two requests entitlements _can_ disappear and there isn't much we can do against it with this batch approach. OpenMeter doesn't delete from the entitlement table (only softdelete), so in theory we should be fine, but we can result in problematic states nevertheless.

openmeter/ent/schema/entitlement.go
```diff
@@ -67,8 +67,9 @@ func (Entitlement) Indexes() []ent.Index {
 		index.Fields("namespace", "id", "subject_key"),
 		index.Fields("namespace", "feature_id", "id"),
 		index.Fields("namespace", "current_usage_period_end"),
-		// index for collecting all the entitlements with due resets
+		// Index for collecting entitlements with due resets
 		index.Fields("current_usage_period_end", "deleted_at"),
+		index.Fields("current_usage_period_end", "created_at", "id").Unique(),
```
GAlexIHU, line 72: We could remove `current_usage_period_end` so this works as a generic cursor, right now it's used combined for both the WHERE's and the ordering which works in this exact case but otherwise it wouldn't come handy. Will check query plans and get back here

openmeter/entitlement/adapter/entitlement.go
```diff
@@ -499,21 +499,133 @@ func (a *entitlementDBAdapter) UpdateEntitlementUsagePeriod(ctx context.Context,
...
+			if len(params.Namespaces) > 0 {
+				query = query.Where(db_entitlement.NamespaceIn(params.Namespaces...))
+			}
+
+			// Let's order by end, created_at, id asc
+			query = query.Order(
+				db_entitlement.ByCurrentUsagePeriodEnd(sql.OrderAsc()),
+				db_entitlement.ByCreatedAt(sql.OrderAsc()),
+				db_entitlement.ByID(sql.OrderAsc()),
+			)
```
GAlexIHU, line 596: Same as here https://github.com/openmeterio/openmeter/pull/2478/files#r2001427850

openmeter/entitlement/adapter/entitlement.go
```diff
@@ -405,7 +405,7 @@ func (a *entitlementDBAdapter) ListEntitlements(ctx context.Context, params enti
 	)
 }
 
-func mapEntitlementEntity(e *db.Entitlement) *entitlement.Entitlement {
+func mapEntitlementEntity(e *db.Entitlement, calculateUsagePeriod bool) *entitlement.Entitlement {
```
turip, line 408: Please don't use naked bool args.

Either please do:

```
type mapEntitlementInput struct {
      ent *db.Entitlement
      calculateUsagePeriod bool
}
func mapEntitlementEntity(in mpEntitlementInput) *entitlement.Entitlement {
```

or

```
type entitlementUsagePeriodCalculation int

const (
calculateCurrentUsagePeriod entitlementUsagePeriodCalculation = iota 
retainCurrentUsagePeriod
)
```

Or given that there's only one place where you set this to false, you can consider options

```
type options func(*mapEntitlementOptions)
mapEntitlementEntity(e *db.Entitlement, options...)
```

and if the default is true, then you can change
```
mapEntitlementEntity(e, true) -> mapEntitlementEntity(e)
mapEntitlementEntity(e, false) -> mapEntitlementEntity(e, disableUsagePeriodRecalculation())
```

openmeter/entitlement/adapter/entitlement.go
```diff
@@ -499,21 +499,133 @@ func (a *entitlementDBAdapter) UpdateEntitlementUsagePeriod(ctx context.Context,
...
+					SetFeatureKey(ent.FeatureKey).
+					SetSubjectKey(ent.SubjectKey).
+					SetCurrentUsagePeriodStart(update.CurrentUsagePeriod.From).
+					SetCurrentUsagePeriodEnd(update.CurrentUsagePeriod.To)
+
+				dbUpdates = append(dbUpdates, create)
+			}
+
+			// Let's do a batch insert with on conflict do update
+			err = repo.db.Entitlement.CreateBulk(dbUpdates...).
```
turip, line 561: defensive coding: Let's do this in chunks (lo.Chunk), as entgo yields this query.

```
INSERT INTO "entitlements" ("created_at", "current_usage_period_end", "current_usage_period_start", "entitlement_type", "feature_id", "feature_key", "id", "namespace", "subject_key", "updated_at") VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10), ($11, $12, $13, $14, $15, $16, $17, $18, $19, $20) ON CONFLICT ("id") DO UPDATE SET "current_usage_period_start" = "excluded"."current_usage_period_start", "current_usage_period_end" = "excluded"."current_usage_period_end" RETURNING "id"
postgres-1    | 2025-03-19 07:31:19.659 UTC [79] DETAIL:  parameters: $1 = '2025-03-19 07:31:19.659434+00', $2 = '2025-04-01 00:00:00+00', $3 = '2025-03-01 00:00:00+00', $4 = 'metered', $5 = '01JPPN0PZ57BA2T3QKNK96MGYM', $6 = 'feature1', $7 = '01JPPN0PZ7G89JQR3EMFN8PMTS', $8 = 'ns1', $9 = 'subject1', $10 = '2025-03-19 07:31:19.659434+00', $11 = '2025-03-19 07:31:19.659434+00', $12 = '2025-06-01 00:00:00+00', $13 = '2025-05-01 00:00:00+00', $14 = 'metered', $15 = '01JPPN0PZ57BA2T3QKNK96MGYM', $16 = 'feature1', $17 = '01JPPN0PZ95BJ19YYD5JBTQDSR', $18 = 'ns1', $19 = 'subject2', $20 = '2025-03-19 07:31:19.659434+00'
```

PostgreSQL allows 64k parameters, so let's do chunks of `(65535 / len(db_entitlement.Columns)) - 1`

turip (commented):
👍

openmeter/entitlement/adapter/entitlement.go
GAlexIHU: I'd much rather have the caller be responsible for calling the methods with reasonable payload size

turip: It's the adapter's responsiblility to know the underlying database's limits. That's why we have this layer. Ideally the caller would never call with that amount of data of course.

openmeter/ent/schema/entitlement.go
```diff
@@ -67,8 +67,9 @@ func (Entitlement) Indexes() []ent.Index {
 		index.Fields("namespace", "id", "subject_key"),
 		index.Fields("namespace", "feature_id", "id"),
 		index.Fields("namespace", "current_usage_period_end"),
-		// index for collecting all the entitlements with due resets
+		// Index for collecting entitlements with due resets
 		index.Fields("current_usage_period_end", "deleted_at"),
+		index.Fields("created_at", "id").Unique(),
```
GAlexIHU, line 72: technically we don't even need `created_at` as `id` is ULID and that is lexicographically sortable

openmeter/entitlement/adapter/entitlement.go
GAlexIHU: - I don't want to introduce a separate error for something we wouldn't handle and what we already get an error for (from pg)
- I'm fine with this method failing above ~8000 rows. I would never want to make it work above that nor want to use it with that many inputs.

openmeter/entitlement/adapter/entitlement.go
```diff
@@ -499,21 +499,127 @@ func (a *entitlementDBAdapter) UpdateEntitlementUsagePeriod(ctx context.Context,
...
+				OnConflict(
+					sql.ConflictColumns(db_entitlement.FieldID),
+					sql.ResolveWith(func(u *sql.UpdateSet) {
+						u.SetExcluded(db_entitlement.FieldCurrentUsagePeriodStart).
+							SetExcluded(db_entitlement.FieldCurrentUsagePeriodEnd)
+					})).Exec(ctx)
```
turip, line 567:
```suggestion
             for _, batch := lo.Chunk(dbUpdates, (65535 / len(db_entitlement.Columns)) - 1) {
			err = repo.db.Entitlement.CreateBulk(batch...).
				OnConflict(
					sql.ConflictColumns(db_entitlement.FieldID),
					sql.ResolveWith(func(u *sql.UpdateSet) {
						u.SetExcluded(db_entitlement.FieldCurrentUsagePeriodStart).
							SetExcluded(db_entitlement.FieldCurrentUsagePeriodEnd)
					})).Exec(ctx)
				if err != nil {
				  return err
				  }
			}
```

There's no separate error.

turip: (Of course this is not commitable, I just made it up ;-) )

GAlexIHU: I see your keenness on adding this, so I will, but just to iterate my opinion a paper trail again:

1. looking out for this is arbitrary: there are lots of system limitations we don't look out for mostly because we dont need to
2. i think we're in agreement that this scenario is misuse: we shouldn't be calling with this many updates so why allow for it?
3. the exact batching is arbitrary, its a "_safe_ upper bound", but still depends on the query. having an exact bound check is not trivial to do efficiently due to how ent works (without templating the query after each update)... `len(Columns)` isn't even necessarily an upper bound for all queries, e.g. when edges are set... but getting the exact param count is problematic:
```go
func getNumberOfQueryParamsFromQuery(q *db.EntitlementCreate) int {
	// Each EntitlementCreate query creates parameters for these fields if they're set
	// This is based on the fields in EntitlementMutation
	count := 0

	// Access the mutation directly to check which fields are set
	mutation := q.Mutation()

	// Required fields - always present in a create query
	// ID, namespace, entitlement_type, feature_id, feature_key, subject_key
	count += 6

	// Time fields - most creations will have these
	// created_at, updated_at
	count += 2

	// Optional fields - need to check if they're set
	if _, exists := mutation.Metadata(); exists {
		count++
	}
	if _, exists := mutation.DeletedAt(); exists {
		count++
	}
	if _, exists := mutation.ActiveFrom(); exists {
		count++
	}
	if _, exists := mutation.ActiveTo(); exists {
		count++
	}
	if _, exists := mutation.MeasureUsageFrom(); exists {
		count++
	}
	if _, exists := mutation.IssueAfterReset(); exists {
		count++
	}
	if _, exists := mutation.IssueAfterResetPriority(); exists {
		count++
	}
	if _, exists := mutation.IsSoftLimit(); exists {
		count++
	}
	if _, exists := mutation.PreserveOverageAtReset(); exists {
		count++
	}
	if _, exists := mutation.Config(); exists {
		count++
	}
	if _, exists := mutation.UsagePeriodInterval(); exists {
		count++
	}
	if _, exists := mutation.UsagePeriodAnchor(); exists {
		count++
	}
	if _, exists := mutation.CurrentUsagePeriodStart(); exists {
		count++
	}
	if _, exists := mutation.CurrentUsagePeriodEnd(); exists {
		count++
	}
	if _, exists := mutation.SubscriptionManaged(); exists {
		count++
	}

	// Also count any edge relationships that might be set
	// Note: We're not checking if edges are actually set to keep this simpler,
	// as they're rarely used in bulk operations for entitlements

	return count
}
```
4. this again opens deadlocks as (assuming everything is a conflict) we're practically acquiring EXCLUSIVE locks in 10k batches for the length of our transaction. if another transaction separately tries to lock 2 rows from two different batches in the unlucky order, we're deadlocked

With that in mind, I'll implement the exact form of this validation you suggested

PR #2481, opened 2025-03-19
Enable granting on subs managed entitlements

GAlexIHU:
## Overview

Might make sense to reenable this due to user requests.

### Context (copied from slack)
Some reasons it made sense to forbid this:
- **non-intuitive behavior** (i think the main reason): the `entitlementId` changes with edits, phase changes, etc... the user expectation might be that it stays the same as long as the user has access to the feature

_(i'm actually in favor of changing this behavior, most likely as an additional effort as part of / after addons)_

- most usecases can be covered by editing the subscription instead: for a very temporary grant this doesn't really make sense (as UI doesn't allow queuing multiple edits, you can't really time the changes), but otherwise a good approach

## Summary by CodeRabbit

- **Refactor**
	- Streamlined the permission granting process by removing intermediary validation steps, resulting in a more direct workflow with updated error handling. These changes simplify the underlying process, potentially affecting responses in specific edge cases.

### DIFF
--- a/openmeter/entitlement/driver/metered.go
+++ b/openmeter/entitlement/driver/metered.go
@@ -120,19 +120,6 @@ func (h *meteredEntitlementHandler) CreateGrant() CreateGrantHandler {
 			return req, nil
 		},
 		func(ctx context.Context, request CreateGrantHandlerRequest) (api.EntitlementGrant, error) {
-			ent, err := h.entitlementConnector.GetEntitlementOfSubjectAt(ctx, request.Namespace, request.SubjectKey, request.EntitlementIdOrFeatureKey, clock.Now())
-			if err != nil {
-				return api.EntitlementGrant{}, err
-			}
-
-			if ent == nil {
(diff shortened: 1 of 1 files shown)


GAlexIHU:
closing as discussed, this is not the intended way to use OpenMeter

christopherhwood:
I realize this is an old PR but I'm curious why this is not an intended use case for OpenMeter to support? It seems this is a very natural way to support subscriptions + overage. 

I don't believe this can be accomplished by editing the subscription as the subscription is just a standard base and overage should be something extra that's variable-charged on top of that.

I also tried to do this with add-ons but it doesn't work because the add-ons renew on the same basis as the underlying subscription, but overages should be one-off additions to the entitlement and not renew.

It seems this is exactly what grants were designed for, but I can't figure out why grants cannot be applied to entitlements managed by subscriptions. With this restriction in place, it seems the only way to get the desired behavior is to basically reimplement subscriptions using raw entitlements and apply grants onto those.

Is there some other suggested way to manage this with OpenMeter? cc @GAlexIHU

PR #2490, opened 2025-03-19
Small API Fixes

GAlexIHU:
## Overview

- Properly return 400 error on zero customizations for Subscription Edit
- Set `om_subscription_managed` metadata annotation for managed subscriptions

## Summary by CodeRabbit

- **New Features**
  - Updated the API documentation to reflect the latest service definitions.
  - Enforced a new requirement ensuring that subscription modifications include at least one customization.
  - Enhanced subscription responses by appending an indicator for managed subscriptions.
  - Introduced a new annotation for entitlement management.

- **Bug Fixes**
  - Improved error reporting for subscription edits, offering clearer validation feedback.
  - Added more detailed documentation and examples for query parameters in the MeterQuery.


turip (approved):

openmeter/entitlement/driver/parser.go
```diff
@@ -54,7 +54,13 @@ func (parser) ToMetered(e *entitlement.Entitlement) (*api.EntitlementMetered, er
 		CurrentUsagePeriod:     *mapPeriod(e.CurrentUsagePeriod),
 		LastReset:              metered.LastReset,
 		PreserveOverageAtReset: convert.ToPointer(metered.PreserveOverageAtReset),
-	}, nil
+	}
+
+	if e.SubscriptionManaged {
+		(*apiRes.Metadata)["om_subscription_managed"] = "true"
```
tothandras, line 60: You could use a constant for the key.

tothandras: `"true"` isn't the best value as a string, does it make sense to do something like: `om_managed_by=subscription` or something? Please also check if the map is `nil`.

tothandras (changes requested):
Please check comments

openmeter/entitlement/driver/parser.go
```diff
@@ -54,7 +54,13 @@ func (parser) ToMetered(e *entitlement.Entitlement) (*api.EntitlementMetered, er
 		CurrentUsagePeriod:     *mapPeriod(e.CurrentUsagePeriod),
 		LastReset:              metered.LastReset,
 		PreserveOverageAtReset: convert.ToPointer(metered.PreserveOverageAtReset),
-	}, nil
+	}
+
+	if e.SubscriptionManaged {
```
turip, line 59: Instead of this, cloud you validate that the metered.Metadata is not nil, clone the metered.Metadata before assembling the type and set these values on the metadata?

openmeter/entitlement/driver/parser.go
tothandras: We should probably also reserve `openmeter.` prefix not to be set by users.

turip: (E.g. notifications uses Annotations to signify that it's not an user servicable data).

GAlexIHU:
replaced by this: https://github.com/openmeterio/openmeter/pull/2529

PR #2525, opened 2025-03-24
feat(customers): get by key

hekike:
- Get customer by id or key in all `GET /api/v1/customers/:customerIdOrKey/*` endpoints
- Resolve customer in subscription HTTP Handler
- Remove CustomerNotFoundError and rely on generic

## Summary by CodeRabbit

- **New Features**
  - Customer endpoints now accept either a unique ID or a custom key, offering increased flexibility in customer lookup.

- **Enhancements**
  - Error messages have been improved for clearer, more consistent feedback when customer data isn’t found.
  - Subscription, billing, and customer processes have been streamlined with a consolidated approach to customer identification.

- **Refactor**
  - Customer input structures have been standardized across services to promote consistency and robustness.
  - Customer retrieval logic has been simplified and clarified in various methods and test cases.

### DIFF
--- a/openmeter/customer/adapter/customer.go
+++ b/openmeter/customer/adapter/customer.go
@@ -249,7 +249,9 @@ func (a *adapter) DeleteCustomer(ctx context.Context, input customer.DeleteCusto
 				}
 
 				if rows == 0 {
-					return nil, customer.NewNotFoundError(customer.CustomerID(input))
+					return nil, models.NewGenericNotFoundError(
+						fmt.Errorf("customer with id %s not found in %s namespace", input.ID, input.Namespace),
+					)
 				}
 
 				// Soft delete the customer subjects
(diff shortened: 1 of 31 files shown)


openmeter/customer/adapter/customer.go
```diff
@@ -291,14 +293,42 @@ func (a *adapter) GetCustomer(ctx context.Context, input customer.GetCustomerInp
...
+				query = query.Where(customerdb.ID(input.CustomerID.ID))
+			} else if input.CustomerKey != nil {
+				query = query.Where(customerdb.Namespace(input.CustomerKey.Namespace))
+				query = query.Where(customerdb.Key(input.CustomerKey.Key))
+			} else if input.CustomerIDOrKey != nil {
+				query = query.Where(customerdb.Namespace(input.CustomerIDOrKey.Namespace))
+				query = query.Where(customerdb.Or(
+					customerdb.ID(input.CustomerIDOrKey.IDOrKey),
+					customerdb.Key(input.CustomerIDOrKey.IDOrKey),
+				))
```
turip, line 309: Please:
-  filter for deletedAt != nil or we bring back the old bug of returning random deleted entities by key

openmeter/customer/adapter/customer.go
```diff
@@ -291,14 +293,42 @@ func (a *adapter) GetCustomer(ctx context.Context, input customer.GetCustomerInp
...
+				})
+
+			if input.CustomerID != nil {
+				query = query.Where(customerdb.Namespace(input.CustomerID.Namespace))
+				query = query.Where(customerdb.ID(input.CustomerID.ID))
+			} else if input.CustomerKey != nil {
+				query = query.Where(customerdb.Namespace(input.CustomerKey.Namespace))
+				query = query.Where(customerdb.Key(input.CustomerKey.Key))
+			} else if input.CustomerIDOrKey != nil {
+				query = query.Where(customerdb.Namespace(input.CustomerIDOrKey.Namespace))
```
turip, line 305: We should have rules for resolution order here: in case the ID and the Key matches we should always return the same entity.

openmeter/customer/customer.go
```diff
@@ -172,10 +209,32 @@ func (i DeleteCustomerInput) Validate() error {
...
+	CustomerIDOrKey *CustomerIDOrKey
+}
 
 func (i GetCustomerInput) Validate() error {
-	return CustomerID(i).Validate()
+	var errs []error
+
+	if i.CustomerID == nil && i.CustomerKey == nil && i.CustomerIDOrKey == nil {
+		return models.NewGenericValidationError(errors.New("customer id or key is required"))
+	}
```
GAlexIHU, line 223: Might be nice to do a validation error if more than 1 is provided (as thats never the intended use), + solves resolution order too (as we don't need it)

openmeter/productcatalog/subscription/http/create.go
```diff
@@ -81,11 +81,10 @@ func (h *handler) CreateSubscription() CreateSubscriptionHandler {
 					}
 				}
 
-				ref := ref.IDOrKey{
-					ID: lo.FromPtrOr(parsedBody.CustomerId, ""),
-				}
-				if ref.ID == "" {
-					ref.Key = lo.FromPtrOr(parsedBody.CustomerKey, "")
+				// Get the customer
+				customer, err := h.getCustomer(ctx, ns, parsedBody.CustomerId, parsedBody.CustomerKey)
```
GAlexIHU, line 85: i'm a bit skeptical of doing service calls mid request body parsing: it won't cause an issue now, but it would be a good assumption to uphold that body parsing has body parsing & handler has service calls...

GAlexIHU (commented):
I think it might be better if we delegate the uncertainty of IdOrKey to the caller / to solve later. We need `findByKey` & `findByID` for other reasons as well

openmeter/customer/adapter/customer.go
hekike: Added

GAlexIHU (dismissed):

turip (approved):

turip (approved):

PR #2526, opened 2025-03-25
chore: filter deleted customers

turip:
## Overview

As we are looking back to the past, but we cannot update a deleted customer.

The check is in the reconciler and not in the sync:
- during event processing we should not encounter a deleted customer
- during sync it's fine to encounter a deleted customer (as we are looking back)

## Summary by CodeRabbit

- **New Features**
  - Enhanced subscription processing by integrating customer account validation.
  - Now automatically excludes processing for subscriptions tied to inactive or deleted customer accounts, ensuring more accurate and reliable billing.

### DIFF
--- a/app/common/billing.go
+++ b/app/common/billing.go
@@ -102,11 +102,12 @@ func NewBillingCollector(logger *slog.Logger, service billing.Service) (*billing
 	})
 }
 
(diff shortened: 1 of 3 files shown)


chrisgacsal:
I'd just remove the `customer.CustomerExists()` call [here](https://github.com/openmeterio/openmeter/blob/main/openmeter/billing/service/customeroverride.go#L78-L80) and allow returning `billing.CustomerOverrideWithDetails` for even deleted Customers. And let the caller to handle the rest.

chrisgacsal (approved):

PR #2535, opened 2025-03-26
feat(api): addons spec

chrisgacsal:
## Overview

Extend Addons API by:
* adding `annotations` field to Addon
* add `mode` to AddonsRateCard

## Summary by CodeRabbit

- **New Features**
  - Introduced a comprehensive set of addon management capabilities. Users can now list, create, view, update, delete, archive, and publish addons directly via the enhanced API.
  - Enhanced addon customization with new system-managed annotations and refined rate card configurations, improving subscription add-on integration.
  - Deprecated legacy draft plan functionality to streamline operations and emphasize the new addon operations.


api/spec/src/productcatalog/addon.tsp
```diff
@@ -100,3 +107,32 @@ enum AddonOrderBy {
...
   updatedAt: "updated_at",
 }
+
+/**
+ * The addon mode defines how the ratecards included in the addon are applied in a subscription.
+ * Override mode means that the ratecards in addon take precedence over the ones which are defined in a plan with for the same feature.
+ * While the extend mode extends the ratecards from plan for the same feature.
+ */
+@friendlyName("AddonRateCardMode")
+enum AddonRateCardMode {
```
GAlexIHU, line 117: Based on the Addon spec doc, extending is a bit heuristic (works defined by the contents of that table...)

I'd suggest we add a type, maybe `conflict`, where we don't expect the customer to have the same feature (no conflict resolution)

tothandras (approved):

api/spec/src/productcatalog/addon.tsp
chrisgacsal: I'd liek to take an iterative approach, meaning I'will focus on teh crud operations first. Tehre are other missing things which needs to be added before Addons actually become fully functional. Thanks for the suggestion but I'd liek address it separately.

PR #2544, opened 2025-03-27
feat(events): entity mutation events

hekike:
- app events: create, update, delete
- app stripe events: create checkout session
- customer events: create, update, delete
- meter events: create, update, delete
- meter service layer
- feature events: create, update, archive
- plan events: create, update, delete, archive, publish
- subscription events: create, update, delete, cancel, continue
- add user id to subscription events if any
- plan json serialization

## Summary by CodeRabbit

- **New Features**
  - Expanded event-driven functionality across key operations (creation, updates, deletions) for customers, meters, products, features, plans, and subscriptions.
  - Introduced session management improvements for better user authentication context and audit trails.
  - Enhanced error handling and event publishing in service methods for apps and subscriptions.
  - Added new event types for app and meter management, enhancing event tracking capabilities.
  - Introduced new event handling for Stripe checkout sessions, including validation and metadata generation.
  - Added new `AppID` field to checkout session outputs for improved tracking.

- **Refactor**
  - Streamlined service initialization and dependency integration for enhanced system consistency and reliability.
  - Updated service configurations to incorporate event bus publishers for improved event handling.
  - Modified method signatures to return additional data (e.g., pointers to objects) for better context.

- **Tests**
  - Updated test environments with consolidated event publisher mocks to ensure stable and reliable integration.
  - Enhanced tests to verify new output structures and event handling logic.


openmeter/customer/event.go
```diff
@@ -0,0 +1,145 @@
...
+		Name:      CustomerCreateEventName,
+		Version:   "v1",
+	})
+}
+
+func (e CustomerCreateEvent) EventMetadata() metadata.EventMetadata {
+	return metadata.EventMetadata{
+		ID:     e.New.ID,
+		Source: metadata.ComposeResourcePath(e.New.Namespace, metadata.EntityCustomer, e.New.ID),
+		Time:   e.New.CreatedAt,
```
hekike, line 46: Is Subject optional?

hekike: we should add `namespace` and `user`

openmeter/customer/adapter/customer.go
```diff
@@ -284,6 +301,12 @@ func (a *adapter) DeleteCustomer(ctx context.Context, input customer.DeleteCusto
 					return nil, fmt.Errorf("failed to delete customer subjects: %w", err)
 				}
 
+				// Publish the customer deleted event
+				customerDeletedEvent := customer.NewCustomerDeleteEvent(*previousCustomer, deletedAt)
+				if err := repo.publisher.Publish(ctx, customerDeletedEvent); err != nil {
+					return nil, fmt.Errorf("failed to publish customer deleted event: %w", err)
+				}
+
```
turip, line 309: Please publish in the service layer if possible. The adapter layer is for db work.

openmeter/customer/event.go
turip: Subject is optional, but strongly encouraged. It should match the Source in this specific case. Subject will determine the partition alignment.

I only see value in `namespace` being added here if we would want to filter for it without having do deserialize the object. Otherwise, it's contained in the entity; that's how we use it. 

What does the user mean in this context? What does it hold for a service key? 

Furthermore events are also created from workers (invoice state change from billing worker), there you would not have an user attribute.

openmeter/event/metadata/entity.go
```diff
@@ -0,0 +1,70 @@
...
+	"time"
+)
+
+// EntityMutationType is the type of mutation that occurred on an entity
+type EntityMutationType string
+
+const (
+	EntityMutationTypeCreate EntityMutationType = "create"
+	EntityMutationTypeUpdate EntityMutationType = "update"
+	EntityMutationTypeDelete EntityMutationType = "delete"
```
turip, line 15: I think this is redundant with the event's type.

openmeter/event/metadata/entity.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+// EventEntityMutationPayload is a payload for an entity mutation event
+type EventEntityMutationPayload[T any] struct {
+	Entity       string             `json:"entity"`       // entity type "customer"
```
turip, line 20: This should be known from the event type. The whole event type/interface exists so that we don't need to deal with this dynamically. 

You just subscribe to the event type and know the payload's type.

Also given the event might be extended with additional fields (such as DeletedAt) I don't see the value of having the type captured here too.

openmeter/customer/event.go
```diff
@@ -0,0 +1,145 @@
...
+	return metadata.GetEventName(metadata.EventType{
+		Subsystem: CustomerEventSubsystem,
+		Name:      CustomerUpdateEventName,
+		Version:   "v1",
+	})
+}
+
+func (e CustomerUpdateEvent) EventMetadata() metadata.EventMetadata {
+	return metadata.EventMetadata{
+		ID:     metadata.GetMutationEventID(metadata.EntityMutationTypeUpdate, e.Previous.ID, e.New.UpdatedAt),
```
turip, line 87: This is autogenerated as an ULID, why do we need this to be said. Is something downstream depends on this?

https://github.com/openmeterio/openmeter/blob/3931852e80aad734048eb6fad89e05927957649d/openmeter/watermill/marshaler/marshaler.go#L103

openmeter/customer/event.go
```diff
@@ -0,0 +1,145 @@
...
+		Source: metadata.ComposeResourcePath(e.New.Namespace, metadata.EntityCustomer, e.New.ID),
+		Time:   e.New.UpdatedAt,
```
turip, line 89: nit: this defaults to now(), which is a good approximation IMHO. (But we can leave it like this)

openmeter/customer/event.go
```diff
@@ -0,0 +1,145 @@
...
+			MutationType: metadata.EntityMutationTypeCreate,
+			New:          &customer,
+		},
+	}
+}
+
+// CustomerCreateEvent is an event that is emitted when a customer is created
+type CustomerCreateEvent struct {
+	metadata.EventEntityMutationPayload[Customer] `json:",inline"`
+}
```
turip, line 32:
```suggestion
type CustomerCreateEvent struct {
	Customer `json:",inline"`
}
```

openmeter/customer/event.go
```diff
@@ -0,0 +1,145 @@
...
+			Previous:     &previous,
+		},
+		deletedAt: deletedAt,
+	}
+}
+
+// CustomerDeleteEvent is an event that is emitted when a customer is deleted
+type CustomerDeleteEvent struct {
+	metadata.EventEntityMutationPayload[Customer] `json:",inline"`
+	deletedAt                                     time.Time
```
turip, line 118:
```suggestion
	Customer `json:",inline"`
```

Customer has a deleted at field.

turip (commented):
after a bit more of thinking.

openmeter/customer/event.go
hekike: `Customer Customer`

openmeter/meter/service/manage.go
```diff
@@ -0,0 +1,223 @@
...
+// DeleteMeter deletes a meter
+func (s *ManageService) DeleteMeter(ctx context.Context, input meter.DeleteMeterInput) error {
+	// Get the meter
+	getMeter, err := s.GetMeterByIDOrSlug(ctx, meter.GetMeterInput(input))
+	if err != nil {
+		return err
+	}
+
+	// Check if the meter is already deleted
+	if getMeter.DeletedAt != nil {
```
GAlexIHU, line 93: getMeter shouldn't return deleted items

openmeter/productcatalog/feature/connector.go
```diff
@@ -137,17 +144,51 @@ func (c *featureConnector) CreateFeature(ctx context.Context, feature CreateFeat
...
+	err = c.featureRepo.ArchiveFeature(ctx, featureID)
+	if err != nil {
+		return err
+	}
+
+	// Get the archived feature
+	archivedFeature, err := c.GetFeature(ctx, featureID.Namespace, featureID.ID, true)
+	if err != nil {
+		return err
+	}
```
GAlexIHU, line 180: We should rewrite DELETEs, Archives, etc... to return the value if we need to use it (instead of refretching)

openmeter/productcatalog/plan/event.go
```diff
@@ -0,0 +1,240 @@
...
+	return PlanCreateEvent{
+		Plan:   plan,
+		UserID: session.GetSessionUserID(ctx),
+	}
+}
+
+// PlanCreateEvent is an event that is emitted when a plan is created
+type PlanCreateEvent struct {
+	Plan   *Plan   `json:"plan"`
+	UserID *string `json:"userId,omitempty"`
```
GAlexIHU, line 34: If this always comes from the context & we want it on all events we could just add to generic metadata and resolve it when publishing the event.. wdyt?

GAlexIHU (approved):

PR #2565, opened 2025-03-31
refactor: add-ons api

chrisgacsal:
## Overview

Remove `mode` from Add-ons API in the first itereation which means taht the ratecards in add-ons will always `extend` existing set coming from Plans. 

## Summary by CodeRabbit

- **Refactor**
  - Streamlined how rate card information is managed for add-on components by consolidating data structures.
  - Removed redundant representations to simplify integration and ensure a more consistent approach across API interfaces.


GAlexIHU (approved):
Just one Q as I wasn't present on that meeting: What's the path when we add back other modes (like override)?

tothandras (approved):

PR #2569, opened 2025-04-01
feat: subscription ratecard discounts

turip:
## Overview

Fixes #(issue)


GAlexIHU (approved):
As its added in subscription.RateCard, which is already used by edit & diff & patch flows, those should work fine. 

Validations also already use subscription.RateCard, so that should be fine as well

As there's no new behavior added (we just store this field), no extra code should be needed

LGTM!

PR #2576, opened 2025-04-02
feat: invoice discounts api v2

turip:
## Overview

Changes:
- Drop invoice level discounts (as we will not implement that yet)
- Move all line specific user input that is present on the rate-card of product catalog to a rate-card object (more on that later)
- Deprecate old fields that are on a rate-card
- Discounts structure is updated to show usage discounts too (non-breaking change)
- Remove charges, as they are represented as detailed lines

### Rate-card

The issue is that the term discount is overloaded; it means two things in the context of invoicing:
- The discount intent/spec (subscription terminology) (e.g. what's present on the plan and subscription)
- The actual discount applied (e.g. the amount in USD, or the usage deducted from quantity)

By separating the intent (rate-card) and the calculated value (line itself) we are future-proofing the line as we have a staging area for the intent and the calculations itself.

### Deprecating fields

The PR takes the following approach for the deprecated fields on the lines:
- For Get/List requests, both the old and the new fields are populated
- For Create/Update requests, only one of them can be non-empty, or a validation error is generated.

The line discount change is a breaking change, but it was not used before by our customers, and until usage discount is used the API will work as expected.

The invoice discounts API part is removed as is was never supported.

## Summary by CodeRabbit

- **New Features**  
  - Introduced a new, unified model for invoice charges and discounts that supports both fixed and usage-based pricing.  
  - Rolled out enhanced rate card options for flat fee and usage-based billing scenarios.

- **Bug Fixes**  
  - Improved error handling during the mapping of discounts and invoice lines.

- **Refactor**  
  - Streamlined the invoicing system by removing legacy fields and deprecating outdated properties.  
  - Enhanced validation processes for invoice-related data.

- **Tests**  
  - Added comprehensive tests to ensure robust mapping and validation of invoice billing details.


api/spec/src/billing/invoices/discounts.tsp
```diff
@@ -0,0 +1,174 @@
...
+import "@typespec/rest";
+import "@typespec/openapi3";
+
+namespace Invoices;
+
+/**
+ * InvoiceGenericCharge represents a charge or discount that can be applied to a line or the entire invoice.
+ */
+@friendlyName("InvoiceGenericCharge")
+model InvoiceGenericCharge {
```
GAlexIHU, line 11: nit: do we need the word 'Generic'?

GAlexIHU (approved):

GAlexIHU (approved):

PR #2639, opened 2025-04-09
refactor: allow single discount per type

turip:
## Overview

This patch changes the plan/subscription/invoicing discount to allow a single discount per type, by changing the discount field of the ratecard to this:

```
{
   "percentage": {
      "percentage": 10,
   }
  {
    "usage" : 10,
  }
}
```

### Ripple effect on invoicing side

The invoicing actually uses this data to calculate, so the line's ratecard discount have been altered to have the same notion.

The calculated discounts (e.g. how many units we have deducted, how much money we have discounted) on the invoice line remain an array, as there can be multiple amount or unit discounts:
- If a line has a percentage and a max spend discount multiple amount discounts might be present.

The DiscountReason is still an union type on the LineDiscount, as that's a valid use case-for that: it allows us to correlate which discount/commitment caused the specific invoice line to be created.

### DB Changes

This split of unit/amount discounts also necessitated/allowed different db tables for the two. This is beneficial as we can now properly set the null/non null flags.

The remaining field rename and table rename is to be done in a separate PR.

## Summary by CodeRabbit

- **New Features**
  - Unifies discount handling by distinguishing between usage‑based and amount‑based discounts across billing, invoicing, and subscription modules.
  - Enhances discount data integrity and clarity for improved invoicing and rate card management.

- **Refactoring**
  - Streamlines discount validations and error handling.
  - Updates database schema and API definitions to support a more consistent discount representation system.
  - Simplifies discount processing logic by directly accessing discount properties instead of using intermediate methods.


openmeter/billing/adapter/invoicelinemapper.go
```diff
@@ -160,7 +160,7 @@ func (a *adapter) mapInvoiceLineWithoutReferences(dbLine *db.BillingInvoiceLine)
 			Currency: dbLine.Currency,
 
 			TaxConfig:         lo.EmptyableToPtr(dbLine.TaxConfig),
-			RateCardDiscounts: lo.FromPtrOr(dbLine.RatecardDiscounts, nil),
+			RateCardDiscounts: lo.FromPtrOr(dbLine.RatecardDiscounts, billing.Discounts{}),
```
chrisgacsal, line 163: nit:

```suggestion
			RateCardDiscounts: lo.FromPtr(dbLine.RatecardDiscounts),
```

openmeter/billing/adapter/invoicelinemapper.go
```diff
@@ -210,71 +210,98 @@ func (a *adapter) mapInvoiceLineWithoutReferences(dbLine *db.BillingInvoiceLine)
...
+}
 
-		if dbDiscount.PreLinePeriodQuantity != nil {
-			return nil, fmt.Errorf("mapping invoice line discount[%s] failed: preLinePeriodQuantity is not nil", dbDiscount.ID)
+func (a *adapter) mapInvoiceLineAmountDiscountFromDB(dbDiscount *db.BillingInvoiceLineDiscount) (billing.AmountLineDiscountManaged, error) {
+	base := billing.LineDiscountBase{
+		Description:            dbDiscount.Description,
+		ChildUniqueReferenceID: dbDiscount.ChildUniqueReferenceID,
+		ExternalIDs: billing.LineExternalIDs{
+			Invoicing: lo.FromPtrOr(dbDiscount.InvoicingAppExternalID, ""),
```
chrisgacsal, line 276: nit

```suggestion
			Invoicing: lo.FromPtr(dbDiscount.InvoicingAppExternalID),
```

openmeter/billing/httpdriver/discounts.go
```diff
@@ -33,37 +31,34 @@ func AsUsageDiscount(d api.BillingDiscountUsage) (billing.UsageDiscount, error)
...
 
-		switch disc {
-		case string(api.BillingDiscountPercentageTypePercentage):
-			pctDiscount, err := d.AsBillingDiscountPercentage()
-			if err != nil {
-				return nil, err
-			}
+		out.Percentage = &billing.PercentageDiscount{
+			PercentageDiscount: productcataloghttp.AsPercentageDiscount(pctDiscount),
+			CorrelationID:      lo.FromPtrOr(discounts.Percentage.CorrelationId, ""),
```
chrisgacsal, line 45:
```suggestion
			CorrelationID:      lo.FromPtr(discounts.Percentage.CorrelationId),
```

openmeter/billing/httpdriver/discounts.go
```diff
@@ -33,37 +31,34 @@ func AsUsageDiscount(d api.BillingDiscountUsage) (billing.UsageDiscount, error)
...
+		if err != nil {
+			return billing.Discounts{}, err
+		}
 
-			out = append(out, billing.NewDiscountFrom(usageDiscount))
-		default:
-			return nil, fmt.Errorf("invalid discount type: %s", disc)
+		out.Usage = &billing.UsageDiscount{
+			UsageDiscount: usageDiscount,
+			CorrelationID: lo.FromPtrOr(discounts.Usage.CorrelationId, ""),
```
chrisgacsal, line 59: Same as above.

openmeter/ent/schema/billing.go
```diff
@@ -555,34 +568,38 @@ func (BillingInvoiceLineDiscount) Fields() []ent.Field {
...
-			}),
+			Deprecated("due to split of amount and usage discount tables"),
 
-		field.String("source_discount").
-			GoType(&billing.Discount{}).
-			ValueScanner(BillingDiscountValueScanner).
+		field.Other("quantity", alpacadecimal.Decimal{}).
 			SchemaType(map[string]string{
-				dialect.Postgres: "jsonb",
+				"postgres": "numeric",
```
chrisgacsal, line 590:
```suggestion
				dialect.Postgres: "numeric",
```

chrisgacsal (approved):

PR #2665, opened 2025-04-14
fix(collector): run-ai metric timestamps

tothandras:
## Summary by CodeRabbit

- **Documentation**
  - Updated configuration example to use the field name `metrics_offset` instead of `metrics_scrape_offset`.

- **Bug Fixes**
  - Adjusted metrics timestamp reporting to use the end time of the measurement period instead of the start time.

### DIFF
--- a/collector/benthos/input/run_ai.go
+++ b/collector/benthos/input/run_ai.go
@@ -96,7 +96,7 @@ input:
     app_id: "${RUNAI_APP_ID:}"
     app_secret: "${RUNAI_APP_SECRET:}"
     schedule: "*/30 * * * * *"
-    metrics_scrape_offset: "30s"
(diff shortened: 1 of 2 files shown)


hekike (approved):

collector/benthos/input/run_ai.go
```diff
@@ -96,7 +96,7 @@ input:
     app_id: "${RUNAI_APP_ID:}"
     app_secret: "${RUNAI_APP_SECRET:}"
     schedule: "*/30 * * * * *"
-    metrics_scrape_offset: "30s"
+    metrics_offset: "30s"
     resource_type: "workload"
```
jason-zhan2025, line 100: it seems metrics_offset only use in here. Is other place also need to change?

PR #2683, opened 2025-04-16
chore: refactor app api

turip:
## Overview

Fixes #(issue)


turip:
Will not happen as our typespec seem to have issues with this.

PR #2786, opened 2025-05-07
refactor(db): create RateCards table

GAlexIHU:
## Overview

As all RateCard info is stored in `jsonb` fields its really hard to change the schema


GAlexIHU:
We wont be doing this for a while as deliberated by me, @tothandras and @chrisgacsal

PR #2803, opened 2025-05-09
Fix Tracing Handler

GAlexIHU:
## Overview

The "tracing wrapper hook" is the first thing registered, but the hook itself registers everything after the server handler runs. This poses some problems:
- we used to always set tracing related attributes after the fact (not necessarily an issue but still)
- if the handler fails, **traces were missing**

As an example, an invalid ingest attempt (that fails validation) previously didnt show up in tracing, now it does
image

## Summary by CodeRabbit

- **Refactor**
  - Improved how telemetry data is collected for HTTP requests, ensuring more accurate route information is captured for monitoring and metrics. No visible changes to functionality for end-users.


tothandras (changes requested):
Let's discuss if we need route pattern at all.

chrisgacsal (approved):
Awesome! :shipit:

PR #2866, opened 2025-05-21
feat(subsadd): add timing support for migration API

GAlexIHU:
## Overview

- add `timing` support for migration API
- add `startingPhase` support for create API

## Summary by CodeRabbit

- **New Features**
	- Added an optional timing parameter to subscription migration requests, allowing users to specify when the migration should take effect (immediately, next billing cycle, or a specific date).

- **Bug Fixes**
	- Improved handling of starting phases during subscription creation and migration, ensuring correct phase selection and validation.

- **Refactor**
	- Streamlined internal logic for filtering plan phases and determining migration timing for more consistent behavior.

- **Tests**
	- Updated migration tests to cover the new timing parameter and starting phase scenarios.

### DIFF
--- a/openmeter/productcatalog/subscription/http/migrate.go
+++ b/openmeter/productcatalog/subscription/http/migrate.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/openmeterio/openmeter/api"
 	plansubscription "github.com/openmeterio/openmeter/openmeter/productcatalog/subscription"
+	"github.com/openmeterio/openmeter/openmeter/subscription"
 	"github.com/openmeterio/openmeter/pkg/framework/commonhttp"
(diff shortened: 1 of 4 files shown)


openmeter/productcatalog/subscription/service/service.go
```diff
@@ -28,3 +30,21 @@ func New(c Config) plansubscription.PlanSubscriptionService {
 		Config: c,
 	}
 }
+
+func (s *service) respectStartingPhase(p *plan.Plan, startingPhase string) error {
```
turip, line 34: Maybe:

```suggestion
func (s *service) removePhasesBeforeStartingPhase(p *plan.Plan, startingPhase string) error {
```

For me it's hard to understand what it means to respect the starting phase. If the function name is more verbose, then it becomes self-explanatory.

turip (approved):

PR #2882, opened 2025-05-23
fix(balanceworker): list entitlements params

tothandras:
Use only one where statement with the correct intent.

https://github.com/openmeterio/openmeter/blob/main/openmeter/entitlement/adapter/entitlement.go#L341-L347

## Overview

Fixes #(issue)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved filtering of deleted entitlements to include only those deleted within the last 24 hours.

### DIFF
--- a/openmeter/entitlement/balanceworker/recalculate.go
+++ b/openmeter/entitlement/balanceworker/recalculate.go
@@ -135,7 +135,6 @@ func (r *Recalculator) Recalculate(ctx context.Context, ns string) error {
 			ctx,
 			entitlement.ListEntitlementsParams{
 				Namespaces:          []string{ns},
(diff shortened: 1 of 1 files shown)


openmeter/entitlement/balanceworker/recalculate.go
```diff
@@ -135,7 +135,6 @@ func (r *Recalculator) Recalculate(ctx context.Context, ns string) error {
 			ctx,
 			entitlement.ListEntitlementsParams{
 				Namespaces:          []string{ns},
-				IncludeDeleted:      true,
```
GAlexIHU, line 138: i believe the current version is the correct intent, without includedeletedafter wont work properly

PR #2886, opened 2025-05-23
feat: implement ubp flat fee

turip:
## Overview

This patch ensures we have feature parity between a usage-based line with a flat price type and the flat fee line. This would allow us to:
- only have usage-based lines on the invoice (then it would become the Line object)
- group flat fee affecting price manipulations together
- clean up the duality of flat fees being either present as a standard line or a detailed line

By default the usage based flat fee line has one detailed line: the flat fee line that we would have added by default.

This is dead code for now. It's not in use; everywhere, we use the flat fee lines instead.

## Notes for reviewer

The following have not been altered [these will be separate prs]:
- Subscription -> Billing sync
- Invoice Update API to support setting quantity
- Create Pending Lines API to support setting quantity 
- Data migration from flat fee lines to UBP Flat lines

## Summary by CodeRabbit

- **New Features**
  - Enabled support for usage-based flat fee billing lines, including creation, validation, and detailed line management.
  - Added support for applying percentage discounts to these lines.
- **Improvements**
  - Made the feature key optional and nullable for usage-based line configurations.
  - Enhanced validation logic to distinguish between usage-based and flat fee lines based on price type.
  - Updated detailed line creation and total calculations for better accuracy.
- **Bug Fixes**
  - Corrected quantity handling and validation for flat fee and usage-based flat fee lines.
- **Tests**
  - Developed comprehensive tests for creation, discount application, and validation of usage-based flat fee lines.
  - Refined existing tests for clarity and coverage of new billing line types.
- **Chores**
  - Modified database schema to make the feature key nullable in usage-based line configurations.

### DIFF
--- a/openmeter/billing/invoiceline.go
+++ b/openmeter/billing/invoiceline.go
@@ -515,19 +515,8 @@ func (i Line) ValidateFee() error {
 		errs = append(errs, errors.New("price should be positive or zero"))
 	}
 
-	if i.Status == InvoiceLineStatusValid {
-		// Valid lines (top level invoice lines must have qty=1)
-		// Given product catalog only supports 1 as quantity, let's restrict ourselves to 1 too
-		// as the product catalog spec drives billing behavior
-		if !i.FlatFee.Quantity.Equal(alpacadecimal.NewFromInt(1)) {
-			errs = append(errs, errors.New("quantity should be 1 for invoice's flat fee lines"))
-		}
-	} else {
-		// detailed lines can have any quantity, but not negative (e.g. unit based price's qty is the
-		// usage)
-		if !i.FlatFee.Quantity.IsPositive() {
-			errs = append(errs, errors.New("quantity should be positive"))
-		}
+	if !i.FlatFee.Quantity.IsPositive() {
+		errs = append(errs, errors.New("quantity should be positive required"))
 	}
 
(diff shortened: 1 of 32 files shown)


openmeter/billing/invoiceline.go
```diff
@@ -570,6 +559,18 @@ func (i *Line) DisassociateChildren() {
...
+	if i.Type != InvoiceLineTypeUsageBased {
+		return false
+	}
+
+	if i.UsageBased.Price.Type() == productcatalog.FlatPriceType {
+		return false
+	}
+
+	return true
+}
```
GAlexIHU, line 572: I know I know very little about the scope being touched here, so this is more for my understanding:

This patch is weird for me as in that we want to support `Flat` pricing modes for usage based lines. I don't see what other attribute of "usage based" remains for a line if the price is flat. 

In the PR description you wrote this is about feature parity between a flat fee line and a usage based line with a flat price (first I thought that would be like a tiered price with a single flat tier, but i realise that would be encapsulated in line.Price so no...)

To finally spit it out, what does a usage based line with a flat price mean / represent? Its weird for me they exist

GAlexIHU: Ah okay, I got the big picture now, TY!

PR #2892, opened 2025-05-26
feat(subscription): zero phases before starting phase instead of delete

GAlexIHU:
## Overview

Currently we simply drop the phases before the starting phase, which results in a somewhat confusing user experience. To improve on it, going forward we simply set their length to zero (all contents being non-effective but still present)

## Notes for reviewer

- `sortHint` is a sad necessity. Having strictly sorted phases would mean everything would be recreated all the time.

`sortHint` implementation has some non-trivial consequences:
- When editing a subscription, if you for some reason wanted to add a phase that starts at the same as an existing phase, that will return a validationError (previously this was undefined behavior)

## Summary by CodeRabbit

- **New Features**
  - Introduced an optional "SortHint" field for subscription phases, allowing for improved sorting and ordering of phases with identical start times.
  - Enhanced API, data models, and creation/update flows to support the new "SortHint" attribute across subscription management features.
  - Added sorting and filtering capabilities for subscription phases based on the "SortHint" value.

- **Bug Fixes**
  - Improved validation and deterministic ordering of subscription phases, particularly for cases with zero-length or overlapping phases.

- **Tests**
  - Added test cases to verify correct handling and uniqueness validation for zero-length entitlements and phase sorting.

- **Chores**
  - Updated database schema and migrations to include the new "sort_hint" column for subscription phases.

### DIFF
--- a/openmeter/ent/db/migrate/schema.go
+++ b/openmeter/ent/db/migrate/schema.go
@@ -2205,6 +2205,7 @@ var (
 		{Name: "name", Type: field.TypeString},
 		{Name: "description", Type: field.TypeString, Nullable: true},
 		{Name: "active_from", Type: field.TypeTime},
+		{Name: "sort_hint", Type: field.TypeUint8, Nullable: true},
 		{Name: "subscription_id", Type: field.TypeString, SchemaType: map[string]string{"postgres": "char(26)"}},
 	}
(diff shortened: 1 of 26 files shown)


chrisgacsal (approved):

GAlexIHU:
CI got stuck somehow
image

turip (approved):

PR #2901, opened 2025-05-27
feat: migrate neked flat fee lines

turip:
## Overview

Fixes #(issue)


tools/migrate/testdata/billing/README.md
```diff
@@ -0,0 +1,76 @@
+# Billing migration checks
```
GAlexIHU, line 1: me like it

GAlexIHU: we discussed it that it would be nice to generalize this solution for future migration tests as the current way of writing raw SQL queries for both seeding & assertions is a PITA

GAlexIHU (dismissed):

PR #2922, opened 2025-05-30
feat(pc): use billing cadence in subscription sync

GAlexIHU:
## Overview

- This is a follow-up of this PR: https://github.com/openmeterio/openmeter/pull/2890
- The two PRs are separated for easier reviews
- They should be collapsed and merged together


pkg/isodate/date_test.go
```diff
@@ -150,6 +171,153 @@ func TestDivisibleBy(t *testing.T) {
...
+			expected: true,
+			hasError: false,
+		},
+		{
+			name:     "1 day not divisible by 1 hour (different units)",
+			larger:   "P1D",
+			smaller:  "PT1H",
+			expected: false,
+			hasError: false,
+		},
```
turip: Nit: I think this is fine for this PR, as most likely no one will use 1-hour durations (maybe for entitlement periods?).

But 1 hours, 12 hours etc. should be divisible by 1 day.

GAlexIHU: we agreed with @tothandras that well disable "timelike" units on input validations for now. we can revisit this later as needed

PR #2924, opened 2025-05-31
Feat/subjects

hekike:
## Overview

Fixes #(issue)


openmeter/ent/schema/subject.go
```diff
@@ -0,0 +1,65 @@
...
+		index.Fields("display_name"),
+	}
+}
+
+// Edges of the Subject.
+func (Subject) Edges() []ent.Edge {
+	return []ent.Edge{
+		// FIXME: enable foreign key constraints
+		// Ent doesn't support foreign key constraints on non ID fields (key)
+		// https://github.com/ent/ent/issues/2549
```
GAlexIHU, line 57: this is ridiculous :D

openmeter/subject/subject.go
```diff
@@ -0,0 +1,65 @@
...
+
+	if s.Key == "" {
+		errs = append(errs, errors.New("key is required"))
+	}
+
+	return errors.Join(errs...)
+}
+
+// ToAPIModel converts the subject to the API model.
+func (s Subject) ToAPIModel() api.Subject {
```
GAlexIHU, line 31: IMO it would be great if this happened on in the `http` subpackage and the api generated type wouldn't be linked to the model

openmeter/subject/service/service.go
```diff
@@ -0,0 +1,66 @@
...
+			}, nil
+		}
+	}
+
+	if subj == nil {
+		return subject.Subject{}, fmt.Errorf("subject nil")
+	}
+
+	return *subj, nil
+}
```
GAlexIHU: I think we can leave this to the entitlementHandler

openmeter/subject/service.go
```diff
@@ -0,0 +1,94 @@
...
+	if len(errs) > 0 {
+		return errors.Join(errs...)
+	}
+
+	return nil
+}
+
+// FIXME: this pattern is unique to this adapter, and should not be reused anywhere else
+// OptionalNullable is a helper to differentiate between a nil value and a value that is not set.
+type OptionalNullable[T any] struct {
```
GAlexIHU, line 92: i know this was just copied over, so its a nit: for this we use `mo.Some[T]` (or related)

openmeter/subject/adapter/adapter.go
```diff
@@ -0,0 +1,18 @@
...
+
+func New(ent *db.Client) subject.Adapter {
+	return &adapter{
+		ent: ent,
+	}
+}
+
+var _ subject.Adapter = (*adapter)(nil)
+
+type adapter struct {
```
GAlexIHU, line 16: I know it's a change request, but if we're adding it to OM, i'd make it transactional, otherwise if we'll start using it there will be surprises

PR #2930, opened 2025-06-02
feat(subs): add BillingAnchor

GAlexIHU:
## Overview

Add custom billing anchor different from subscription start

## TODO

- [x] assert phaseiterator & sync that they use correct values
- [ ] add integration test for first of month billing usecase

## Summary by CodeRabbit

- **New Features**
  - Added support for specifying a billing anchor date when creating subscriptions, allowing alignment of billing cycles to a custom anchor.
  - The billing anchor is now visible in subscription details and is normalized to the nearest recurrence before the subscription start if provided.
- **Bug Fixes**
  - Improved validation and normalization of billing anchor dates to ensure correct alignment with billing cycles.
- **Tests**
  - Expanded and refined test coverage for subscriptions with custom billing anchors and aligned billing periods.
- **Documentation**
  - Updated API documentation and schema definitions to include the new billing anchor property for subscriptions.


tools/migrate/migrations/20250602125415_subs-billing-anchor.up.sql
```diff
@@ -0,0 +1,9 @@
+-- modify "subscriptions" table
+ALTER TABLE "subscriptions" ADD COLUMN "billing_anchor" timestamptz;
+
+-- update all entries to set the billing anchor to the active_from time
+UPDATE "subscriptions" SET "billing_anchor" = "active_from" WHERE "billing_anchor" IS NULL;
```
turip: The original spec was that we align the billing anchor to the phase's start to prevent pro-rating at the beginning of the phases if the alignment mode is set to phase. 

I would recommend to add two fields:
- The alignment kind which is phase or anchor (enum)
- And an optional billing anchor if the alignment is set to anchor.

The major difference is that in phase mode the alignment is guaranteed to the phase's start so no prorating occurs at the beginning of the phase.

If the billing anchor is set to anchor, then each phase's first period will be pro-rated so that later items are aligned.

GAlexIHU: This comment is quite stale at this point, but just to reply in writing here too, phase type alignment won't be supported explicitly.

Currently, BillingAnchor is defaulted to SubscriptionStart. Using said default can result in all phase changes being prorated if for instance the subscription has a trial.

In the simple case, where phase duration is divisible by billing cadence, and the anchor is aligned to the first "paid" phase, the behavior will effectively be same as "phase aligned" billing.

chrisgacsal (approved):

PR #2949, opened 2025-06-05
feat: add store_row_id to clickhouse schema [BREAKING CHANGE]

turip:
Make sure we have a unique id per line, later on we can use this to clean up any duplicate lines in the events table as a last effort fix.

Please note that OpenMeter is not capable of migrating clickhouse (yet). After this patch you must execute the following against your events table:

```sql
ALTER TABLE "om_events" ADD COLUMN "store_row_id" String;
```

### DIFF
--- a/openmeter/streaming/clickhouse/event_query_test.go
+++ b/openmeter/streaming/clickhouse/event_query_test.go
@@ -19,7 +19,7 @@ func TestCreateEventsTable(t *testing.T) {
 				Database:        "openmeter",
 				EventsTableName: "om_events",
 			},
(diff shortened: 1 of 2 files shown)


GAlexIHU (dismissed):

openmeter/streaming/clickhouse/event_query.go
```diff
@@ -32,6 +32,7 @@ func (d createEventsTable) toSQL() string {
 	sb.Define("data", "String")
 	sb.Define("ingested_at", "DateTime")
 	sb.Define("stored_at", "DateTime")
+	sb.Define("store_row_id", "String")
```
hekike, line 35: nit: not sure we need `row` in naming just `store_id`

hekike (approved):

hekike:
Please don't forget to add BREAKING CHANGE to commit message and migration SQL.

openmeter/streaming/clickhouse/event_query.go
turip: I was intentionally not picking that: store_id can mean a lot of things that's not specific to the row itself.

PR #2960, opened 2025-06-06
feat(meter): add subject display name to csv export

hekike:
## Summary by CodeRabbit

- **New Features**
  - Introduced the ability to query meter data and download results in CSV format, including enriched subject display names when available.

- **Refactor**
  - Improved internal structure to support subject information in meter data queries.
  - Updated router configuration to integrate subject service dependency with meter handlers.

### DIFF
--- a/openmeter/subject/subject.go
+++ b/openmeter/subject/subject.go
@@ -1,9 +1,12 @@
 package subject
 
 import (
+	"context"
 	"errors"
+	"fmt"
 
 	"github.com/openmeterio/openmeter/api"
+	"github.com/openmeterio/openmeter/pkg/pagination"
 )
 
 // Subject represents a subject in the system.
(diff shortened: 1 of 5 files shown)


openmeter/subject/subject.go
```diff
@@ -62,3 +65,37 @@ func (s SubjectKey) Validate() error {
...
+func ListAll(ctx context.Context, service Service, namespace string, params ListParams) ([]Subject, error) {
+	subjects := []Subject{}
+	limit := 100
+	page := 1
+
+	for {
+		params := params
+		params.Page = pagination.NewPage(page, limit)
+
+		result, err := service.List(ctx, namespace, params)
```
turip: The service should accept List without page params and list all. We should only enforce pagination on the httphandler side.

Given subject expands no ent subresources we should be fine.

turip (dismissed):

turip (approved):

PR #2961, opened 2025-06-07
refactor: split lines

turip:
## Overview

Previously, the line object was used for 4 distinct reasons:
- Invoice level flat fee lines (now deprecated)
- Split lines put onto the gathering invoice (as grouping)
- Usage-based lines
- Detailed lines

This patch moves the split lines from the gathering invoice into a separate entity. This is required as this will allow us to:
- Have separate types for detailed and usage based lines (so that we don't have to do assertions all around the code with Status == 'valid'/'detailed' etc).
- Have dedicated table for those (so that we don't have a lot of nullable fields just because the table describes generic behavior)
- Get rid of the Parent relation that makes marshaling tricky and justifies the *Line pointer handling (e.g. we don't need to build the line tree in memory anymore).
- We can close the gap between the [API](https://github.com/openmeterio/openmeter/pull/2912) and the backend representation (as this duality is quire errorprone)

The existing tests are passing, the only test removed is the adapter test that handled split line updates.

The migration is validated using the SQLc based fixture setup to ensure that we are correctly migrating to the new setup.

## Summary by CodeRabbit

- **New Features**
  - Introduced support for split line groups in billing, enabling hierarchical grouping of usage-based invoice lines.
  - Added CRUD operations for split line groups and enhanced invoice line fetching to return individual lines or grouped hierarchies.
  - Added validation and aggregation methods for split line groups, including service period and discount handling.

- **Bug Fixes**
  - Improved validation to detect overlapping child references in split line groups.
  - Fixed handling of split lines in invoice recalculations and detailed line processing.

- **Refactor**
  - Replaced progressive billing parent/child model with split line group hierarchy.
  - Updated interfaces, services, and internal logic to use split line group IDs and hierarchies.
  - Simplified invoice line retrieval, patch generation, and test code to align with split line groups.
  - Removed obsolete split line status and related code.

- **Database Migration**
  - Migrated existing split invoice lines to split line groups, removing split line statuses and updating schema constraints.

- **Documentation**
  - Enhanced migration and test documentation to reflect split line group changes.

- **Tests**
  - Added tests for split line group migration and hierarchy validation.
  - Updated existing tests to use split line group structures.
  - Removed tests related to the old line splitting model.


openmeter/billing/invoicelinesplitgroup.go
```diff
@@ -0,0 +1,282 @@
...
+// LineOrHierarchy is a wrapper around a line or a split line hierarchy
+
+type LineOrHierarchyType string
+
+const (
+	LineOrHierarchyTypeLine      LineOrHierarchyType = "line"
+	LineOrHierarchyTypeHierarchy LineOrHierarchyType = "hierarchy"
+)
+
+type LineOrHierarchy struct {
```
hekike, line 281: I'm slightly confused why the public API is the struct and the interface the internal and not the other way around.

hekike: it could be:

```
type Line interface {
  Type()
  AsCharge()
  AsGroup()
}
```

openmeter/billing/invoicelinesplitgroup.go
```diff
@@ -0,0 +1,282 @@
...
+}
+
+func (i LineWithInvoiceHeader) Clone() LineWithInvoiceHeader {
+	return LineWithInvoiceHeader{
+		Line:    i.Line.Clone(),
+		Invoice: i.Invoice,
+	}
+}
+
+type SplitLineHierarchy struct {
```
hekike, line 193: I find hard to wrap my head around naming.

hekike (approved):

openmeter/billing/invoicelinesplitgroup.go
turip: Yes, but
a) charge is used in a completely different meaning in the codebase, it returns a line
b) it does not just return the group but it returns the lines referenced by it that's why it's called hierarchy.

PR #2964, opened 2025-06-07
refactor: split lines DDL

turip:
## Overview

Fixes #(issue)


openmeter/ent/schema/billing.go
```diff
@@ -292,13 +292,39 @@ func (m TotalsMixin) Fields() []ent.Field {
 	}
 }
 
+type InvoiceLineBaseMixin struct {
+	mixin.Schema
+}
+
+func (InvoiceLineBaseMixin) Fields() []ent.Field {
+	return []ent.Field{
+		field.Time("period_start"),
```
hekike: should we rename it to BillingPeriod/InvoicePeriod or ServicePeriod depending which one this one is?

openmeter/ent/schema/billing.go
```diff
@@ -557,6 +579,85 @@ func (BillingInvoiceLineDiscountBase) Fields() []ent.Field {
...
+		field.String("ratecard_discounts").
+			GoType(&billing.Discounts{}).
+			ValueScanner(BillingDiscountsValueScanner).
+			SchemaType(map[string]string{
+				dialect.Postgres: "jsonb",
+			}).
+			Optional().
+			Nillable(),
+
+		field.String("feature_key").
```
hekike, line 612: should this be a foreign key to a feature?
.immutable()?

openmeter/ent/schema/billing.go
```diff
@@ -557,6 +579,85 @@ func (BillingInvoiceLineDiscountBase) Fields() []ent.Field {
...
+		field.String("price").
+			GoType(&productcatalog.Price{}).
+			ValueScanner(PriceValueScanner).
+			SchemaType(map[string]string{
+				dialect.Postgres: "jsonb",
+			}),
+
+		// Subscriptions metadata
+		field.String("subscription_id").
+			Optional().
```
hekike, line 626: .Immutable()?

openmeter/ent/schema/billing.go
```diff
@@ -557,6 +579,85 @@ func (BillingInvoiceLineDiscountBase) Fields() []ent.Field {
...
+				dialect.Postgres: "jsonb",
+			}).
+			Optional().
+			Nillable(),
+
+		field.String("feature_key").
+			Optional().
+			Nillable(),
+
+		field.String("price").
```
hekike, line 617: .Immutable()?

openmeter/ent/schema/billing.go
```diff
@@ -557,6 +579,85 @@ func (BillingInvoiceLineDiscountBase) Fields() []ent.Field {
...
+func (BillingInvoiceSplitLineGroup) Mixin() []ent.Mixin {
+	return []ent.Mixin{
+		entutils.ResourceMixin{},
+		InvoiceLineBaseMixin{},
+	}
+}
+
+func (BillingInvoiceSplitLineGroup) Fields() []ent.Field {
+	return []ent.Field{
+		field.String("unique_reference_id").
```
hekike, line 599: any reason it's not just the ID?

openmeter/ent/schema/billing.go
turip: It's service period, changed. However only for the split line group as the other line types would require additional changes.

openmeter/ent/schema/billing.go
turip: Immutable added. I would love to have a foreign key to feature_id. Given that  subscription only stores it as feature key, I would not do this (plan already saves feature_id references).

We should make feature_id references in subscription too. cc @GAlexIHU

openmeter/ent/schema/billing.go
turip: For now I have added the immutable flag, but it's still a question if we allow price changes for progressively billed lines.

openmeter/ent/schema/billing.go
turip: This is used to look up specific subscription item periods, it's format is something like this:

`01JMAY1MBNCBHZ9WZ2CM7GFESQ/default/input_tokens_used/v[0]/period[0]`

This is necessary because periods are ephemeral (calculated in memory) from the subscription side, so we cannot have a single ID reference here.

This is optional, as manually added lines do not have this.

hekike (approved):

PR #3010, opened 2025-06-18
feat(app): remove default

hekike:
## Changes

- Remove the concept of default app (use default billing profile instead)
- The create billing profile API only accepts App IDs (earlier accepted app types)
- The create Stripe checkout session API fallbacks to the customer billing profile apps instead of the default Stripe app
- The app customer data API fallbacks to the customer billing profile app instead of the default app
- Option to install marketplace app without billing profile creation

## Summary by CodeRabbit

- **New Features**
  - Added option to create a billing profile during marketplace app installation.
  - Enhanced Stripe checkout session handler to resolve app IDs from billing profiles automatically.

- **Bug Fixes**
  - Improved app resolution logic for customers and billing profiles, removing dependency on default app flags.

- **Refactor**
  - Removed all "default" app properties and logic from UI, API, and backend.
  - Simplified billing profile and app creation to require explicit app IDs.
  - Updated tests to provision billing profiles using explicit app IDs.
  - Consolidated marketplace install request payload schemas for consistency.

- **Chores**
  - Updated API specifications to remove "default" properties and add new install request payload schema.

## Summary by CodeRabbit

- **New Features**
  - Added option to create a billing profile during marketplace app installation.
  - Enhanced Stripe checkout session handler to resolve app IDs from billing profiles automatically.

- **Bug Fixes**
  - Improved app resolution logic for customers and billing profiles, removing dependency on default app flags.

- **Refactor**
  - Removed all "default" app properties and related logic from UI, API, and backend.
  - Simplified billing profile and app creation to require explicit app IDs.
  - Updated tests to provision billing profiles using explicit app IDs.
  - Consolidated marketplace install request payload schemas for consistency.
  - Centralized app resolution around billing profiles instead of default app flags.

- **Chores**
  - Updated API specifications to remove "default" properties and add new install request payload schema.


openmeter/app/stripe/httpdriver/checkout_session.go
```diff
@@ -125,3 +153,76 @@ func (h *handler) CreateAppStripeCheckoutSession() CreateAppStripeCheckoutSessio
 		)...,
 	)
 }
+
+// resolveAppID resolves the app ID from the billing profile
+func (h *handler) resolveAppIDFromBillingProfile(ctx context.Context, namespace string, customerId *customer.CustomerID) (app.AppID, error) {
```
hekike, line 158: earlier this resolved to the "default" Stripe app which was always the first one that was installed if someone didn't change it via API

openmeter/app/httpdriver/customer.go
```diff
@@ -271,63 +273,79 @@ func (h *handler) getCustomerData(ctx context.Context, namespace string, apiApp
...
 		// Parse as custom invoicing app
 		apiCustomInvoicingCustomerData, err := apiApp.AsCustomInvoicingCustomerAppData()
 		if err != nil {
 			return nil, nil, fmt.Errorf("error converting to custom invoicing app: %w", err)
 		}
 
-		// Get app ID from API data or get default app
-		app, err := h.getApp(ctx, namespace, apiCustomInvoicingCustomerData.Id, app.AppTypeCustomInvoicing)
+		// Resolve app
+		resolvedApp, err := h.resolveCustomerApp(ctx, customerID, app.AppTypeCustomInvoicing, apiCustomInvoicingCustomerData.Id)
```
hekike, line 297: Earlier this resolved to the default Stripe app which was always the first one installed except someone didn't change it explicitly via API.

api/spec/src/app/marketplace.tsp
```diff
@@ -116,6 +116,25 @@ model MarketplaceOAuth2InstallAuthorizeRequest {
...
 
+/**
+ * Marketplace install request payload.
+ */
+@friendlyName("MarketplaceInstallRequestPayload")
+model MarketplaceInstallRequestPayload {
+  /**
+   * Name of the application to install.
+   *
+   * If not set defaults to the marketplace item's description.
```
GAlexIHU: without context, "item description" feels weird, but then i suppose it doesn't have a name prop, right?

api/spec/src/app/marketplace.tsp
```diff
@@ -116,6 +116,25 @@ model MarketplaceOAuth2InstallAuthorizeRequest {
...
+   */
+  name?: string;
+
+  /**
+   * If true, a billing profile will be created for the app.
+   * The Stripe app will be also set as the default billing profile if the current default is a Sandbox app.
```
GAlexIHU, line 133: Can't we do this separately? I mean from API i wouldn't necessarily expect this to change the default... then on web UI, we could have a pop-up for manual confirmation. wdyt?

openmeter/app/httpdriver/customer.go
```diff
@@ -271,63 +273,79 @@ func (h *handler) getCustomerData(ctx context.Context, namespace string, apiApp
...
+		Customer: customerID,
 	})
 	if err != nil {
-		return nil, fmt.Errorf("error getting default %s app: %w", appType, err)
+		return nil, fmt.Errorf("error getting default profile: %w", err)
+	}
+
+	mergedProfile := customerOverride.MergedProfile
+
+	// If one of them uses the app type, anothers shouldn't use the same app type with different ID
```
GAlexIHU, line 336: can we instead check this please?

openmeter/app/sandbox/helpers.go
```diff
@@ -35,31 +37,34 @@ func AutoProvision(ctx context.Context, input AutoProvisionInput) (app.App, erro
...
-				return nil, fmt.Errorf("cannot create sandbox app: %w", err)
-			}
+		return nil, fmt.Errorf("cannot list apps: %w", err)
+	}
 
-			return input.AppService.GetDefaultApp(ctx, app.GetDefaultAppInput{
-				Namespace: input.Namespace,
-				Type:      app.AppTypeSandbox,
-			})
+	// If there is no sandbox app, we need to provision a new one
```
GAlexIHU, line 49: nit/q: is there a point to having more than one sandbox apps? If so, we should probably be stricter with the check & see if "our" sandbox app exists (and if not create it) maybe via annotations or stg...

openmeter/app/stripe/entity/input.go
```diff
@@ -333,29 +330,11 @@ func (i CreateCheckoutSessionInput) Validate() error {
 			return errors.New("app and customer must be in the same namespace")
 		}
 
-		if i.CustomerKey != nil {
```
GAlexIHU, line 336: why did you get rid of customerkey support?

openmeter/app/stripe/httpdriver/checkout_session.go
```diff
@@ -125,3 +153,76 @@ func (h *handler) CreateAppStripeCheckoutSession() CreateAppStripeCheckoutSessio
...
+			appID = profile.Apps.Payment.GetID()
+			stripeApps = append(stripeApps, profile.Apps.Payment)
+
+			if profile.Default {
+				foundDefault = true
+			}
+		}
+	}
+
+	// If no default profile is found, check if there is only one stripe app and use it
```
GAlexIHU: (why) do we want this?

openmeter/ent/schema/app.go
```diff
@@ -27,18 +27,12 @@ func (App) Fields() []ent.Field {
...
 		field.String("type").GoType(appentitybase.AppType("")).Immutable(),
 		field.String("status").GoType(appentitybase.AppStatus("")),
-		field.Bool("is_default").Default(false),
 	}
 }
 
 func (App) Indexes() []ent.Index {
 	return []ent.Index{
 		index.Fields("namespace", "type"),
-		index.Fields("namespace", "type", "is_default").
```
GAlexIHU, line 37: now that we've learned it, should we use "deprecates" instead of deletes? (so the field is not deleted bot not used anymore... wdyt?

openmeter/app/stripe/entity/input.go
hekike: It's resolved in HTTP handler now.

api/spec/src/app/marketplace.tsp
hekike: Currently it's always enabled as there was no flag for it. That's why the new flag default=true.

openmeter/ent/schema/app.go
chrisgacsal: I think it is ok to drop the index, but would deprecate the _is_default_ field instead of dropping it.

openmeter/app/httpdriver/customer.go
hekike: It's already checked here, which is actually even more strict:
https://github.com/openmeterio/openmeter/blob/c50be79e80079adaad9b950117d8bc6e455f1d75/openmeter/billing/service/profile.go#L55

openmeter/app/sandbox/helpers.go
hekike: I can see someone wanting to test customer profile overrides with multiple sandbox apps

openmeter/app/stripe/httpdriver/checkout_session.go
hekike: comment is wrong, fixed it

openmeter/ent/schema/app.go
hekike: I would like to avoid Ent deprecates as it keeps getter/setter, just setter does something while getter doesn't. I think it's confusing.

api/spec/src/app/marketplace.tsp
hekike: Updated

openmeter/ent/schema/app.go
GAlexIHU: > I think it is ok to drop the index, but would deprecate the _is_default_ field instead of dropping it.

sry yeah thats what i meant

GAlexIHU: > I would like to avoid Ent deprecates as it keeps getter/setter, just setter does something while getter doesn't. I think it's confusing.

im ok with that 👍

GAlexIHU (approved):

PR #3053, opened 2025-06-30
refactor(streaming): rm meter management

chrisgacsal:
## Overview

Remove `meter` management from `streaming` as it became unused since meters are not backed by materialized view tables in Clickhouse anymore.

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -203,15 +203,6 @@ func main() {
 		})
 	})
 
-	for _, meter := range conf.Meters {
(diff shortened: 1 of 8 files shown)


cmd/server/main.go
```diff
@@ -203,15 +203,6 @@ func main() {
 		})
 	})
 
-	for _, meter := range conf.Meters {
```
GAlexIHU, line 206: don't we need this somewhere?

GAlexIHU (approved):
we still create the meters in pg, right?

PR #3073, opened 2025-07-07
feat(billing): ignore sync annotation

GAlexIHU:
## Overview

Adds ignore annotation to billing (will be used for period fix migration)

Fixes OM-1239

## Summary by CodeRabbit

* **New Features**
  * Added support for storing and managing annotations on billing invoice lines, including a new annotations field that accepts arbitrary key-value data.
  * Introduced the ability to mark invoice lines to be ignored during subscription synchronization using a specific annotation.
  * Enhanced querying and updating capabilities for invoice line annotations.

* **Bug Fixes**
  * Improved handling of zero-length service periods to ensure accurate intersection logic.

* **Tests**
  * Added tests to verify correct handling of ignored invoice lines and validation of period intersection logic.

* **Database**
  * Added a new JSONB column and GIN index for annotations on invoice lines, with corresponding migration scripts.

### DIFF
--- a/openmeter/billing/worker/subscription/sync.go
+++ b/openmeter/billing/worker/subscription/sync.go
@@ -629,6 +629,15 @@ func (h *Handler) getPatchesForExistingLineOrHierarchy(existingLine billing.Line
 }
 
 func (h *Handler) getPatchesForExistingLine(existingLine *billing.Line, expectedLine *billing.Line, invoiceByID InvoiceByID) ([]linePatch, error) {
+	// Lines can be manually marked as ignored in syncing, which is used for cases where we're doing backwards incompatible changes
+	if ignore, ok := expectedLine.Annotations[billing.AnnotationSubscriptionSyncIgnore]; ok && ignore == true {
+		return nil, nil
+	}
+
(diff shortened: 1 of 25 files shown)


openmeter/billing/worker/subscription/sync.go
```diff
@@ -629,6 +629,15 @@ func (h *Handler) getPatchesForExistingLineOrHierarchy(existingLine billing.Line
 }
 
 func (h *Handler) getPatchesForExistingLine(existingLine *billing.Line, expectedLine *billing.Line, invoiceByID InvoiceByID) ([]linePatch, error) {
+	// Lines can be manually marked as ignored in syncing, which is used for cases where we're doing backwards incompatible changes
+	if ignore, ok := expectedLine.Annotations[billing.AnnotationSubscriptionSyncIgnore]; ok && ignore == true {
+		return nil, nil
+	}
```
turip, line 635: We should be careful using this, as the line in question might have not yet been synchronized to billing, thus those lines will never get synced.

turip (approved):
Thank you!

openmeter/billing/worker/subscription/sync.go
turip: never mind, this is for existing lines only.

PR #3119, opened 2025-07-18
chore: fix metadata test

turip:
## Overview

This way we don't need to mess with sorting.

## Summary by CodeRabbit

* **Tests**
  * Updated a test case to simplify input data and expected output for metadata serialization.


chrisgacsal:
@turip I have created a proper fix for this: https://github.com/openmeterio/openmeter/pull/3120.

PR #3149, opened 2025-07-28
feat: billing use streaming specified minimum resolution [OM-1469]

turip:
## Overview

This patch ensures that the whole billing stack uses 1s resolution for both usage_based/metered and usage_based/flat_fee lines.

This is required as due to collection moved to the draft invoices if an invoice contains both types, we might create two invoices.

The overal approach is that all inputs to billing is sanitized to use per second resolution entries, plus a migration is added so that subscription sync will fill the gaps between the lines of the same subscription item. This is done at multiple places:
- LineService ensures that CreatePendingLines truncate the periods.
- HTTPDriver ensures that Simulation and Invoice Update truncates the period.

We are not updating the old invoices to prevent changes in already issued invoices.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy and consistency of time period handling across billing and invoicing by standardizing truncation to a 1-second resolution.
  * Enhanced validation and comparison of invoice line periods to prevent issues with zero-length or misaligned service periods.
  * Updated test cases to reflect new time truncation logic and ensure reliable results.

* **Chores**
  * Renamed internal constants for minimum window size duration to improve clarity.
  * Added a database migration to annotate certain invoice lines for synchronization purposes.

* **Documentation**
  * Added and updated comments to clarify billing line handling and truncation behavior.

### DIFF
--- a/openmeter/billing/httpdriver/invoiceline.go
+++ b/openmeter/billing/httpdriver/invoiceline.go
@@ -13,6 +13,7 @@ import (
 	"github.com/openmeterio/openmeter/openmeter/customer"
 	"github.com/openmeterio/openmeter/openmeter/productcatalog"
 	productcataloghttp "github.com/openmeterio/openmeter/openmeter/productcatalog/http"
+	"github.com/openmeterio/openmeter/openmeter/streaming"
 	"github.com/openmeterio/openmeter/pkg/clock"
 	"github.com/openmeterio/openmeter/pkg/convert"
 	"github.com/openmeterio/openmeter/pkg/currencyx"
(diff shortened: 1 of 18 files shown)


openmeter/billing/httpdriver/invoiceline.go
```diff
@@ -557,11 +558,11 @@ func mapSimulationLineToEntity(line api.InvoiceSimulationLine) (*billing.Line, e
 
 			Status: billing.InvoiceLineStatusValid,
 			Period: billing.Period{
-				Start: line.Period.From,
-				End:   line.Period.To,
+				Start: line.Period.From.Truncate(streaming.MinWindowSizeDuration),
```
hekike: Should we rename `MinWindowSizeDuration` to `MinimumWindowSizeDuration`? Took me a while to realize it doesn't mean minute.

turip: Sure

tothandras (approved):

PR #3193, opened 2025-08-07
fix(ent): make expiration count uint32 from uint8

GAlexIHU:
There was no hard reason for `expiration.count` to be a `uint8`, and it overflows for specific use cases like 365 days.

## Summary by CodeRabbit

* **New Features**
  * The expiration period count now supports values between 1 and 1000, allowing for greater flexibility in setting expiration durations.

* **Bug Fixes**
  * Improved validation ensures that the expiration count is always within the allowed range, preventing invalid configurations.

* **Refactor**
  * Updated internal handling to support larger expiration period counts for enhanced scalability.


hekike (approved):

hekike (changes requested):
how about `uint` instead of `int`? It cannot be negative.

hekike (approved):
how about `uint` instead of `int`? It cannot be negative.

PR #3223, opened 2025-08-13
fix: adds a config option to enable TLS/SSL without needing certs loc…

mleonidas:
…ally (insecure)

## Overview

Fixes #(issue)

## Summary by CodeRabbit

- New Features
  - Added an option to allow insecure TLS for Kafka SSL connections (e.g., self-signed certificates). This enables TLS without SASL when explicitly opted in.
  - Expands connectivity to Kafka deployments using SSL with untrusted/temporary certificates while keeping current defaults.
  - Existing behavior remains unchanged unless the option is enabled; SASL_SSL handling is unaffected.

### DIFF
--- a/app/config/ingest.go
+++ b/app/config/ingest.go
@@ -61,6 +61,7 @@ func (c KafkaIngestConfiguration) Validate() error {
 type KafkaConfiguration struct {
 	Broker           string
 	SecurityProtocol string
+	TLSInsecure      bool
 	SaslMechanisms   string
 	SaslUsername     string
 	SaslPassword     string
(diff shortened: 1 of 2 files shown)


mleonidas:
Notably, I encountered this issue while attempting to deploy to an internal Kubernetes cluster in AWS, utilizing MSK as the Kafka cluster. We have our cluster configured for TLS, but due to the certs not being locally or in the container, the brokers would fail to connect. 


edit: also sorry about the long commit msg, I've never used `cz` cli, I have a git template that neogit reads in, but the commit hooks would fail until I used `cz` :/

turip:
@mleonidas as a sidenote, you can specify the required CAs here: https://github.com/openmeterio/openmeter/blob/main/deploy/charts/openmeter/templates/deployment.yaml#L69-L73

PR #3224, opened 2025-08-13
fix: period calc fix

turip:
## Overview

When doing the period fix we have not implemented the recurrence calculation logic besides the date overflow calculation.

This patch addresses that gap.


pkg/timeutil/compare.go
```diff
@@ -5,3 +5,10 @@ import "time"
 func Compare(a, b time.Time) int {
 	return int(a.Sub(b))
 }
+
+func Later(t1 time.Time, t2 time.Time) time.Time {
```
GAlexIHU, line 9: lo.Latest is fine

GAlexIHU (approved):

PR #3319, opened 2025-09-07
fix(customer): check if customer has entitlements

hekike:
Do not allow to delete customers with active entitlements.


tothandras (changes requested):
https://github.com/openmeterio/openmeter/blob/main/openmeter/subscription/validators/customer/validator.go#L82
Please add the check here

tothandras (approved):

PR #3394, opened 2025-09-16
feat(collector): add configurable pending pod inclusion for kubernetes input

tothandras:


collector/benthos/input/kubernetes.go
```diff
@@ -192,9 +202,35 @@ func (in *kubernetesResourcesInput) ReadBatch(ctx context.Context) (service.Mess
...
-				}) {
+				// Filter pods based on their status and configuration.
+				shouldInclude := false
+
+				switch pod.Status.Phase {
+				case corev1.PodRunning:
+					shouldInclude = true
+				case corev1.PodPending:
+					shouldInclude = in.includePendingPods
+				// Phase not set, check container statuses
```
hekike, line 214: please extend comment to explain why `""` phase is possible

tothandras: It shouldn't be, but the documentation states:

> Most recently observed status of the pod. This data may not be up to date. Populated by the system.

I've also added logging if this happens.

PR #3442, opened 2025-09-29
refactor(subs): subscription validations

GAlexIHU:
> Part II of Multi-Subscription support

## Overview

**Does not change** any functionality, just prep work for later changes

- Standardize subscription service & adapter methods
- Standardize view building
- Renames SubscriptionValidator interfaces
- Moves uniqueness validations inside a new validator service
  - for that to work breaks up Service interface in two
- Rewrites uniqueness validation so its more composable

## Summary by CodeRabbit

- **New Features**
  - Expanded subscription views, batch entitlement/item retrieval, and a new unique-subscription validation flow to detect overlapping subscriptions.

- **API Changes**
  - Subscription list filter renamed Customers → CustomerIDs.
  - Active-in-period semantics now use a start-bounded period with validation.
  - Service constructors now return initialization errors; validation interfaces updated.

- **Performance/Reliability**
  - Listing and restore use pagination and a standardized time source for more consistent, scalable queries.


openmeter/subscription/validators/subscription/components.go
```diff
@@ -6,84 +6,82 @@ import (
...
+			return okResult(append(specs, spec))
 		},
 	)
 }
 
-func (v SubscriptionUniqueConstraintValidator) includeSubViewUnique(view subscription.SubscriptionView) func(mo.Result[[]subscription.SubscriptionView]) mo.Result[[]subscription.SubscriptionView] {
-	return result.FlatMap(
-		func(views []subscription.SubscriptionView) mo.Result[[]subscription.SubscriptionView] {
-			return mo.Ok(lo.UniqBy(append(views, view), func(i subscription.SubscriptionView) string {
+func (v SubscriptionUniqueConstraintValidator) includeSubViewUnique(view subscription.SubscriptionView) func(result[[]subscription.SubscriptionView]) result[[]subscription.SubscriptionView] {
```
turip: please refactor it so that it returns a proper []subscription.SubscriptionView], error

openmeter/subscription/validators/subscription/validator.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (v SubscriptionUniqueConstraintValidator) ValidateContinued(ctx context.Context, view subscription.SubscriptionView) error {
+	return v.pipelineAfter(ctx, view).Error()
+}
+
+func (v SubscriptionUniqueConstraintValidator) ValidateDeleted(ctx context.Context, view subscription.SubscriptionView) error {
+	return v.pipelineAfter(ctx, view).Error()
+}
+
+func (v SubscriptionUniqueConstraintValidator) pipelineAfter(ctx context.Context, view subscription.SubscriptionView) result[[]subscription.SubscriptionSpec] {
```
turip: Please remove result and pipeline and try to refactor it in a way, that it's easy to read. e.g. in this example it would be super nice to return just the right type instead of just piping maps just so that the original fuction doesn't have to return the proper return values.

This is especially true given that all of the collectSubs call is followed by a mapSubsToView.

GAlexIHU:
Regarding the two comments @turip

I recognize we generally don't advise using monads in our codebase, and that there is an effort getting rid of `samber/mo`. I'm perfectly fine getting rid of `samber/mo`, but I think there is a strong argument for using `pipeline` style abstractions in this specific case (scoping them package local).

Currently there are 5 post validations and 2 pre validations. The post validations will likely look identical to each other and their current form going forward. In the follow up PRs, the 2 pre validations will be expanded for all 5 methods, and they will include small variations, similar to how there are currently small variations between `ValidateCreate`, `ValidateContinue` and `pipelineAfter`. 

The benefit of using a pipeline-like abstraction, combined with good naming, is it's brevity. When having six almost identical execution flow, it's really easy to see the difference between them due to how short the code segment is. Just try to compare the two examples below (following our otherwise lint rules) and see which is easier for you to parse ALL differences between them.

Example 1
```go
func (v SubscriptionUniqueConstraintValidator) ValidateCreate(ctx context.Context, namespace string, spec subscription.SubscriptionSpec) error {
	return pipe5(
		v.collectSubs(ctx, namespace, spec.CustomerId, spec.ActiveFrom),
		v.mapSubsToViews(ctx),
		v.mapViewsToSpecs(),
		v.validateUniqueConstraint(ctx, func(err error) error {
			return errors.New("inconsistency error: already scheduled subscriptions are overlapping")
		}),
		v.includeSubSpec(spec),
		v.validateUniqueConstraint(ctx, nil),
	).Error()
}

func (v SubscriptionUniqueConstraintValidator) ValidateContinue(ctx context.Context, view subscription.SubscriptionView) error {
	// We're only validatint that the subscription can be continued indefinitely
	spec := view.AsSpec()
	spec.ActiveTo = nil

	return pipe6(
		v.collectSubs(ctx, view.Customer.Namespace, view.Customer.ID, view.Subscription.ActiveFrom),
		v.mapSubsToViews(ctx),
		v.filterSubViews(func(v subscription.SubscriptionView) bool {
			return v.Subscription.ID != view.Subscription.ID
		}),
		v.mapViewsToSpecs(),
		v.validateUniqueConstraint(ctx, func(err error) error {
			return errors.New("inconsistency error: already scheduled subscriptions are overlapping")
		}),
		v.includeSubSpec(spec),
		v.validateUniqueConstraint(ctx, nil),
	).Error()
}
```

Example 2
```go
func (v SubscriptionUniqueConstraintValidator) ValidateCreateLong(ctx context.Context, namespace string, spec subscription.SubscriptionSpec) error {
	subs, err := v.collectCustomerSubscriptionsStarting(ctx, namespace, spec.CustomerId, spec.ActiveFrom)
	if err != nil {
		return err
	}

	views, err := v.Config.QueryService.ExpandViews(ctx, subs)
	if err != nil {
		return err
	}

	specs := slicesx.Map(views, func(v subscription.SubscriptionView) subscription.SubscriptionSpec {
		return v.AsSpec()
	})

	specs, err = v.validateUniqueConstraintNonMonadic(ctx, specs)
	if err != nil {
		return errors.New("inconsistency error: already scheduled subscriptions are overlapping")
	}

	specs = append(specs, spec)

	_, err = v.validateUniqueConstraintNonMonadic(ctx, specs)
	if err != nil {
		return err
	}

	return nil
}

func (v SubscriptionUniqueConstraintValidator) ValidateContinueLong(ctx context.Context, view subscription.SubscriptionView) error {
	// We're only validatint that the subscription can be continued indefinitely
	spec := view.AsSpec()
	spec.ActiveTo = nil

	subs, err := v.collectCustomerSubscriptionsStarting(ctx, view.Customer.Namespace, view.Customer.ID, view.Subscription.ActiveFrom)
	if err != nil {
		return err
	}

	views, err := v.Config.QueryService.ExpandViews(ctx, subs)
	if err != nil {
		return err
	}

	views = lo.Filter(views, func(v subscription.SubscriptionView, _ int) bool {
		return v.Subscription.ID != view.Subscription.ID
	})

	specs := slicesx.Map(views, func(v subscription.SubscriptionView) subscription.SubscriptionSpec {
		return v.AsSpec()
	})

	specs, err = v.validateUniqueConstraintNonMonadic(ctx, specs)
	if err != nil {
		return errors.New("inconsistency error: already scheduled subscriptions are overlapping")
	}

	specs = append(specs, spec)

	_, err = v.validateUniqueConstraintNonMonadic(ctx, specs)
	if err != nil {
		return err
	}

	return nil
}
```

Of course you're familiar with this exact case as it's already in this PR, but I think it's fair to assume of me that the first one was easier to parse, as its an effective 6-8 lines per method versus 30. The second two doesn't even fit on screen at the same time. If you have multiple of these with small variations I think brevity is really beneficial to quickly understand what the differences are.

To address a different point, discussed over on slack, that _in the future (or even now) we might have contributors who don't feel comfortable or familiar seeing functional patterns_, I don't find it realistic that we'd have contributors who'd get lost at the sight of this, and even if it might not be trivial at first sight to everyone what a `pipeline` does, I think it's visible by my example that writing it this way it's a lot easier to see what the `validation` does (which would be enough in most cases).

These will only get more complex as multi-subscription finishes, and I did put effort into figuring out what would be the best way to structure these to make them the easiest to understand so I have a fairly strong stance on this one. All rules and conventions have exceptions, I think this exception brings value.

turip:
I would prefer the example 2: for me that's more readable and easier to understand. If others do not have objections, feel free to merge it.

@chrisgacsal / @tothandras Please review this PR if you will.

chrisgacsal:
> I would prefer the example 2: for me that's more readable and easier to understand. If others do not have objections, feel free to merge it.
> 
> @chrisgacsal / @tothandras Please review this PR if you will.

Same here.

openmeter/subscription/validators/subscription/components.go
```diff
@@ -0,0 +1,62 @@
...
+// This file contains components from which the validation pipelines are built
+func (v SubscriptionUniqueConstraintValidator) validateUniqueConstraint(ctx context.Context, specs []subscription.SubscriptionSpec) ([]subscription.SubscriptionSpec, error) {
+	switch multiSubscriptionEnabled, err := v.Config.FeatureFlags.IsFeatureEnabled(ctx, subscription.MultiSubscriptionEnabledFF); {
+	case err != nil:
+		return nil, fmt.Errorf("failed to check if multi-subscription is enabled: %w", err)
+	case multiSubscriptionEnabled:
+		return specs, subscription.ValidateUniqueConstraintByFeatures(specs)
+	default:
+		return specs, subscription.ValidateUniqueConstraintBySubscriptions(specs)
+	}
```
chrisgacsal:
```suggestion
    multiSubscriptionEnabled, err := v.Config.FeatureFlags.IsFeatureEnabled(ctx,subscription.MultiSubscriptionEnabledFF)
    if err != nil:
		return nil, fmt.Errorf("failed to check if multi-subscription is enabled: %w", err)
	}
		
	if  multiSubscriptionEnabled {
		return specs, subscription.ValidateUniqueConstraintByFeatures(specs)
	}

	return specs, subscription.ValidateUniqueConstraintBySubscriptions(specs)

```

chrisgacsal (changes requested):
LGTM with one nit.

chrisgacsal (approved):

PR #3470, opened 2025-10-06
feat(database): add new advanced filter group by column to features

tothandras:
## Summary by CodeRabbit

- New Features
  - Added advanced meter group-by filters for Features.
  - API now supports setting/clearing advanced group-by filters; reads prefer the advanced field and fall back to existing filters when absent.
- Chores
  - Database migration adds an advanced_meter_group_by_filters column and auto-populates it from existing meter_group_by_filters where applicable.
- Tests
  - Added migration tests covering null, empty, single, and multi-filter scenarios to validate backfill and compatibility.


tools/migrate/migrations/20251006141236_feature-advanced-meter-group-by-filters.up.sql
```diff
@@ -0,0 +1,19 @@
...
+SET "advanced_meter_group_by_filters" = (
+    SELECT COALESCE(
+        jsonb_object_agg(
+            key,
+            jsonb_build_object('$eq', value)
+        ),
+        '{}'::jsonb
+    )
+    FROM jsonb_each_text("meter_group_by_filters") AS kv(key, value)
+)
```
GAlexIHU, line 16: based on our recent bad experiences, let's at least try to add a test for it. Also, if we want it to be part of the automatic migration (without timeouts), this would have to run really fast (which it most likely wont), so this would need to be a manual deploy

PR #3478, opened 2025-10-07
feat(api): add meter group by values endpoint

tothandras:
## Summary by CodeRabbit

- **New Features**
  - New GET endpoint to list distinct meter group-by values with optional RFC3339 from/to and search filters (defaults to last 24h).

- **SDKs**
  - Go and JavaScript clients updated with request builders, parameter types, and string[] responses.

- **Backend**
  - Routing, handler, streaming connector, query builder, and validation added to support the endpoint and enforce time/window rules.

- **Tests**
  - Unit tests and mock implementations added to cover the new endpoint.


openmeter/streaming/clickhouse/group_by_values_query.go
```diff
@@ -0,0 +1,52 @@
...
+func (d listGroupByValuesQuery) toSQL() (string, []interface{}) {
+	tableName := getTableName(d.Database, d.EventsTableName)
+
+	// Helper function to get the column
+	getColumn := columnFactory(d.EventsTableName)
+
+	groupByJSONPath := d.Meter.GroupBy[d.GroupByKey]
+
+	sb := sqlbuilder.ClickHouse.NewSelectBuilder()
+	sb.Select(sb.As(fmt.Sprintf("DISTINCT JSON_VALUE(%s, '%s')", getColumn("data"), escapeJSONPathLiteral(groupByJSONPath)), "group_by_values"))
```
hekike: would `query.Var(groupByJSONPath)` work here to move to arg?

openmeter/streaming/connector.go
```diff
@@ -43,6 +43,8 @@ type Connector interface {
 	ListEventsV2(ctx context.Context, params ListEventsV2Params) ([]RawEvent, error)
 	// ListSubjects lists the subjects that have events in the database
 	ListSubjects(ctx context.Context, params ListSubjectsParams) ([]string, error)
+	// ListGroupByValues lists the group by values that have events in the database
+	ListGroupByValues(ctx context.Context, params ListGroupByValuesParams) ([]string, error)
```
hekike, line 47: could/should this replace `ListSubjects`? Looks like it does the same. Or maybe keep on interface just make it call into `ListGroupByValues`.

openmeter/meter/httphandler/query.go
```diff
@@ -205,3 +205,74 @@ func (h *handler) QueryMeterPost() QueryMeterPostHandler {
...
+			ns, err := h.resolveNamespace(ctx)
+			if err != nil {
+				return ListGroupByValuesRequest{}, err
+			}
+
+			return ListGroupByValuesRequest{
+				namespace:  ns,
+				idOrSlug:   params.IdOrSlug,
+				groupByKey: params.GroupByKey,
+				from:       params.From,
```
hekike, line 251: let's add a validation that from-to duration cannot be more than 90 days or that from must be younger than 90 days so we don't scan more than 3 partitions

openmeter/streaming/clickhouse/group_by_values_query.go
tothandras: In `Select`, yes. I'll move it.

openmeter/streaming/connector.go
tothandras: For ListSubjects meter is not required. We can consolidate it later.

PR #3840, opened 2026-02-03
feat(ledger): write POC code for primitive APIs

GAlexIHU:
## Overview

This is a conversation starter

Fixes #(issue)


openmeter/ledger/primitives.go
```diff
@@ -0,0 +1,124 @@
...
+
+type Ledger interface {
+	// GetAccount retreives an account from the Ledger
+	GetAccount(ctx context.Context, address Address) (Account, error)
+
+	// SetUpTransaction creates a new transaction on the Ledger and returns it without committing it
+	SetUpTransaction(ctx context.Context, at time.Time, entries []LedgerEntryInput) (Transaction, error)
+
+	// CommitGroup commits a list of transactions on the Ledger atomically
+	CommitGroup(ctx context.Context, group TransactionGroup) error
```
GAlexIHU: It could be argued that if we commit in groups, then setting up the transactions doesn't actually need the ledger (i.e. could be stg like a free-floating builder). I think it'd still be nice if it was on the ledger, as it lets as easily validate against the ledger at setup time -- even though we'd validate again at commit, it'd probably be better to have errors returned at setup to not overload the handling of the commit

openmeter/ledger/primitives.go
```diff
@@ -0,0 +1,124 @@
...
+
+	Key() string // can be typed but str should be fine
+	Value() TValue
+}
+
+// ----------------------------------------------------------------------------
+// Transaction primitives
+// ----------------------------------------------------------------------------
+
+type EntryType string
```
GAlexIHU: maybe just use +/- ("+" = debit, "-" = credit), it would be a better DB format for aggregations e.g.

PR #4119, opened 2026-04-10
fix(subs-sync): startup when credits are disabled

GAlexIHU:
## Overview

To make sure if credits_only config leaks into subscriptions in a disabled environment we still don't error on sync

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced handling of credit-only subscriptions when the charges service is disabled. The system now gracefully processes these subscriptions by providing generation time limits instead of returning errors, while optimizing reconciliation by skipping unnecessary patch-planning operations to improve billing operation stability.


GAlexIHU:
we won't be this paranoid

PR #4128, opened 2026-04-10
fix: adjust flat-fee shape to match usage-based

turip:
## Overview

Unify external data structure format between usage-based and flat fee charges.

## Summary by CodeRabbit

* **New Features**
  * Flat-fee charges now persist a dedicated, more detailed status and surface realized data (payments, credits, accrued usage) separately so readers see committed realizations explicitly.

* **Refactor**
  * Charge representation split into a durable base row and separate realization data, and services now update base vs. realization data independently to avoid redundant writes and clarify lifecycle behavior.


GAlexIHU (dismissed):
thx for doing this!

GAlexIHU (approved):

PR #4147, opened 2026-04-15
feat(ledger): transaction listing API

GAlexIHU:
## Overview

- extra annotations on transactions (so we can filter by them)
- page based transaction listing API (long-term goal is to change public API to be cursor based, will likely be an increment after discussions)
- add cursored point-in-time balance evaluation
- implement handlers through temp customerbalance facade

## Notes for reviewer

- balance and account APIs will be completely refactored with better package boundaries (eventually)

## Summary by CodeRabbit

* **New Features**
  * Paginated customer credit transactions endpoint with type and currency filters.
  * Transactions show available balance before/after and surface charge/subscription/feature labels.
  * Historical balance queries support cursor-scoped views (balance as-of a transaction).

* **Chores**
  * Server/router now wires ledger/account resolver and respects the credits feature flag.
  * Ledger/collector flows and transaction commits now propagate annotations so metadata is retained.


openmeter/ledger/chargeadapter/annotations.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func chargeAnnotationsForUsageBasedCharge(charge chargeusagebased.Charge) models.Annotations {
+	return chargeTransactionAnnotations(
+		models.NamespacedID{
+			Namespace: charge.Namespace,
+			ID:        charge.ID,
+		},
+		charge.Intent.Subscription,
+		ptrIfNotEmpty(charge.State.FeatureID),
```
turip:
```suggestion
		lo.EmptyableToPtr(charge.State.FeatureID),
```

api/v3/handlers/customers/credits/list_transactions.go
```diff
@@ -0,0 +1,347 @@
...
+				Currency:       request.CurrencyFilter,
+				CreditMovement: creditMovement,
+			}
+
+			result, err := h.ledger.ListTransactionsByPage(ctx, listIn)
+			if err != nil {
+				return ListCreditTransactionsResponse{}, fmt.Errorf("list transactions: %w", err)
+			}
+
+			items, err := mapCreditTransactions(result.Items)
```
turip: I would add some validation here that the list result is single currency only, as single currency is assumed on the rest of execution flow.

api/v3/handlers/customers/credits/list_transactions.go
```diff
@@ -0,0 +1,347 @@
...
+				if args.Params.Filter.Currency != nil {
+					currency := currencyx.Code(*args.Params.Filter.Currency)
+					req.CurrencyFilter = &currency
+				}
+			}
+
+			return req, nil
+		},
+		func(ctx context.Context, request ListCreditTransactionsRequest) (ListCreditTransactionsResponse, error) {
+			creditMovement, empty := creditMovementFromTypeFilter(request.TypeFilter)
```
turip: I think there's way too much logic here inside the http handler. It would be better if we just provide a service call in my opinion so that we are not mixing the http mapping and the actual calculation logic.

turip (dismissed):

turip (dismissed):

turip (approved):

PR #4149, opened 2026-04-16
chore: improve entitlement lookup performance

turip:
## Overview

Two fixes:
- make adapter query simpler
- Skip entitlement fetching in case of IDs
- For list preload customer IDs thus postgresql can rely on proper indexes
- Use feature_key/feature_id IN for lighter queries

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed entitlement filtering to properly handle nonexistent customer keys by returning empty results
  * Improved accuracy of feature identifier resolution when using feature IDs versus feature keys

* **Performance Improvements**
  * Optimized database queries for entitlement lookups by streamlining customer identification logic

### DIFF
--- a/openmeter/entitlement/adapter/entitlement.go
+++ b/openmeter/entitlement/adapter/entitlement.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"entgo.io/ent/dialect/sql"
+	"github.com/oklog/ulid/v2"
 	"github.com/samber/lo"
 
 	customeradapter "github.com/openmeterio/openmeter/openmeter/customer/adapter"
@@ -86,10 +87,9 @@ func (a *entitlementDBAdapter) GetActiveEntitlementOfCustomerAt(ctx context.Cont
 				Where(EntitlementActiveAt(at)...).
 				Where(
 					db_entitlement.Or(db_entitlement.DeletedAtGT(at), db_entitlement.DeletedAtIsNil()),
+					db_entitlement.CustomerID(customerID),
 					db_entitlement.HasCustomerWith(
-						customerdb.Namespace(namespace),
 						customerNotDeletedAt(at),
-						customerdb.ID(customerID),
 					),
 					db_entitlement.Namespace(namespace),
 					db_entitlement.FeatureKey(featureKey),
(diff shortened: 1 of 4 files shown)


openmeter/entitlement/adapter/entitlement.go
```diff
@@ -357,16 +374,30 @@ func (a *entitlementDBAdapter) ListEntitlements(ctx context.Context, params enti
...
-						continue
+				featureIDs := make([]string, 0, len(params.FeatureIDsOrKeys))
+				featureKeys := make([]string, 0, len(params.FeatureIDsOrKeys))
+
+				for _, idOrKey := range params.FeatureIDsOrKeys {
+					if _, err := ulid.Parse(idOrKey); err == nil {
+						featureIDs = append(featureIDs, idOrKey)
+					} else {
+						featureKeys = append(featureKeys, idOrKey)
 					}
```
GAlexIHU, line 385: i think we still don't have "non ulid" validation on keys

GAlexIHU (approved):

openmeter/entitlement/adapter/entitlement.go
turip: Yes, that's right. I tought I have checked that :/

turip: Yes so it's enforced by API validation:

  The API spec also models resource keys separately from ULIDs and gives keys the ^[a-z0-9]+(?:_[a-z0-9]+)*$ pattern in api/spec/packages/aip/src/shared/properties.tsp:16. Since a canonical ULID is 26 chars of Crockford
  base32 and typically includes uppercase letters, it does not fit that key type either.


The legacy/v1 create-feature request uses key: Key in the TypeSpec model at api/spec/packages/legacy/src/productcatalog/features.tsp:22. That Key type is defined with the resource-key regex in api/spec/packages/legacy/src/
  types.tsp:16, so a canonical ULID does not match the schema.

  More importantly, the v1 HTTP handler in openmeter/productcatalog/driver/feature.go:116 maps the request into feature.CreateFeatureInputs and calls the same domain method connector.CreateFeature(...). That method contains
  the hard rejection in openmeter/productcatalog/feature/connector.go:196:

```
  if _, err := ulid.Parse(feature.Key); err == nil { ... "Feature key cannot be a valid ULID" }
```
  So for v1 there are two guards:

  1. API schema: key must match the legacy Key pattern.
  2. Domain logic: even if something bypassed schema validation, feature creation still rejects ULID keys.

PR #4155, opened 2026-04-16
feat: implement realtime expand

turip:
## Overview

Add support for getting the total usage for the charge. The total usage includes the already booked items.

## Summary by CodeRabbit

* **New Features**
  * Added realtime_usage expansion so charge responses can include current usage totals.

* **Improvements**
  * Charges responses now populate realtime totals when realtime_usage is requested.
  * List behavior updated to allow combining include/exclude status filters when valid.

* **Bug Fixes**
  * Validation now rejects requests that set both inclusion and exclusion status filters simultaneously.

* **Tests**
  * Added end-to-end test verifying realtime usage expansion across multiple usage-based charges.

### DIFF
--- a/openmeter/billing/charges/usagebased/service/get.go
+++ b/openmeter/billing/charges/usagebased/service/get.go
@@ -141,16 +141,18 @@ func (s *service) expandChargesUsage(ctx context.Context, namespace string, char
 				}
 			}()
 
-				ratingResult, err := s.rater.GetRatingForUsage(ctx, usagebasedrating.GetRatingForUsageInput{
-					Charge:         charge,
-					Customer:       customerOverridesById[charge.GetCustomerID()],
-					FeatureMeter:   featureMeter,
-					StoredAtOffset: storedAt,
-				})
-				if err != nil {
-					err = fmt.Errorf("rating charge %s: %w", charge.ID, err)
-					return
-				}
+			var ratingResult usagebasedrating.GetRatingForUsageResult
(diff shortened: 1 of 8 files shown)


openmeter/billing/charges/usagebased/service/get.go
```diff
@@ -2,9 +2,25 @@ package service
...
+	"github.com/openmeterio/openmeter/pkg/clock"
 	"github.com/openmeterio/openmeter/pkg/framework/transaction"
+	"github.com/openmeterio/openmeter/pkg/ref"
+	"github.com/openmeterio/openmeter/pkg/slicesx"
+	"github.com/samber/lo"
+	"golang.org/x/sync/semaphore"
+)
+
+const (
+	defaultWorkerCount = 5
```
GAlexIHU: nit: its not a worker

GAlexIHU (dismissed):
why do we need to load live usage for list of charges? (i suppose old line feature-parity but still, these are super expensive)

openmeter/billing/charges/usagebased/service/get.go
turip: renamed

turip:
> why do we need to load live usage for list of charges? (i suppose old line feature-parity but still, these are super expensive)

This is for upcoming charges, we are only loading this if the realtime expand is set (and we should not).

This is feature parity with the gathering invoice.

Once we have periodic realizations we can make it faster by relying on the previous snapshots for some meter types such as count/sum.

GAlexIHU (approved):

PR #4179, opened 2026-04-20
feat: usagebased payment flows

turip:
## Overview

Usage-based credit_then_invoice charges now model payment as part of the active lifecycle instead of hiding it inside final. The state machine  was updated so finalized usage moves into active.payment_pending on invoice issuance, then into active.authorized on payment authorization, and  only reaches final once payment is settled. This makes the charge status reflect the real economic lifecycle and keeps the final meta-state  reserved for fully settled charges.

**Note**: Current states are temporary until we have progressive billing in place.

  To support that flow, usage-based payment handling was added at the run level. The branch introduces run payment persistence for usage-based  realization runs, expands the usage-based handler contract to cover invoice payment authorization and settlement, and reuses the flat-fee ledger  handling pattern for authorized and settled events. The billing invoice/line flow now drives those usage-based payment transitions so late-
  arriving payment events can update already-finalized usage runs correctly.

  The generic charge state machine persistence behavior was also tightened up. FireAndActivate(...) now persists successful transitions directly, and AdvanceUntilStateStable(...) no longer performs an extra write afterward. 

## Summary by CodeRabbit

* **New Features**
  * Invoice-backed charges now remain in active states during payment authorization and settlement and only reach final once payment is fully settled.
  * Introduced explicit payment-authorized and payment-settled handling so invoiced usage and ledger bookings occur at the correct lifecycle points.

* **Tests**
  * Expanded unit and end-to-end tests covering authorization, settlement, direct-paid flows, and related state transitions.

### DIFF
--- a/.agents/skills/charges/SKILL.md
+++ b/.agents/skills/charges/SKILL.md
@@ -54,6 +54,7 @@ The generic rule is:
 - type-specific packages own type-specific lifecycle and persistence
 - `AdvanceCharges(...)` is a facade method, not the state machine itself
 - type-specific service subpackages may own reusable realization mechanics when the parent service becomes too broad; keep state-machine decisions in the lifecycle files and move only mechanical operations such as rating snapshots, realization persistence, credit allocation/correction, and realization lineage persistence into these helpers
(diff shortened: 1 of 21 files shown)


GAlexIHU (approved):
👍

PR #4206, opened 2026-04-22
feat(entitlement): allow grants on subscription-managed entitlements

### LINKED ISSUE #3293: I realize this is an old PR but I'm curious why this is not an intended use case for OpenMeter to support? It seems this is a very natural way to support subscriptions + overage.
I realize this is an old PR but I'm curious why this is not an intended use case for OpenMeter to support? It seems this is a very natural way to support subscriptions + overage. 

I don't believe this can be accomplished by editing the subscription as the subscription is just a standard base and overage should be something extra that's variable-charged on top of that.

I also tried to do this with add-ons but it doesn't work because the add-ons renew on the same basis as the underlying subscription, but overages should be one-off additions to the entitlement and not renew.

It seems this is exactly what grants were designed for, but I can't figure out why grants cannot be applied to entitlements managed by subscriptions. With this restriction in place, it seems the only way to get the desired behavior is to basically reimplement subscriptions using raw entitlements and apply grants onto those.

Is there some other suggested way to manage this with OpenMeter? cc @GAlexIHU

_Originally posted by @christopherhwood in https://github.com/openmeterio/openmeter/issues/2481#issuecomment-3226424350_

ankitk0305:
## Problem

Fixes #3293

Users cannot apply grants (one-time credit top-ups) to entitlements that are managed by a subscription. The `PreUpdate` hook in the subscription guard returns a `403 Forbidden` for any external modification of a subscription-managed entitlement, including grant creation.

This forces users into a poor workaround: abandoning subscriptions entirely and managing entitlements + grants manually with raw entitlements.

### Use case

A customer has a subscription plan that grants 100 API calls/month. They exceed their limit and the operator wants to add a one-time overage grant of 50 extra calls — without changing the subscription plan or creating a recurring add-on. This is a natural billing pattern that should be supported.

## Root Cause

In `openmeter/entitlement/hooks/subscription/hook.go`, `PreUpdate` blocks **all** external mutations on subscription-managed entitlements:

```go
func (h *hook) PreUpdate(ctx context.Context, ent *entitlement.Entitlement) error {
    if subscription.IsSubscriptionOperation(ctx) {
        return nil
    }
    if subscription.AnnotationParser.HasSubscription(ent.Annotations) {
        return models.NewGenericForbiddenError(...) // blocks grants too
    }
    return nil
}
```

`CreateGrant` in `entitlement/metered/entitlement_grant.go` calls this hook before creating the grant, so every grant attempt on a subscription-managed entitlement is rejected.

## Solution

Grants are ad-hoc credit top-ups — they do not alter the entitlement structure (type, feature key, config) that the subscription owns. The guard should distinguish between structural mutations (still blocked) and grant creation (now allowed).

The fix follows the existing `IsSubscriptionOperation` context pattern:

1. **`subscription/context.go`** — add `NewGrantOperationContext` / `IsGrantOperation` to mark a context as a grant operation
2. **`entitlement/hooks/subscription/hook.go`** — `PreUpdate` passes through when `IsGrantOperation(ctx)` is true
3. **`entitlement/metered/entitlement_grant.go`** — `CreateGrant` wraps `ctx` with `NewGrantOperationContext` before calling `PreUpdate`

## Changes

```
openmeter/subscription/context.go                  +15
openmeter/entitlement/hooks/subscription/hook.go   +6
openmeter/entitlement/metered/entitlement_grant.go +2
```

**Invariants preserved:**
- Subscriptions still fully own entitlement lifecycle (create / delete / config changes remain blocked for external callers)
- Grant creation on non-subscription entitlements is unaffected
- The `IsSubscriptionOperation` bypass for the subscription subsystem itself is unchanged

## Test Plan

- [ ] Create a subscription with a metered entitlement
- [ ] Issue a grant on the subscription-managed entitlement via `POST /api/v2/customers/{id}/entitlements/{featureKey}/grants` — should return `200` instead of `403`
- [ ] Verify the grant appears in `GET .../grants`
- [ ] Verify the subscription still controls the entitlement lifecycle (delete entitlement directly should still return `403`)
- [ ] Run `make test`

## Summary by CodeRabbit

* **Chores**
  * Updated Helm chart dependencies: Kafka (30.1.8 → 32.4.4), PostgreSQL (16.1.2 → 17.1.0), and Redis (20.2.1 → 23.1.1)

* **New Features**
  * Added support for grant operations in subscription entitlements, enabling specific entitlement updates during grant creation workflows

### DIFF
--- a/deploy/charts/openmeter/Chart.lock
+++ b/deploy/charts/openmeter/Chart.lock
@@ -4,12 +4,12 @@ dependencies:
   version: 0.23.3
 - name: kafka
   repository: oci://registry-1.docker.io/bitnamicharts
-  version: 30.1.8
+  version: 32.4.4
 - name: postgresql
   repository: oci://registry-1.docker.io/bitnamicharts
-  version: 16.1.2
+  version: 17.1.0
 - name: redis
   repository: oci://registry-1.docker.io/bitnamicharts
-  version: 20.2.1
-digest: sha256:984e797866c51fc1d921e3490d3f88b1023257345849449ffb0adb457e39a529
-generated: "2025-08-17T14:11:37.936915+02:00"
+  version: 23.1.1
+digest: sha256:0000000000000000000000000000000000000000000000000000000000000000
+generated: "2026-04-22T00:00:00.000000+00:00"
--- a/deploy/charts/openmeter/Chart.yaml
+++ b/deploy/charts/openmeter/Chart.yaml
@@ -19,15 +19,15 @@ dependencies:
     repository: "https://docs.altinity.com/clickhouse-operator/"
     condition: clickhouse.operator.install
   - name: kafka
-    version: "30.1.8"
+    version: "32.4.4"
     repository: "oci://registry-1.docker.io/bitnamicharts"
     condition: kafka.enabled
   - name: postgresql
-    version: "16.1.2"
+    version: "17.1.0"
     repository: "oci://registry-1.docker.io/bitnamicharts"
     condition: postgresql.enabled
   - name: redis
-    version: "20.2.1"
+    version: "23.1.1"
     repository: "oci://registry-1.docker.io/bitnamicharts"
     condition: redis.enabled
 
--- a/deploy/charts/openmeter/values.yaml
+++ b/deploy/charts/openmeter/values.yaml
@@ -125,6 +125,9 @@ kafka:
   # **Not recommended for production environments.**
   enabled: true
 
+  image:
+    registry: registry.bitnami.com
+
   listeners:
     client:
       name: plain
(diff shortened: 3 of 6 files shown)


GAlexIHU:
Hi @ankitk0305 

The subscription system works by "closing" and "recreating" entitlements according to subscription / addon changes. This is partly due to the limitations of the current entitlement-grant system. Manual granting is disabled on subscription managed entitlements because the two systems cannot currently reconcile and the results would be confusing to the end user: let's say, you add a manual grant of +100. On the next subscription edit (or any general "change type event"), that 100 grant would be lost with the closed entitlement.

GAlexIHU:
On a related note, if you check recent activity in the repo, the first version of the new credit system is close to its first release, that will be a better solution for most usecases.

PR #4209, opened 2026-04-22
feat(api): better transaction list API

GAlexIHU:
## Summary

This PR refactors customer credit transaction listing to return user-meaningful activity while moving the endpoint to cursor pagination semantics that are consistent with our current implementation approach (opaque tokens).

## Why

The old transaction list exposed low-level ledger rows 1:1, which is confusing for users (especially for funded credits where one purchase may map to multiple ledger entries).  
We now present activity aligned with what users expect to see: funded and consumed movements.

## What Changed

- Reworked credit transaction domain semantics:
  - kept `funded` and `consumed`
  - removed `adjusted` from this listing surface for now

- Introduced funded activity source from Charges:
  - added funded credit activity listing in Charges adapter/service
  - funded activity rows are built from charge/grant data (charge-centric representation)

- Kept consumed activity source from Ledger:
  - consumed rows continue to come from ledger transactions

- Added loader-based aggregation in customer balance:
  - separate funded and ledger loaders
  - merge-sorted by transaction cursor tuple (`booked_at`, `created_at`, `id`)
  - removed page-based dependency from this flow

- Cursor pagination updates:
  - switched endpoint query to `Common.CursorPaginationQuery`
  - endpoint response uses opaque cursor tokens (not URI links), with explicit comments documenting why
  - opaque token payload still encodes the 3-part internal cursor (`booked_at`, `created_at`, `id`)
  - implemented `before`/`after` handling through the stack for this use case
  - removed `first`/`last` cursors from API response meta (kept `next`/`previous`)

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added cursor-based pagination for credit transaction listings
  * Added support for listing funded credit activities with cursor pagination

* **Changes**
  * Replaced page-based pagination with cursor-based pagination for credit transactions
  * Removed "adjusted" transaction type; credit transactions now support only "funded" and "consumed" types
  * Updated credit transaction API response structure with new pagination metadata format


openmeter/billing/charges/adapter/search.go
```diff
@@ -104,6 +109,157 @@ func (a *adapter) ListCharges(ctx context.Context, input charges.ListChargesInpu
 	})
 }
 
+func (a *adapter) ListFundedCreditActivities(ctx context.Context, input charges.ListFundedCreditActivitiesInput) (charges.ListFundedCreditActivitiesResult, error) {
```
turip: Do you foresee that this will be needed for other types of charges or just credit purchases?

If it's for credit purchase putting it into creditpurchase might make sense, esp that there you can reuse the adapter methods for testing.

GAlexIHU: just for creditpurchase so the "+" values show up as a single row. i'll move it there

turip (approved):

PR #4211, opened 2026-04-23
feat(api): add v3 list events endpoint

tothandras:
## Summary by CodeRabbit

* **New Features**
  * New API endpoint to list ingested metering events; responses include ingestion and storage timestamps.

* **Improvements**
  * Added deep-object filters for stored_at, identity, metadata, and customer.
  * Flexible sorting by time, ingested_at, or stored_at.
  * Pagination query behavior refined and next-page signaling now only set when a full page of results is returned.

* **Tests**
  * New tests covering event list parsing, filtering, sorting, and pagination.


api/spec/packages/aip/src/common/pagination.tsp
```diff
@@ -15,12 +15,21 @@ model CursorPaginationQuery {
   /**
    * Determines which page of the collection to retrieve.
    */
-  #suppress "@openmeter/api-spec-aip/casing" "common model"
-  @query
-  @useRef("../../../../common/definitions/metadatas.yaml#/components/parameters/CursorPageQuery")
+  @query(#{ explode: true, style: "deepObject" })
   page?: {
```
GAlexIHU, line 19: q: do we wanna support first / last (as per [AIP](https://kong-aip.netlify.app/aip/158/#cursor)) ?

api/v3/handlers/events/list.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+	if len(f.Oeq) > 0 {
+		values = append(values, f.Oeq...)
+	}
+
+	if len(values) == 0 {
+		return nil, nil
+	}
+
+	return &filter.FilterString{In: &values}, nil
```
gergely-kurucz-konghq, line 210: Do I get it right, that Eq and Oeq will be interpreted the same way in the end? An "In" matching will be done, i.e. any event with matching Eq or Oeq customerId will be returned?

gergely-kurucz-konghq: Or is it even possible to do a query including both Eq and Oeq filters? Does something prevent this at a higher layer?

openmeter/streaming/clickhouse/event_query_v2.go
```diff
@@ -81,25 +82,49 @@ func (q queryEventsTableV2) toSQL() (string, []interface{}) {
...
+					query.LessThan("id", q.Params.Cursor.ID),
+				),
+			)
+		}
+	}
+
+	direction := "DESC"
+	if q.Params.SortOrder == sortx.OrderAsc {
+		direction = "ASC"
+	}
```
gergely-kurucz-konghq: Tha API spec writes this:
> The `asc` suffix is optional as the default sort order is ascending.

Seems to me the default is DESC in this implementation. Which is correct?

gergely-kurucz-konghq (changes requested):
Minor comments for clarification, otherwise LGTM!

api/spec/packages/aip/src/common/pagination.tsp
tothandras: no, it's optional and due to the events dynamic nature, we won't be able to add it

openmeter/streaming/clickhouse/event_query_v2.go
```diff
@@ -81,25 +82,49 @@ func (q queryEventsTableV2) toSQL() (string, []interface{}) {
...
+
+	// Order by sort column and id for stable ordering; tie-break flips with the primary direction.
+	query.OrderBy(fmt.Sprintf("%s %s", sortCol, direction)).OrderBy(fmt.Sprintf("id %s", direction))
```
chrisgacsal:
```suggestion
    switch q.Params.SortOrder {
	case sortx.OrderAsc:
		query.OrderByAsc(sortCol).OrderByAsc("id")
	case sortx.OrderDesc:
		fallthrough
	default:
		query.OrderByDesc(sortCol).OrderByDesc("id")
	}
```

chrisgacsal (changes requested):

api/v3/handlers/events/list.go
tothandras: Yes. The behavior is an `AND` operation between the operands, see https://github.com/openmeterio/openmeter/blob/main/api/v3/filters/convert.go
For `customer_id`, as it's not part of the CH table, we can only support `in`

turip (approved):

PR #4224, opened 2026-04-24
feat(productcatalog): expose taxCodeId in taxConfig API

borbelyr-kong:
## Summary by CodeRabbit

* **New Features**
  * Added optional tax code identifier support for tax configurations; ID takes precedence over provided Stripe codes.
  * Automatic backfill of Stripe code from stored tax-code mappings when using the tax code identifier.

* **Deprecated**
  * Stripe and external invoicing fields marked deprecated in favor of the tax code identifier.

* **Tests**
  * Expanded coverage for tax-code resolution across plans, addons, subscriptions; improved test isolation to avoid shared-state issues.


mark-vass-konghq (dismissed):

openmeter/productcatalog/tax.go
```diff
@@ -163,6 +164,55 @@ func (s StripeTaxConfig) Clone() StripeTaxConfig {
...
+	}
+
+	switch {
+	case cfg.TaxCodeID != nil:
+		tc, err := svc.GetTaxCode(ctx, taxcode.GetTaxCodeInput{
+			NamespacedID: models.NamespacedID{Namespace: namespace, ID: *cfg.TaxCodeID},
+		})
+		if err != nil {
+			if taxcode.IsTaxCodeNotFoundError(err) {
+				return models.NewGenericValidationError(fmt.Errorf("tax code %s not found", *cfg.TaxCodeID))
```
tothandras, line 191: Should we just return `taxcode.IsTaxCodeNotFoundError` as is? Can you explain why validation error instead of not found error is being returned?

openmeter/productcatalog/tax.go
```diff
@@ -163,6 +164,55 @@ func (s StripeTaxConfig) Clone() StripeTaxConfig {
...
+			}
+			return fmt.Errorf("resolving tax code %s: %w", *cfg.TaxCodeID, err)
+		}
+		if m, ok := tc.GetAppMapping(app.AppTypeStripe); ok {
+			cfg.Stripe = &StripeTaxConfig{Code: m.TaxCode}
+		} else {
```
tothandras, line 197: Add test for this case.

openmeter/billing/service/profile.go
```diff
@@ -631,32 +630,8 @@ func (s *Service) ResolveStripeAppIDFromBillingProfile(ctx context.Context, name
 	return appID, nil
 }
 
-// resolveDefaultTaxCode resolves the TaxCode entity for a DefaultTaxConfig's Stripe.Code and
-// stamps TaxCodeID back onto the pointed-to config before it is persisted. Always re-resolves
-// from Stripe.Code so that changing the code (txcd_A → txcd_B) on an already-stamped config
-// updates the FK rather than leaving the stale value. When Stripe is nil or Code is empty,
-// TaxCodeID is explicitly cleared so that a read-modify-write that removes the Stripe code
-// does not leave a stale FK in the DB. No-op when taxConfig is nil;
```
tothandras, line 639: Do we still cover this behavior for clearing Stripe tax codes?

tothandras (changes requested):

openmeter/billing/service/profile.go
borbelyr-kong: Yes, it is covered with `ResolveTaxConfig`. Also it has a test `TestProfileUpdateClearsTaxCodeFK`.

openmeter/productcatalog/tax.go
borbelyr-kong: This function runs on the write path, so when an `IsTaxCodeNotFoundError` happens it is because of a malformed input. I think in this case the 400 with the validation error is the right choice.

openmeter/productcatalog/tax.go
borbelyr-kong: I've added unit tests for `ResolveTaxConfig`, including the case where `TaxCodeID` resolves to an entity with no Stripe app mapping.

tothandras (dismissed):

tothandras (approved):

PR #4251, opened 2026-04-28
Chore/e2e httpyac skill

rolosp:
## Overview

This is an experiment for creating a httpyac `.http` file  e2e solution with a dedicated skill that would take the `e2e-nl` natural language testing representation as an input and produce an output for httpYac.


rolosp:
During the e2e tool discussion we agreed on not going further with httpYac

PR #4263, opened 2026-04-30
feat(api): add tax id and tax behavior to charges list api

mark-vass-konghq:
@coderabbit

## Summary by CodeRabbit

* **New Features**
  * Optional per-charge TaxConfig added for flat-fee and usage-based charges; structured tax_code reference introduced and takes precedence over deprecated tax_code_id and legacy provider codes.
  * Shared TaxCodeReference model added and API mappings updated to surface tax_code consistently.

* **Tests**
  * New unit and integration tests covering TaxConfig conversion, precedence rules, API mappings, and listing behavior.

* **Documentation**
  * Deprecated fields updated with guidance to use tax_code instead.


api/spec/packages/aip/src/customers/charges/charges.tsp
```diff
@@ -154,6 +154,34 @@ model ChargeBase<T extends ChargeType> {
...
+  @summary("Tax configuration")
+  tax_config?: ChargeTaxConfig;
+}
+
+/**
+ * Tax configuration of a charge.
+ */
+@friendlyName("BillingChargeTaxConfig")
+@summary("Charge tax configuration")
+model ChargeTaxConfig {
```
tothandras: We should probably reuse [`Billing.TaxConfig`](https://github.com/openmeterio/openmeter/blob/main/api/spec/packages/aip/src/billing/tax.tsp#L7) instead of introducing a new type.

tothandras (changes requested):

api/spec/packages/aip/src/customers/charges/charges.tsp
tothandras: Since `tax_code_id` is already part of some public API's, we should deprecate it and introduce `tax_code { id }` as in your `ChargeTaxConfig`.

api/spec/packages/aip/src/billing/tax.tsp
tothandras:
```suggestion
  * @deprecated Use `tax_code` instead. When both `stripe.code` and `tax_code` are provided,
  * `tax_code` takes precedence and `stripe.code` is ignored.
```
Same for other comments.

api/spec/packages/aip/src/billing/tax.tsp
```diff
@@ -1,5 +1,17 @@
 namespace Billing;
 
+/**
+ * A reference to a tax code entity.
+ */
+@friendlyName("BillingTaxCodeReference")
+@summary("Tax code reference")
+model TaxCodeReference {
```
borbelyr-kong: Could we use `Shared.ResourceReference<Tax.TaxCode>` ?

borbelyr-kong (dismissed):

tothandras (approved):

PR #4318, opened 2026-05-07
Feat(subscription addon): GetByID endpoint

borosr:
Example request: 
```sh
curl --request GET \
  --url 'http://localhost:8888/api/v3/openmeter/subscriptions/01KQZ4K3R4YT29RS928Z1VC92T/addons/01KQZ4S6447Y41HZ3WMWQZ7QTS'
  ```

## Summary by CodeRabbit

* **New Features**
  * Added a GET API to fetch subscription add-on associations, returning addon reference, quantity with effective timestamp, active-from/active-to window, audit metadata, and supporting lookup by add-on ID or by subscription+addon identifier.

* **Tests**
  * Updated subscription add-on service tests to cover the new retrieval and lookup behaviors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openmeterio/openmeter/pull/4318)

### CODE REVIEW

File: api/v3/handlers/subscriptions/subscriptionaddons/convert.go [hunk 1]
```diff
@@ -0,0 +1,51 @@
...
+		Addon:       apiv3.AddonReference{Id: a.Addon.ID},
+		CreatedAt:   a.CreatedAt,
+		UpdatedAt:   a.UpdatedAt,
+		DeletedAt:   a.DeletedAt,
+		Labels:      labels.FromMetadata(a.Metadata),
+		Quantity:    inst.Quantity,
+		QuantityAt:  now,
+		ActiveFrom:  *union.From,
+		ActiveTo:    union.To,
+	}, nil
```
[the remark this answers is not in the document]
borosr: I used v1 implementation as an inspiration. But if we want to handle this, what about `lo.FromPtrOr(union.From, now)`? @tothandras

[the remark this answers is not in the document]
tothandras: Good call — applied `lo.FromPtrOr(union.From, now)`. The current invariant (`GetInstances` filters out periods with `From == nil`) keeps it safe today, but defending against the three stacked "should never happen" guards is cheap and worth it. v1's `mapping.go:97` left as-is for now.

openmeter/subscription/addon/service.go
```diff
@@ -67,3 +69,41 @@ func (i ListSubscriptionAddonsInput) Validate() error {
...
+// by the subscription and add-on identifiers.
+type GetSubscriptionAddonInput struct {
+	models.NamespacedID
+
+	// SubscriptionID
+	SubscriptionID string `json:"subscriptionId"`
+
+	// AddonIDOrKey
+	AddonIDOrKey string `json:"addonIdOrKey"`
+}
```
GAlexIHU, line 78: if we wanna go fancy you should use `pkg/filter` otherwise this is technically solvable with the former API + a == check

the way you did it here just feels a bit weird to me

openmeter/subscription/addon/service.go
```diff
@@ -67,3 +69,41 @@ func (i ListSubscriptionAddonsInput) Validate() error {
...
+			errs = append(errs, errors.New("subscription id is not a valid ULID"))
+		}
+
+		if i.AddonIDOrKey == "" {
+			errs = append(errs, errors.New("add-on id or key must be provided if assignment id is not provided"))
+		}
+	} else {
+		if _, err := ulid.Parse(i.ID); err != nil {
+			errs = append(errs, errors.New("assignment id is not a valid ULID"))
+		}
```
GAlexIHU: e.g. here, `i.NamespacedID.Namespace` but then `i.ID` not `i.NamespacedID.ID` (+ rolled up props can get overloaded if you expand this struct)

also this validation is just a bit weird for me, i think you'd either 1) validate case where get by id 2) validate case where get by subs + idOrKey 3) likely dont allow weird combinations

api/v3/handlers/subscriptions/subscriptionaddons/get.go
```diff
@@ -0,0 +1,59 @@
...
+			if err != nil {
+				return GetSubscriptionAddonRequest{}, err
+			}
+
+			return GetSubscriptionAddonRequest{
+				NamespacedID: models.NamespacedID{
+					Namespace: ns,
+					ID:        params.SubscriptionAddonID,
+				},
+				SubscriptionID: params.SubscriptionID,
```
GAlexIHU, line 38: yeah i think idorkey isnt even used anywhere, you use normal id here

GAlexIHU (changes requested):
the way you chose to do this feels a bit weird for me. can you please just copy the implementation/api from somewhere else where we do this?

api/v3/handlers/subscriptions/subscriptionaddons/get.go
borosr: yes, subscription has ID only

openmeter/subscription/addon/service.go
borosr: Thanks for the good catches.
For the validation what do you think about removing the AddonIDOrKey filter once and for all? 
Since, the same addon can be attached to the certain subscription multiple times, the subscriptionID+addonID pair is not going to return a unique result as it would work in the plan.

And after that he validation would look like this:
```go

	if i.NamespacedID.Namespace == "" {
		errs = append(errs, errors.New("namespace is required"))
	}

	if i.NamespacedID.ID != "" {
		if _, err := ulid.Parse(i.NamespacedID.ID); err != nil {
			errs = append(errs, errors.New("assignment id is not a valid ULID"))
		}
	}
	if i.SubscriptionID != "" {
		if _, err := ulid.Parse(i.SubscriptionID); err != nil {
			errs = append(errs, errors.New("subscription id is not a valid ULID"))
		}
	}
```
The subscriptionID should be optional, because in some internal usages, we only have the subscriptionAddonID.

GAlexIHU: yeah i think thats fine, also if `pkg/filter` can be used here @tothandras  would probably like that

openmeter/subscription/addon/service.go
```diff
@@ -67,3 +69,32 @@ func (i ListSubscriptionAddonsInput) Validate() error {
...
+func (i GetSubscriptionAddonInput) Validate() error {
+	var errs []error
+
+	if err := i.NamespacedID.Validate(); err != nil {
+		errs = append(errs, err)
+	}
+
+	if _, err := ulid.Parse(i.NamespacedID.ID); err != nil {
+		errs = append(errs, errors.New("assignment id is not a valid ULID"))
+	}
```
tothandras: We don't usually do this elsewhere.

GAlexIHU (dismissed):

tothandras (dismissed):

GAlexIHU (dismissed):

api/v3/handlers/subscriptions/subscriptionaddons/convert.go
```diff
@@ -15,8 +17,10 @@ import (
 func toAPISubscriptionAddon(addon subscriptionaddon.SubscriptionAddon) (apiv3.SubscriptionAddon, error) {
 	now := clock.Now()
 
-	// If no instance is active at `now`, quantity stays 0.
-	inst, _ := addon.GetInstanceAt(now)
+	inst, found := addon.GetInstanceAt(now)
+	if !found {
+		return apiv3.SubscriptionAddon{}, fmt.Errorf("no instance is active at %s", now.Format(time.RFC3339))
```
tothandras: This should return apierror or models error.

tothandras (changes requested):

tothandras (approved):

PR #4329, opened 2026-05-09
feat(api): add governance query operation

tothandras:
## Summary by CodeRabbit

* **New Features**
  * Added a Governance API to evaluate customer feature access in bulk with per-customer results, timestamps, and partial error info
  * Exposed governance endpoints and tags on the public API surface
  * Introduced reusable cursor-based pagination for queries

* **Documentation**
  * Added a shared generic inline error model for consistent per-item/partial-error reporting

* **Refactor**
  * Consolidated product-catalog validation errors to reuse the shared inline error model

[Review Change Stack](https://app.coderabbit.ai/change-stack/openmeterio/openmeter/pull/4329)


api/spec/packages/aip/src/governance/governance.tsp
```diff
@@ -0,0 +1,240 @@
...
+  customer: Customers.Customer;
+
+  /**
+   * Whether the customer has any available prepaid credit balance.
+   *
+   * Only populated when the request was made with `include_credits=true`.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Has available credits")
+  has_available_credits?: boolean;
```
gergely-kurucz-konghq: According to the proposal document, there is no `has_available_credits` in the response. Instead, if a feature is credit gated AND `include_credits: true` was in the request, then if that feature is not accessible due to credit is not available, it will have a `denied_reason: "no_credit_available"`.

Is this right, @chrisgacsal?

tothandras: Updated

api/spec/packages/aip/src/governance/governance.tsp
```diff
@@ -0,0 +1,234 @@
...
+  FeatureUnavailable: "feature_unavailable",
+
+  /**
+   * The feature `key` referenced by the request is not configured in the
+   * organization.
+   */
+  FeatureNotFound: "feature_not_found",
+
+  /**
+   * The customer has no available prepaid credit balance.
```
gergely-kurucz-konghq, line 194:
```suggestion
   * The feature is credit-gated and the customer has no available prepaid credit balance.
```

WDYT?

api/spec/packages/aip/src/governance/governance.tsp
```diff
@@ -0,0 +1,234 @@
...
+ */
+@friendlyName("GovernanceQueryFeaturesFilter")
+model GovernanceQueryFeaturesFilter {
+  /**
+   * List of feature keys to evaluate access for.
+   */
+  @summary("Feature Keys")
+  @minItems(1)
+  @maxItems(1000)
+  keys: string[];
```
gergely-kurucz-konghq: AFAIK we communicated a max of 50 customers and 50 features. Do we want to surface this limit on the API as well? Is it hard to update this limit once the endpoint goes public?

WDYT @chrisgacsal?

tothandras: Yes, let me limit it.

api/spec/packages/aip/src/governance/governance.tsp
```diff
@@ -0,0 +1,234 @@
...
+  @visibility(Lifecycle.Read)
+  @summary("Updated at")
+  updated_at: Shared.DateTime;
+}
+
+/**
+ * Access status for a single feature.
+ */
+@friendlyName("GovernanceFeatureAccess")
+model GovernanceFeatureAccess {
```
gergely-kurucz-konghq, line 137: The proposal document also contains a `description` field for each feature access object. Did you discuss with @chrisgacsal that it is not needed?

gergely-kurucz-konghq: Correction, in the proposal the object shape looks like this:
```json
{
  "has_access": false,
  "reason": {
    "code": "NO_CREDIT_AVAILABLE",
    "description": "Customer has no credit available."
  }
}
```

api/spec/packages/aip/src/governance/governance.tsp
```diff
@@ -0,0 +1,234 @@
...
+
+/**
+ * Filters to apply to the governance query.
+ */
+@friendlyName("GovernanceQueryFilters")
+model GovernanceQueryFilters {
+  #suppress "@openmeter/api-spec-aip/doc-decorator" "-"
+  customers: GovernanceQueryCustomersFilter;
+  #suppress "@openmeter/api-spec-aip/doc-decorator" "-"
+  features?: GovernanceQueryFeaturesFilter;
```
gergely-kurucz-konghq: Do I understand it correctly, that with this, the request object shape would be:
```json
{
  "customers": {
     "keys": [
      ...
    ]
  },
  "features": {
     "keys": [
      ...
    ]
  },
  "include_credits": true
}
```
?

api/spec/packages/aip/src/governance/governance.tsp
tothandras: `denied_reason` ha `message`

api/spec/packages/aip/src/governance/governance.tsp
gergely-kurucz-konghq: Versus the proposed shape:
```json
{
  "customers": [
    "usage_attribution_key1",
    "usage_attribution_key2",
    "customer_key1"
  ],
  "features": [
    "feature_key1",
    "feature_key2",
    "feature_key3",
    "feature_key4",
    "feature_key5"
  ],
  "include_credits": true
}
```

api/spec/packages/aip/src/governance/governance.tsp
gergely-kurucz-konghq: My bad! I did not fully parse how this works:
```
model GovernanceFeatureAccessDeniedReason
  is Shared.BaseError<GovernanceFeatureAccessDeniedReasonCode>;
```
Got it, cool!

api/spec/packages/aip/src/governance/governance.tsp
chrisgacsal: It makes sense only if we expect customers, features to be filtered by other than `keys` properties. Lets keep it as it is proposed in this spec so we can extend it easier later on.

api/spec/packages/aip/src/governance/governance.tsp
chrisgacsal: Yeah, lets limit both customer and feature filter to 50 each.

api/spec/packages/aip/src/governance/governance.tsp
chrisgacsal: Agreed.

chrisgacsal (dismissed):

gergely-kurucz-konghq (approved):

PR #4336, opened 2026-05-11
feat(api): add v3 customer and subscription list filters

tothandras:
## Release Notes

* **New Features**
  * Extended customer list filtering with `usage_attribution_subject_key`, `plan_key`, and `billing_profile_id` parameters
  * Added `plan_key` filter support to subscription list endpoint
  * Improved filtering precision for both endpoints with enhanced query parameter validation

[Review Change Stack](https://app.coderabbit.ai/change-stack/openmeterio/openmeter/pull/4336)


openmeter/customer/adapter/customer.go
```diff
@@ -56,18 +57,35 @@ func (a *adapter) ListCustomers(ctx context.Context, input customer.ListCustomer
...
-		if input.Subject != nil {
-			query = query.Where(customerdb.HasSubjectsWith(
-				customersubjectsdb.SubjectKeyContainsFold(*input.Subject),
-				customersubjectsdb.Or(
-					customersubjectsdb.DeletedAtIsNil(),
-					customersubjectsdb.DeletedAtGTE(now),
-				),
-			))
+		if input.PlanKey != nil {
+			applyActiveSubscriptionFilterWithPlanKey(query, now, input.PlanKey)
```
solkimicreb, line 61: Does this include 'canceled' subs? Not a biggi but canceled are technically "active". It means it will be cancelled at a scheduled date but is still active. (Inactive means already canceled)

api/spec/packages/aip/src/customers/operations.tsp
```diff
@@ -19,20 +19,18 @@ namespace Customers;
...
+  #suppress "@openmeter/api-spec-aip/doc-decorator" "filter field"
   name?: Common.StringFieldFilter;
-
-  /**
-   * Filter customers by primary email.
-   */
+  #suppress "@openmeter/api-spec-aip/doc-decorator" "filter field"
   primary_email?: Common.StringFieldFilter;
+  #suppress "@openmeter/api-spec-aip/doc-decorator" "filter field"
+  usage_attribution_subject_key?: Common.StringFieldFilter;
```
solkimicreb, line 29: Generic app wide feedback. I think we do not really have subjectKey anymore just `subject`. API uses both at places and customer has `subject_key` so it's not the time to change it imo just a thought.

solkimicreb (approved):

turip (approved):

api/spec/packages/aip/src/customers/operations.tsp
tothandras: Unfortunately, it's already usage_attribution.subject_keys on the customer :(

PR #4339, opened 2026-05-11
chore: usagebased credit-then-invoice [shrink]

turip:
## Overview

Usage-based credit-then-invoice shrink preserves pending gathering lines when no realization run needs deletion. In that case, the existing gathering line is updated in place instead of being deleted and  recreated.

When shrink targets an immutable invoice-backed realization, the invoice and ledger bookings remain intact. The invoice receives the immutable-invoice/prorating validation warning, and the old realization  run is retained as audit history.

Realization runs that represent invoice history that should have been deleted but could not be deleted because credit-note/prorating support is missing are marked as `invalid_due_to_unsupported_credit_note`. Their original run type is preserved in `initial_type`.

Voided billing history, meaning deleted runs and unsupported-credit-note runs, is ignored by future billing math. This applies to rating, billed-history quantity calculation, aggregate totals, payment- settlement checks, shrink/extend run selection, and ledger pending-balance impact.

Ledger balances are not reversed for immutable invoice history during shrink. The immutable invoice remains the financial source of truth until proper credit-note/prorating support exists.

## Summary by CodeRabbit

* **New Features**
  * Added native support for shrinking usage-based charges in CreditThenInvoice flows.
  * Added an unsupported-credit-note pathway to gracefully handle deletions targeting immutable invoices (audit-only).

* **Bug Fixes**
  * Fixed invoice date handling so extend/shrink patches preserve explicit invoice timestamps.
  * Billing aggregation and rating now ignore voided/unsupported billing-history runs.

* **Documentation**
  * Clarified charge patch semantics, invoice-at handling, and lifecycle guidance.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openmeterio/openmeter/pull/4339)


tothandras:
## /review findings (Claude Code + Codex adversarial pass)

Posting the three findings with confidence ≥ 8. Lower-confidence items (migration rolling-deploy concern, `unsupportedCreditNoteLines` nil-branch behavior, open-by-default enum policy) are kept out for brevity but happy to expand if useful.

### 🔴 CRITICAL — `DeleteCharge` uses stale `DeletedAt` check (confidence 9/10)

`openmeter/billing/charges/usagebased/service/creditheninvoice.go:245`

```go
for _, run := range s.Charge.Realizations {
    // Deleted realizations were already cleaned up through billing, so the
    // charge delete patch must not emit another invoice deletion for them.
    if run.DeletedAt != nil {   // ← should be run.IsVoidedBillingHistory()
        continue
    }
    ...
    patches = append(patches, invoiceupdater.NewDeleteLinePatch(...))
}
```

If a charge has a run that was reclassified to `RealizationRunTypeInvalidDueToUnsupportedCreditNote` via the new immutable-invoice shrink path, deleting the charge will emit `DeleteLinePatch` for the immutable-invoice line again. That re-triggers `OnUnsupportedCreditNote` and re-appends the `"line should be deleted, but the invoice is immutable"` validation warning. The whole point of the new `IsVoidedBillingHistory()` predicate is to unify these two cases — this site missed the migration.

**Fix:** replace `run.DeletedAt != nil` with `run.IsVoidedBillingHistory()`.

---

### 🔴 CRITICAL — `getRunForLine` uses stale `DeletedAt` check (confidence 9/10)

`openmeter/billing/charges/usagebased/service/creditheninvoice.go:741`

```go
func getRunForLine(charge usagebased.Charge, lineID string) (usagebased.RealizationRun, error) {
    for _, run := range charge.Realizations {
        // Deleted realizations are no longer valid targets for invoice lifecycle callbacks.
        if run.DeletedAt != nil {   // ← should be run.IsVoidedBillingHistory()
            continue
        }
        if run.LineID != nil && *run.LineID == lineID {
            return run, nil
        }
    }
    ...
}
```

Same defect, different blast radius. Callers are `recordPaymentAuthorized` and `recordPaymentSettled` in `payments.go:55-91`. If a payment callback arrives for an immutable-invoice line whose run was voided post-issuance, this helper returns the voided run and `BookInvoicedPaymentAuthorized` / `SettleInvoicedPayment` then books a payment record onto an audit-only run.

`areAllInvoicedRunsSettled` (`payments.go:127`) correctly skips voided runs when deciding settlement state — good — but the spurious payment record is already persisted to the voided run by then. Exposure is small in the test scenarios because all immutable invoices in the suite are fully credit-covered (no fiat tail), but for a partially-credited immutable invoice with a real payment, this is silent ledger drift on the very run the PR introduces.

**Fix:** replace `run.DeletedAt != nil` with `run.IsVoidedBillingHistory()`.

---

### 🟡 INFORMATIONAL — Test gap in `TestUsageBasedCreditThenInvoiceShrinkExtendShrinkPreservesImmutableRunsAndLedger` (confidence 8/10)

`test/credits/credit_then_invoice_test.go:2080-2084`

```go
secondRun, err := charge.Realizations.GetByID(secondRunID.ID)
s.NoError(err)
s.Nil(secondRun.DeletedAt)
s.Equal(extendedServicePeriodTo, secondRun.ServicePeriodTo)
s.RequireAllRunsNonDeleted(charge.Realizations)
```

This is the test that drives the immutable-invoice shrink path end-to-end, and it never asserts `secondRun.Type == RealizationRunTypeInvalidDueToUnsupportedCreditNote`. `RequireAllRunsNonDeleted` only inspects `DeletedAt`. The simpler scenario at `:1295` does assert both `Type` and `InitialType` — that pattern should be repeated here, otherwise the most complex shrink scenario can silently regress to leaving the run as `Final`.

**Fix:**

```go
s.Equal(usagebased.RealizationRunTypeInvalidDueToUnsupportedCreditNote, secondRun.Type)
s.Equal(usagebased.RealizationRunTypeFinalRealization, secondRun.InitialType)
s.True(secondRun.IsVoidedBillingHistory())
```

---

### Validated and dropped (no action needed)

- **`ApplyPatches` reorder** (`invoiceupdate.go:38-49`): deletes are staged into `updatedLinesByInvoiceID` before `provisionUpcomingLines`; per-invoice commits happen later. Safe.
- **`After`→`!Before` strict-inequality on `PatchShrink.ValidateWith`** (`patchshrink.go:132-156`): the only in-repo caller (`reconciler.go:161`) already gates with strict `Before`. Compatible.
- **Double-billing across immutable invoice in the shrink-extend-shrink test**: the assertion of 15 credits consumed against 8 units of usage in the final service period is the PR's documented behavior — invoice is source of truth, credit notes will eventually refund.

---

🤖 Review run via /review (Claude Opus 4.7 + Codex adversarial pass). Findings below confidence 8 omitted from this comment.

tothandras (dismissed):

turip:
> openmeter/billing/charges/usagebased/service/creditheninvoice.go:741

So for the critical ones:

- First: valid but doesn't matter (as we are just adding validation issues to the invoice), fixed regardlesss
- Second: That's valid but the fix is not DeletedAt != nil -> IsVoided, but rather removing that altogether, as if a payment event comes we still need to book that regardless of line status :confused:


For the test coverage issue:

adding tests, but it was the behavior before

tothandras (approved):

PR #4344, opened 2026-05-11
fix: properly soft delete app custom invoicing customer associations upon customer delete [OM-357]

gergely-kurucz-konghq:
## Overview

Fix soft-delete unique constraint for app custom invoicing customers

### What

- Converts the unique index on `app_custom_invoicing_customers` to a **partial unique index** scoped to non-deleted rows (`WHERE deleted_at IS NULL`).
- Updates `UpsertCustomerData` to use `ON CONFLICT ... WHERE (deleted_at IS NULL)` to match the partial index.
- Adds `DeletedAtIsNil()` filter to `DeleteCustomerData` for consistency.
- Extends `customer.DeleteCustomer` to soft-delete `app_custom_invoicing_customers` rows when a customer is deleted.
- Extends the customer adapter test suite to cover the new cascade.

### Why

Mirrors the fix applied to `app_customers` and `app_stripe_customers` in the previous PR (#4325). Soft-deleted `app_custom_invoicing_customers` records were blocking re-creation of the same customer–app association after deletion, because the unique index covered all rows including soft-deleted ones.

### How

- Migration drops the existing full-table unique index on `(namespace, app_id, customer_id)` and recreates it as a partial index filtered to `deleted_at IS NULL`.
- Ent schema updated with `entsql.IndexWhere("deleted_at IS NULL")` + `make generate` run.
- Upsert switches from `OnConflictColumns(...)` to `OnConflict(sql.ConflictColumns(...), sql.ConflictWhere(sql.IsNull(...)))`.
- `customer.DeleteCustomer` now soft-deletes associated `app_custom_invoicing_customers` rows in the same transaction.
- Tests: `seed()` extended with `AppCustomInvoicing` + `AppCustomInvoicingCustomer` rows; assertions added to `Cascade_AllChildren`, `PreservesAlreadyDeletedChildren`, and `DifferentNamespaceIsolation`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced customer deletion logic to properly cascade soft-delete operations across all related custom invoicing records and ensure data consistency.
  * Refined database constraints to enforce uniqueness only on active records, preventing conflicts when handling soft-deleted data while maintaining referential integrity.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openmeterio/openmeter/pull/4344)

### DIFF
--- a/openmeter/customer/adapter/customer.go
+++ b/openmeter/customer/adapter/customer.go
@@ -12,6 +12,7 @@ import (
 	"github.com/openmeterio/openmeter/openmeter/customer"
 	entdb "github.com/openmeterio/openmeter/openmeter/ent/db"
 	appcustomerdb "github.com/openmeterio/openmeter/openmeter/ent/db/appcustomer"
+	appcustominvoicingcustomerdb "github.com/openmeterio/openmeter/openmeter/ent/db/appcustominvoicingcustomer"
 	appstripecustomerdb "github.com/openmeterio/openmeter/openmeter/ent/db/appstripecustomer"
 	billingcustomeroverridedb "github.com/openmeterio/openmeter/openmeter/ent/db/billingcustomeroverride"
 	customerdb "github.com/openmeterio/openmeter/openmeter/ent/db/customer"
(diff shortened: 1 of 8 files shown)


openmeter/customer/adapter/customer.go
```diff
@@ -409,6 +410,17 @@ func (a *adapter) DeleteCustomer(ctx context.Context, input customer.DeleteCusto
 			return fmt.Errorf("failed to delete app stripe customer: %w", err)
 		}
 
+		// Soft delete the app custom invoicing customer associations
+		err = repo.db.AppCustomInvoicingCustomer.Update().
```
borosr, line 414: Could you please extend the customer_test.go to cover this case, too?

borosr (changes requested):

openmeter/customer/adapter/customer.go
gergely-kurucz-konghq: Fixed in latest commit.

tothandras (dismissed):

borosr (dismissed):
LGTM

borosr (approved):

PR #4358, opened 2026-05-13
feat(sdk): init Go, Python and TypeScript SDKs

tothandras:


api/spec/packages/aip-client-go/models/components/sortquery.go
```diff
@@ -0,0 +1,23 @@
...
+
+package components
+
+import "github.com/openmeterio/openmeter/api/spec/packages/aip-client-go/internal/utils"
+
+// Sort query.
+//
+// The `asc` suffix is optional as the default sort order is ascending. The `desc`
+// suffix is used to specify a descending order.
+type SortQuery struct {
```
borosr, line 11: This looks incorrect to me, since the sort should be string type.

api/spec/packages/aip-client-go/smoke_test.go
```diff
@@ -0,0 +1,72 @@
...
+			"meta": map[string]any{
+				"page": map[string]any{"size": 1, "number": 1, "total": 1},
+			},
+		})
+	}))
+	defer srv.Close()
+
+	sdk := sdkpkg.New(sdkpkg.WithServerURL(srv.URL))
+	ctx := context.Background()
+	res, err := sdk.OpenMeterMeters.ListMeters(ctx, operations.ListMetersRequest{})
```
borosr, line 62: The method naming is redundant, can you please change it? Either way would be great, like: `sdk.OpenMeterMeters.List(...)` or `sdk.OpenMeter.ListMeters(...)`

I personally like the first option a little bit more.

api/spec/packages/aip-client-go/aipclientgo.go
```diff
@@ -0,0 +1,215 @@
...
+	"https://us.api.konghq.com/v3",
+}
+
+// HTTPClient provides an interface for supplying the SDK with a custom HTTP client
+type HTTPClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
+// String provides a helper function to return a pointer to a string
+func String(s string) *string { return &s }
```
borosr, line 42: This part is redundant, since the types package already has a pointer helper file. Can you please use that instead of these?

api/spec/packages/aip-client-go/aipclientgo.go
```diff
@@ -0,0 +1,215 @@
...
+	sdk := &OpenMeter{
+		SDKVersion: "0.0.1",
+		sdkConfiguration: config.SDKConfiguration{
+			UserAgent:  "aipclientgo-go/0.0.1",
+			ServerList: ServerList,
+			ServerVariables: []map[string]string{
+				{
+					"port": "8888",
+				},
+				{},
```
borosr, line 161: Can you please remove the empty map instances from here?

api/spec/packages/aip-client-go/openmeteraddons.go
```diff
@@ -0,0 +1,728 @@
...
+	} else {
+		baseURL = *o.ServerURL
+	}
+
+	opURL, err := url.JoinPath(baseURL, "/openmeter/addons")
+	if err != nil {
+		return nil, fmt.Errorf("error generating URL: %w", err)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, "GET", opURL, nil)
```
borosr, line 59: Can you please modify the skill to use http.MethodGet and other constants for method setup?

api/spec/packages/aip-client-go/openmeteraddons.go
```diff
@@ -0,0 +1,728 @@
...
+		},
+	}
+
+	switch httpRes.StatusCode {
+	case 200:
+		body, err := io.ReadAll(httpRes.Body)
+		if err != nil {
+			return nil, err
+		}
+		var out map[string]any
```
borosr, line 89: This way the response value would be nearly typeless, which is not a good thing, can you please force it to unmarshal into the real response format? Also, this is not a very optimal way for parsing an io.Reader, please use the `json.NewDecoder(httpRes.Body).Decode(res.Body)` instead.

api/spec/packages/aip-client-go/models/operations/listaddons.go
```diff
@@ -0,0 +1,16 @@
...
+
+type ListAddonsRequest struct {
+	Page   map[string]any                     `queryParam:"name=page,style=deepObject,explode=true"`
+	Sort   *components.SortQuery              `queryParam:"name=sort"`
+	Filter *components.ListAddonsParamsFilter `queryParam:"name=filter,style=deepObject,explode=true"`
+}
+
+type ListAddonsResponse struct {
+	HTTPMeta components.HTTPMetadata
+	Body     *map[string]any
```
borosr, line 15: Please force the generator to populate here the real ListAddon response format.

api/spec/packages/aip-client-go/smoke_test.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		t.Fatalf("ListMeters failed: %v", err)
+	}
+	if res == nil {
+		t.Fatal("ListMeters returned nil response")
+	}
+	if res.HTTPMeta.Response == nil {
```
borosr, line 69: Smoke test it not correct, please change it to check the type of the response body.

PR #4401, opened 2026-05-20
feat: introduce common interface for feature gate

borosr:
# What
Introduce common interface to handle feature gate requests.

# Why
We want to control specific cases using feature gate.

## Summary by CodeRabbit

* **New Features**
  * Added feature-gating support with a default no-op gate for safe rollout.
  * Introduced a comprehensive feature-gate interface supporting typed flag evaluations (bool/int/float/string/json) and context/org scoping.
  * Billing worker and API server now accept and propagate feature-gate settings so features can be toggled at runtime.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openmeterio/openmeter/pull/4401?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


turip (dismissed):

pkg/gatex/gatex.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+var instance FeatureGate = NoopFeatureGate{}
+
+func GetDefault() FeatureGate {
+	if instance == nil {
+		instance = NoopFeatureGate{}
+	}
+	return instance
+}
```
turip: Please instead use wire to inject the noop implementation. If you want to do such things sync.Once should be used, but the dependency injection should handle this.

openmeter/billing/worker/worker.go
```diff
@@ -106,12 +110,18 @@ func New(opts WorkerOptions) (*Worker, error) {
 		}
 	}
 
+	featureGate := opts.FeatureGate
+	if featureGate == nil {
+		featureGate = gatex.GetDefault()
+	}
```
turip: Throw an error if the FeatureGate is nil in the Config. Noop/defaulting should be handled by wire.

pkg/gatex/gatex.go
```diff
@@ -0,0 +1,75 @@
+package gatex
```
turip: nit: we only use the x prefix is there's a system/frequently used package that we have name collusion with. Gate is fine.

openmeter/billing/worker/worker.go
borosr: Thanks, done

pkg/gatex/gatex.go
borosr: Thanks, changed

chrisgacsal (dismissed):

chrisgacsal (dismissed):

turip (approved):

chrisgacsal (approved):

PR #4406, opened 2026-05-21
fix: ent reset list logic

GAlexIHU:
## Overview

TLDR we were returning more than we should have hence the log storm (the relevant change is only the extra filter, the TS comparison changes are just for semantic correctness)

Fixes #(issue)

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened entitlement filtering so expired usage periods are identified more accurately: activity is evaluated at the current time, results are limited to metered entitlements, current usage period end must be set and ≤ highwatermark, deletion is considered relative to now, and eligibility requires a measure anchor or a usage-reset.
* **Tests**
  * Updated tests to cover resettable metered entitlements, set the clock explicitly, and validate inclusion of reset-anchored items.

[Review Change Stack](https://app.coderabbit.ai/change-stack/openmeterio/openmeter/pull/4406?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/openmeter/entitlement/adapter/entitlement.go
+++ b/openmeter/entitlement/adapter/entitlement.go
@@ -771,11 +771,13 @@ func (a *entitlementDBAdapter) ListActiveEntitlementsWithExpiredUsagePeriod(ctx 
 			now := clock.Now()
 
 			query := withAllUsageResets(repo.db.Entitlement.Query(), params.Namespaces).
-				Where(EntitlementActiveAt(params.Highwatermark)...).
+				Where(EntitlementActiveAt(now)...).
 				Where(
+					db_entitlement.EntitlementTypeEQ(db_entitlement.EntitlementType(entitlement.EntitlementTypeMetered)),
 					db_entitlement.CurrentUsagePeriodEndNotNil(),
 					db_entitlement.CurrentUsagePeriodEndLTE(params.Highwatermark),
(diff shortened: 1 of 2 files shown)


openmeter/entitlement/adapter/entitlement.go
```diff
@@ -771,11 +771,13 @@ func (a *entitlementDBAdapter) ListActiveEntitlementsWithExpiredUsagePeriod(ctx
...
 
 			query := withAllUsageResets(repo.db.Entitlement.Query(), params.Namespaces).
-				Where(EntitlementActiveAt(params.Highwatermark)...).
+				Where(EntitlementActiveAt(now)...).
 				Where(
+					db_entitlement.EntitlementTypeEQ(db_entitlement.EntitlementType(entitlement.EntitlementTypeMetered)),
 					db_entitlement.CurrentUsagePeriodEndNotNil(),
 					db_entitlement.CurrentUsagePeriodEndLTE(params.Highwatermark),
 					db_entitlement.Or(db_entitlement.DeletedAtIsNil(), db_entitlement.DeletedAtGT(now)),
+					entitlementHasResetAnchor(),
```
tothandras: Could you inline this? Seems to be only used in a single place.

tothandras (approved):

PR #4468, opened 2026-06-01
fix(telemetry): instrument v3 API routes with tracing

gergely-kurucz-konghq:
## What

Instruments the v3 API with OpenTelemetry HTTP tracing/metrics (on par with v1), gives server spans meaningful low-cardinality names, and adds an optional per-operation child span behind a global toggle.

## Why

v3 routes had **no** otelhttp instrumentation (the v3 group was registered before the telemetry middleware), and server spans had empty/unhelpful names — making v3 traces hard to find and group. We also wanted handler-vs-middleware timing visibility without inflating span volume across the whole API.

## How

- **v3 instrumentation:** materialize the router-hook middlewares once and apply the same slice to both the v3 and v1 groups (hooks run once; no double side effects).
- **Span name:** `METHOD /route/{template}` from the chi route pattern (`r.Pattern` → chi `RouteContext` → fallback). Unresolved mount wildcards (`/api/v3/*`) are skipped.
- **`http.route` attribute + metric label:** the exact route template — the canonical low-cardinality routing dimension. `url.path` is path-only (no query string).
- **Cardinality bounding:** the fallback path masks ULIDs, UUIDs, numeric ids, and over-long opaque segments to `:id`, and truncates pathologically deep paths — so unmatched/scanner traffic can't blow up span-name cardinality.
- **Per-operation child span:** application-level span named after the operation (e.g. `list-customers`), nested under the server span, exposing the middleware-vs-handler split. **Off by default**, enabled via `telemetry.trace.operationSpans`, since it adds a span to every request.

> Note on otelhttp v0.68: the span name is set at span start (and re-set on a post-handler pass when `r.Pattern` is set), so naming is done in `WithSpanNameFormatter` rather than via a post-start `SetName` (a no-op in this chi setup).

## Summary by CodeRabbit

* **Improvements**
  * Enhanced OpenTelemetry instrumentation for richer request tracing and reduced span cardinality by masking dynamic path segments.
  * Normalized server span naming and improved URL-path attribution for clearer traces.
* **New Features**
  * Optional per-operation child spans (disabled by default) for more granular application-level tracing.
* **Improvements**
  * Consolidated middleware collection so routing groups share consistent middleware.
* **Tests**
  * Added tests validating path masking and truncation behavior.

### DIFF
--- a/app/common/telemetry.go
+++ b/app/common/telemetry.go
@@ -6,6 +6,7 @@ import (
 	"log/slog"
 	"net/http"
 	"os"
+	"strings"
 	"time"
 
 	health "github.com/AppsFlyer/go-sundheit"
@@ -14,6 +15,7 @@ import (
 	"github.com/go-chi/chi/v5/middleware"
 	"github.com/go-slog/otelslog"
 	"github.com/google/wire"
+	"github.com/oklog/ulid/v2"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/collectors"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -278,21 +280,53 @@ func NewTelemetryRouterHook(meterProvider metric.MeterProvider, tracerProvider t
 
 					routePattern := chi.RouteContext(r.Context()).RoutePattern()
 
+					// Record the route template as http.route on the span and request
+					// metrics — the canonical low-cardinality routing dimension. The span
+					// *name* is set separately by WithSpanNameFormatter below: otelhttp
+					// fixes the name at tracer.Start and a post-start SetName here is a
+					// no-op in this setup, so we cannot rename the span from this point.
 					span := trace.SpanFromContext(r.Context())
-					span.SetName(routePattern)
 					span.SetAttributes(semconv.URLPath(r.URL.String()), semconv.HTTPRoute(routePattern))
 
-					labeler, ok := otelhttp.LabelerFromContext(r.Context())
-					if ok {
+					if labeler, ok := otelhttp.LabelerFromContext(r.Context()); ok {
 						labeler.Add(semconv.HTTPRoute(routePattern))
 					}
 				}),
 				"",
 				otelhttp.WithMeterProvider(meterProvider),
 				otelhttp.WithTracerProvider(tracerProvider),
+				otelhttp.WithSpanNameFormatter(func(_ string, r *http.Request) string {
+					// Name the server span "METHOD /route/{template}". chi's route pattern
+					// is the exact, low-cardinality template (path params appear as
+					// "{param}", so both ids and arbitrary keys collapse). It is already
+					// resolved by the time otelhttp calls this formatter. Fall back to the
+					// ULID-collapsed raw path when no route matched (e.g. 404s).
+					name := lowCardinalityPath(r.URL.Path)
+					if rctx := chi.RouteContext(r.Context()); rctx != nil {
+						if pat := rctx.RoutePattern(); pat != "" {
+							name = pat
+						}
+					}
+					return r.Method + " " + name
+				}),
 			)
 		})
 	}
+}
+
+// lowCardinalityPath collapses ULID path segments to ":id" so span names stay
+// bounded. OpenMeter resource IDs are ULIDs, so this keeps parameterized routes
+// (e.g. "/api/v3/openmeter/customers/01KFTS.../billing") from exploding span-name
+// cardinality, yielding "/api/v3/openmeter/customers/:id/billing". The exact route
+// template is carried separately in the http.route attribute.
+func lowCardinalityPath(path string) string {
+	segments := strings.Split(path, "/")
+	for i, seg := range segments {
+		if _, err := ulid.ParseStrict(seg); err == nil {
+			segments[i] = ":id"
+		}
+	}
+	return strings.Join(segments, "/")
 }
 
 type RuntimeMetricsCollector struct{}
--- a/app/common/telemetry.go
+++ b/app/common/telemetry.go
@@ -286,7 +286,7 @@ func NewTelemetryRouterHook(meterProvider metric.MeterProvider, tracerProvider t
 					// fixes the name at tracer.Start and a post-start SetName here is a
 					// no-op in this setup, so we cannot rename the span from this point.
 					span := trace.SpanFromContext(r.Context())
-					span.SetAttributes(semconv.URLPath(r.URL.String()), semconv.HTTPRoute(routePattern))
+					span.SetAttributes(semconv.URLPath(r.URL.Path), semconv.HTTPRoute(routePattern))
 
 					if labeler, ok := otelhttp.LabelerFromContext(r.Context()); ok {
 						labeler.Add(semconv.HTTPRoute(routePattern))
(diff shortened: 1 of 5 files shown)


app/common/telemetry.go
```diff
@@ -278,23 +280,55 @@ func NewTelemetryRouterHook(meterProvider metric.MeterProvider, tracerProvider t
...
+					// resolved by the time otelhttp calls this formatter. Fall back to the
+					// ULID-collapsed raw path when no route matched (e.g. 404s).
+					name := lowCardinalityPath(r.URL.Path)
+					if rctx := chi.RouteContext(r.Context()); rctx != nil {
+						if pat := rctx.RoutePattern(); pat != "" {
+							name = pat
+						}
+					}
+					return r.Method + " " + name
+				}),
```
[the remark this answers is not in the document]
gergely-kurucz-konghq: We verified the opposite with logs + Tempo. At formatter time RoutePattern() returns the full template (/api/v1/features/{featureId}), and the exported span name is that template. chi's nested-router descent populates the pattern before otelhttp's inner tracer.Start.

We proved a post-start SetName on the otelhttp server span is a no-op in this chi setup (SetAttributes sticks, SetName doesn't). That suggestion would produce empty names — it's literally the approach that failed. The formatter is the only reliable naming point, which is why it's used.

lowCardinalityPath is only the fallback for unmatched routes (404s). Matched routes use the chi template, which already renders every path param as {param} — ids and keys. For a 404 no route matched, so no param semantics exist to collapse; ULID-collapse is the only sensible heuristic for scanner noise. Nothing to fix.

> "use r.URL.Path not r.URL.String()"

Fixed in the prior change.

app/common/telemetry.go
```diff
@@ -278,23 +280,55 @@ func NewTelemetryRouterHook(meterProvider metric.MeterProvider, tracerProvider t
...
 				}),
 				"",
 				otelhttp.WithMeterProvider(meterProvider),
 				otelhttp.WithTracerProvider(tracerProvider),
+				otelhttp.WithSpanNameFormatter(func(_ string, r *http.Request) string {
+					// Name the server span "METHOD /route/{template}". chi's route pattern
+					// is the exact, low-cardinality template (path params appear as
+					// "{param}", so both ids and arbitrary keys collapse). It is already
+					// resolved by the time otelhttp calls this formatter. Fall back to the
+					// ULID-collapsed raw path when no route matched (e.g. 404s).
```
mark-vass-konghq: clarify the span-name comment. In otelhttp v0.68.0, the formatter runs once at span start and again after the handler only when r.Pattern is set. The route is not fully resolved
    on the first call, so the comment saying it is “already resolved” is misleading. Prefer checking r.Pattern first in the formatter and documenting the two-pass behavior.

pkg/framework/transport/httptransport/handler.go
```diff
@@ -79,7 +86,15 @@ func (h handler[Request, Response]) ServeHTTP(w http.ResponseWriter, r *http.Req
...
+		name := h.operationNameFunc(ctx)
+		ctx = contextx.WithAttr(ctx, string(semconv.HTTPRouteKey), name)
+
+		// Start an application-level span named after the operation, as a child of
+		// the otelhttp server span. The server span stays route-named; this one
+		// carries the operation identity (e.g. "query-governance-access").
+		var span trace.Span
+		ctx, span = tracer.Start(ctx, name)
+		defer span.End()
 	}
```
mark-vass-konghq, line 116: this adds an operation child span to every httptransport handler with WithOperationName, not only v3. That likely doubles span volume for many v1/v2 paths too. Confirm that telemetry cost/blast radius is intended, or make operation spans opt-in/configurable.

openmeter/server/server.go
```diff
@@ -92,7 +92,9 @@ func NewServer(config *Config) (*Server, error) {
 	r := chi.NewRouter()
 	r.Use(server.NewPoweredByMiddleware())
 
-	v3Middlewares := []server.MiddlewareFunc{
+	// Collect router-hook middlewares (e.g. otelhttp tracing/metrics) so v3
+	// has the same OTEL HTTP instrumentation as the v1 router group.
+	v3Middlewares := append(collectMiddlewareHooks(config.RouterHooks.Middlewares), []server.MiddlewareFunc{
```
mark-vass-konghq: invokes middleware hooks once for v3, then the v1 group invokes the same hook functions again. Current telemetry hook is fine, but future hooks with
    side effects would run twice. Safer option: materialize hook middlewares once and apply the same slice to both v3 and v1

openmeter/server/server.go
mark-vass-konghq: invokes middleware hooks once for v3, then the v1 group invokes the same hook functions again. Current telemetry hook is fine, but future hooks with side effects would run twice. Safer option: materialize hook middlewares once and apply the same slice to both v3 and v1

app/common/telemetry.go
```diff
@@ -278,23 +280,55 @@ func NewTelemetryRouterHook(meterProvider metric.MeterProvider, tracerProvider t
...
 		})
 	}
 }
 
+// lowCardinalityPath collapses ULID path segments to ":id" so span names stay
+// bounded. OpenMeter resource IDs are ULIDs, so this keeps parameterized routes
+// (e.g. "/api/v3/openmeter/customers/01KFTS.../billing") from exploding span-name
+// cardinality, yielding "/api/v3/openmeter/customers/:id/billing". The exact route
+// template is carried separately in the http.route attribute.
+func lowCardinalityPath(path string) string {
```
mark-vass-konghq, line 357: only masks ULID path segments. For unmatched routes, arbitrary slugs/keys can still leak into fallback span names. Consider a fixed unmatched fallback, broader ID masking, or segment count/length caps.

mark-vass-konghq (commented):
I didn’t find a clear merge-blocking bug. Possible improvements I’d consider

pkg/framework/transport/httptransport/handler.go
gergely-kurucz-konghq: It is intended, as with this addition it is possible to differentiate between middleware execution time vs actual handler execution time (the difference of the outer request span and the operation child span).

I considered the cost/blast radius negligible. My current understanding is that at most this adds 1 more span to each HTTP request trace, which is the operation span wrapping the handler logic itself. Nothing is duplicated (well, if a v1/v2 endpoint for some reason only provided a single span, then it technically will be doubled).

Regardless, making this new operation span opt-in might be a good idea, let me think about it.

mark-vass-konghq (approved):

app/common/telemetry.go
```diff
@@ -278,23 +285,127 @@ func NewTelemetryRouterHook(meterProvider metric.MeterProvider, tracerProvider t
...
+					// a usable route name, so skip it (from any source) and fall back to the
+					// ULID-collapsed path.
+					route := r.Pattern
+					if route == "" {
+						if rctx := chi.RouteContext(r.Context()); rctx != nil {
+							route = rctx.RoutePattern()
+						}
+					}
+					if route == "" || strings.Contains(route, "*") {
+						route = lowCardinalityPath(r.URL.Path)
```
turip, line 331: qq: why don't we just add the final path as an attribute instead?

gergely-kurucz-konghq: It is already added, see line 294. This is the span name, which was completely missing since the previous implementation of setting the span name did not work.

turip (approved):

PR #4470, opened 2026-06-01
Fix/settlement mode api

rolosp:
## Summary by CodeRabbit

* **New Features**
  * Optional settlement mode for subscription create/change (credit-only or credit-then-invoice) with request-level override.
  * Feature flag control for credits enabling—platform can block credit-only operations when disabled.

* **Refactor**
  * Redesigned feature-gate wiring and middleware across APIs and services for more consistent, reliable feature-flag evaluation.


turip:
LGFM please fix the etoe failures

turip:
Also did you check if the settmenet mode is exposed on subscription reads for the v3 api?

api/v3/handlers/subscriptions/featuregate.go
```diff
@@ -0,0 +1,15 @@
+package subscriptions
+
+func (h *handler) isCreditsEnabled(ns string) (bool, error) {
+	if !h.credits.Enabled {
+		return false, nil
+	}
+	if h.featureGate == nil {
+		return true, nil
```
GAlexIHU: why enabled by default?

openmeter/productcatalog/subscription/http/change.go
```diff
@@ -118,6 +118,20 @@ func (h *handler) ChangeSubscription() ChangeSubscriptionHandler {
...
+					settlementMode = lo.ToPtr(productcatalog.SettlementMode(*parsedBody.SettlementMode))
+				}
+
+				creditEnabled, err := h.isCreditsEnabled(ns)
+				if err != nil {
+					return ChangeSubscriptionRequest{}, fmt.Errorf("failed to change subscription: %w", err)
+				}
+
+				if !creditEnabled && lo.FromPtr(settlementMode) == productcatalog.CreditOnlySettlementMode {
+					return ChangeSubscriptionRequest{}, models.NewGenericValidationError(fmt.Errorf("credits are not enabled on this deployment of OpenMeter"))
```
GAlexIHU: thinking out loud: shouldn't this validation also run on subscription internal ops during general validate? (idk if it runs there or not didn't check). there isn't a non-api control path currently but generally i'd prefer business validations in domain layers not api layer...

e2e/productcatalog_test.go
```diff
@@ -1321,4 +1325,53 @@ func TestSettlementMode(t *testing.T) {
 		require.NotNil(t, res.JSON200)
 		assert.Equal(t, lo.ToPtr(api.BillingSettlementModeCreditThenInvoice), res.JSON200.SettlementMode)
 	})
+
+	t.Run("Should reject a plan-based subscription with credit_only settlement mode when credit is disabled", func(t *testing.T) {
```
GAlexIHU, line 1367: do we have a test that asserts the settlementmode "overwrite" working?

GAlexIHU (approved):
good job

api/v3/handlers/subscriptions/featuregate.go
borosr: By default if no feature gate available the features should be managed from the config vars or file. This is the reason of the default allow logic.

GAlexIHU (changes requested):
sry, actually if the settlement_mode = credit_only comes from the plan the check is bypassed i think

GAlexIHU (approved):

PR #4473, opened 2026-06-03
feat: add v3 operations to v1 client

rolosp:


rolosp:
this was a fallback solution and we have a better SDK solution now for v3 we can close this PR

PR #4481, opened 2026-06-04
feat(credits): add feature filters to ledger credits

GAlexIHU:
# Summary

- add feature filters to credit purchases and carry them through ledger routing, collection, correction, and lineage
- persist feature filters and advance lineage feature dimensions with generated Ent code and migrations
- ❗ add a temporary create-time restriction for feature-filtered credit grants until customer balance semantics are defined

## Summary by CodeRabbit

* **New Features**
  * Enabled feature-filtered credits: restrict credits to specific metered features and route them to matching billing sub-accounts.
  * Collection, advance backfill and settlement flows now respect feature restrictions during allocation and routing.

* **Behavior Changes**
  * Creation of credit grants with feature filters is currently rejected (validation error) until balance semantics are defined.

* **Documentation**
  * Clarified guidance for Postgres text[] array handling.

* **Tests / Migrations**
  * Added extensive tests and DB migrations to support feature filters.


openmeter/ledger/chargeadapter/creditpurchase.go
```diff
@@ -204,8 +209,9 @@ func (h *creditPurchaseHandler) issueCreditPurchaseGroup(ctx context.Context, ch
 		ID:        charge.Intent.CustomerID,
 	}
 	annotations := chargeAnnotationsForCreditPurchaseCharge(charge)
+	featureFilters := charge.Intent.FeatureFilters.Strings()
```
turip:
```suggestion
	featureFilters := charge.Intent.FeatureFilters.Normalize()
```

I don't like that Strings actually also normalizes. FeatureFilters will be usable as []string input.

Let's drop the Strings() member, as the type itself or .Normalize() output is fully interchangeable with []string

tools/migrate/migrations/atlas.sum
```diff
@@ -206,3 +206,5 @@ h1:MUO8x7HjdEQ+JK6ZMubbAOj7wEPcy/LzifR3e+QPBg8=
 20260520130000_repair_rebased_migrations.up.sql h1:XJhbK1CCplq5BEJi+VvHDVAcLUBJqHNF1LFGhaSa2Oo=
 20260520130500_add_ledger_tax_behavior.up.sql h1:M8Lq8lE/O+RwPLJ/7qe5tVddruuAxOcBAeJ1usFEveg=
 20260527120000_dedupe_tax_codes_by_app_mapping.up.sql h1:/tOMj6sA5vrD6+xJLZXeJaxnY4fJoxhfDGzRCPqWeEA=
+20260601160313_add_credit_feature_filters.up.sql h1:DmZ47ZlkfHYhhF/z7zNGtmFvBFpG0aQ7SuR66ImEcNo=
+20260603141000_add_credit_lineage_advance_features.up.sql h1:bzApRhx4a6Y3Sixrl6AO2I+qXz0+Aab6SSFWrhx6rS0=
```
turip: May I ask you to merge these two pretty please?

tools/migrate/migrations/20260603141000_add_credit_lineage_advance_features.up.sql
```diff
@@ -0,0 +1,2 @@
+-- modify "credit_realization_lineages" table
+ALTER TABLE "credit_realization_lineages" ADD COLUMN "advance_features" jsonb NULL;
```
turip: I know postgresql []string is a pain, but shouldn't we use that instead? GIN indexes are not composable afaik.

This is more like a question, so I can live with jsonb.

GAlexIHU: yeah i was also surprised when saw this in the generation but this is the ent default (didnt bother to check how we do elsewhere but i assume we do this) *i did check now, we don't actually use field.Strings() anywhere, updated repo guidance on it... thx for making me check agian 👍

as for GIN: i think btree would do practically nothing on []text, it would only be good for identity checks (&hence sorting) i think you'd need GIN for membership checks either way... but still, lets follow our patterns :D

turip (dismissed):

turip (approved):

PR #4522, opened 2026-06-12
perf(governance): resolve customer keys in one bulk query [OM-387]

gergely-kurucz-konghq:
## What

Optimize the governance access query (`POST /api/v3/openmeter/governance/query`) by resolving customer keys in a single bulk query instead of one DB round trip per key.

## Why

`resolveCustomers` looped `GetCustomerByUsageAttribution` once per customer key — N serial DB round trips over the **full** key set (before pagination, so page size didn't help). The e2e benchmark showed this made the customers axis linear; it was the dominant cost at high customer counts.

## How

- Add `customer.GetCustomersByUsageAttribution(ctx, {Namespace, Keys})` on the customer `Service` + `Adapter`: **one** query matching `customer.key IN keys OR subject_key IN keys` (mirrors the single-key `GetCustomerByUsageAttribution` semantics), subjects eager-loaded, soft-deleted rows filtered. Returns `[]Customer` — multi-get-by-set convention, not paginated.
- Rewrite governance `resolveCustomers` to call it once and reconstruct `matched[]` / customer-not-found errors in memory from each customer's key + subject keys. Result shape unchanged → sort → paginate → `resolveAccess` untouched; behavior preserved.

Collapses N customer-resolution queries to 1.

**Benchmark (p50, baseline → new):** `100×10` −21%, whole `C=10` row −17…−29%. High-feature cells flat (`100×100` ≈0) — bounded by the still-serial `resolveAccess` `GetAccess` loop, a deliberate follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added bulk customer resolution by usage-attribution keys, returning deduplicated matches and supporting lookups via both customer keys and active subject keys.
* **Refactor**
  * Updated governance access customer resolution to resolve the full key set in one bulk call, then map and report per-key results in-memory.
* **Tests**
  * Added adapter/service coverage for mixed-key resolution, deduplication, namespace isolation, soft-deleted exclusions, and input validation.
* **Documentation**
  * Refreshed pagination/phase notes for the new single-bulk-lookup resolution flow.

### DIFF
--- a/openmeter/customer/adapter/customer_test.go
+++ b/openmeter/customer/adapter/customer_test.go
@@ -526,17 +526,6 @@ func TestGetCustomersByUsageAttribution(t *testing.T) {
 		require.NoError(t, err)
 		assert.ElementsMatch(t, []string{idA}, customerIDs(got), "must only return the customer in the queried namespace")
 	})
-
-	t.Run("ValidationError", func(t *testing.T) {
-		env := newTestEnv(t)
-		ns := ulid.Make().String()
-
-		_, err := env.adapter.GetCustomersByUsageAttribution(t.Context(), customer.GetCustomersByUsageAttributionInput{
-			Namespace: ns,
-			Keys:      nil,
-		})
-		require.Error(t, err, "empty key set must fail validation")
-	})
 }
 
 // --- assertion helpers ---
--- a/openmeter/customer/adapter/customer_test.go
+++ b/openmeter/customer/adapter/customer_test.go
@@ -358,6 +358,188 @@ func TestDeleteCustomer(t *testing.T) {
 	})
 }
 
+// seedCustomerWithKey creates an active customer with the given key and subject keys.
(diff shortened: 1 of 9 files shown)


openmeter/customer/adapter/customer_test.go
```diff
@@ -358,6 +358,188 @@ func TestDeleteCustomer(t *testing.T) {
...
+	return cust.ID
+}
+
+func customerIDs(customers []customer.Customer) []string {
+	ids := make([]string, 0, len(customers))
+	for _, c := range customers {
+		ids = append(ids, c.ID)
+	}
+	return ids
+}
```
chrisgacsal, line 392: Use something like:

```
lo.Map(customers []customer.Customer, func(customer customer.Customer, _ int) string {
     return customer.ID
})
```

borbelyr-kong (dismissed):

borbelyr-kong (approved):

PR #4539, opened 2026-06-18
fix(taxcode): guard deletion of referenced tax codes and tolerate soft-deleted references

borbelyr-kong:


borbelyr-kong:
@greptileai review this draft

PR #4541, opened 2026-06-18
feat(api): add invoice V3 GET endpoint

borosr:
## Overview

Fixes #(issue)

## Summary by CodeRabbit

* **New Features**
  * Added billing invoice retrieval to the JavaScript SDK, including a new `Invoices` surface and an `invoices.get` operation to fetch invoices by ID.
  * Expanded invoice-related models/types across the SDK to support workflow, status/validation details, and invoice/line/discount structures.
  * Introduced a v3 “billing invoices” GET by ID endpoint with new end-to-end test coverage.
* **Bug Fixes**
  * Improved v3 invoice conversion behavior and invoice error encoding (including validation and not-found handling).
* **Documentation**
  * Updated the JavaScript SDK README to document the new Invoices resource and operation.


api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,1105 @@ namespace Invoices;
...
+
+  /**
+   * The service period covered by this invoice.
+   *
+   * Optional because some invoice types (e.g. immediate flat fees) do not have a
+   * meaningful service period.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Service period")
+  service_period?: Shared.ClosedPeriod;
```
turip: required

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,1105 @@ namespace Invoices;
...
+ */
+@friendlyName("BillingInvoiceCustomerUsageAttribution")
+model BillingInvoiceCustomerUsageAttribution {
+  /**
+   * The subject keys whose usage is attributed to this invoice.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Subject keys")
+  subject_keys: string[];
+}
```
turip: don't we have something reusable on customer.

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,1105 @@ namespace Invoices;
...
+
+/**
+ * Severity level of an invoice validation issue.
+ */
+@friendlyName("BillingInvoiceValidationIssueSeverity")
+enum BillingInvoiceValidationIssueSeverity {
+  /**
+   * Default zero value. Reserved for forward compatibility.
+   */
+  Unknown: "unknown",
```
turip: Do we want to do that or just introduce new values as needed? (Asking) Is this something we are doing in other places of the system.

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,1105 @@ namespace Invoices;
...
+   * The line was created manually.
+   */
+  Manual: "manual",
+}
+
+/**
+ * Lifecycle status of an invoice line item.
+ */
+@friendlyName("BillingInvoiceLineStatus")
+enum BillingInvoiceLineStatus {
```
turip: let's drop this line status altogether, this should be removed.

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,1105 @@ namespace Invoices;
...
+  @visibility(Lifecycle.Read)
+  @summary("Managed by")
+  managed_by: BillingInvoiceLineManagedBy;
+
+  /**
+   * Current status of the line item.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Status")
+  status: BillingInvoiceLineStatus;
```
turip: drop

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,1105 @@ namespace Invoices;
...
+
+  /**
+   * Three-letter ISO 4217 currency code for the line item.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Currency")
+  currency: Shared.CurrencyCode;
```
turip: drop for now, it's on the invoice

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,1105 @@ namespace Invoices;
...
+  @visibility(Lifecycle.Read)
+  @summary("Service period")
+  service_period: Shared.ClosedPeriod;
+
+  /**
+   * The timestamp at which this line item is to be included in an invoice.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Invoice at")
+  invoice_at: Shared.DateTime;
```
turip: drop, this is gathering invoice only

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,1105 @@ namespace Invoices;
...
+ * Rate card configuration snapshot for a usage-based invoice line.
+ */
+@friendlyName("BillingInvoiceLineRateCard")
+model BillingInvoiceLineRateCard {
+  /**
+   * The price definition used to calculate charges for this line.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Price")
+  price?: ProductCatalog.Price;
```
turip: required

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,1105 @@ namespace Invoices;
...
+  surcharge?: Shared.Numeric;
+
+  /**
+   * Whether the tax is inclusive (already included in the price) or exclusive (added
+   * on top of the price).
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Behavior")
+  behavior?: Billing.TaxBehavior;
+}
```
turip: let's drop this altogether, tax is not yet supported, we will add tax calculation details when we have that codepath ready.

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,1105 @@ namespace Invoices;
...
+  @visibility(Lifecycle.Read)
+  @summary("Managed by")
+  managed_by: BillingInvoiceLineManagedBy;
+
+  /**
+   * Current status of the detailed line.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Status")
+  status: BillingInvoiceLineStatus;
```
turip: drop

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,1105 @@ namespace Invoices;
...
+  @visibility(Lifecycle.Read)
+  @summary("Service period")
+  service_period: Shared.ClosedPeriod;
+
+  /**
+   * The timestamp at which this detailed line is to be included in an invoice.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Invoice at")
+  invoice_at: Shared.DateTime;
```
turip: drop

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,1105 @@ namespace Invoices;
...
+  @summary("Price")
+  price?: ProductCatalog.Price;
+
+  /**
+   * The payment term for this detailed line (in advance or in arrears).
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Payment term")
+  payment_term?: ProductCatalog.PricePaymentTerm;
+}
```
turip: let's have inline qty and per unit amount in the parent type. Let's not do payment term.

api/spec/packages/aip/src/invoices/invoice.tsp
borosr: I found a few places, but those are mostly *Reason enums. Let me remove this, and keep only the possible values.

api/spec/packages/aip-client-javascript/src/sdk/invoices.ts
```diff
@@ -0,0 +1,18 @@
...
+import { getBillingInvoice } from '../funcs/invoices.js'
+import type {
+  GetBillingInvoiceRequest,
+  GetBillingInvoiceResponse,
+} from '../models/operations/invoices.js'
+
+export class Invoices {
+  constructor(private readonly _client: Client) {}
+
+  async getBilling(
```
tothandras: We need to patch the generator, billing should be dropped (`invoices.get()`)

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,1105 @@ namespace Invoices;
 @maxLength(256)
 @friendlyName("BillingInvoiceNumber")
 scalar InvoiceNumber extends string;
+
+// ============================================================================
+// Invoice top-level discriminated union
+// ============================================================================
```
tothandras:
```suggestion

```

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,1105 @@ namespace Invoices;
...
+ * - `credit_note`: a credit note reducing a previous invoice amount.
+ */
+@friendlyName("BillingInvoice")
+@discriminated(#{ discriminatorPropertyName: "type", envelope: "none" })
+union BillingInvoice {
+  @summary("Standard invoice")
+  standard: BillingStandardInvoice,
+
+  @summary("Credit note invoice")
+  credit_note: BillingCreditNoteInvoice,
```
tothandras:
```suggestion
  // @summary("Credit note invoice")
  // credit_note: BillingCreditNoteInvoice,
```

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,1105 @@ namespace Invoices;
...
+/**
+ * The type of a billing invoice.
+ */
+@friendlyName("BillingInvoiceType")
+enum BillingInvoiceType {
+  /**
+   * Default zero value. Reserved for forward compatibility.
+   */
+  Unknown: "unknown",
+
```
tothandras:
```suggestion
```

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,1105 @@ namespace Invoices;
...
+  /**
+   * A standard invoice for charges owed by the customer.
+   */
+  Standard: "standard",
+
+  /**
+   * A credit note reducing the amount owed on a previous invoice.
+   */
+  CreditNote: "credit_note",
```
tothandras:
``` suggestion
  //
  // /**
  //  * A credit note reducing the amount owed on a previous invoice.
  //  */
  // CreditNote: "credit_note",
```

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,1105 @@ namespace Invoices;
...
+}
+
+// ============================================================================
+// Common Invoice Header (shared base fields)
+// ============================================================================
+
```
tothandras:
```suggestion
```

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,1105 @@ namespace Invoices;
...
+/**
+ * Indicates what entity manages the lifecycle of an invoice line.
+ */
+@friendlyName("BillingInvoiceLineManagedBy")
+enum BillingInvoiceLineManagedBy {
+  /**
+   * Default zero value. Reserved for forward compatibility.
+   */
+  Unknown: "unknown",
```
tothandras: We should drop unknown everywhere

api/spec/packages/aip-client-javascript/src/sdk/invoices.ts
borosr: what do you think about changing the operation id from `get-billing-invoice` to `get-invoice`? that solves the problem without patching.

tothandras: Yes, do that

api/v3/handlers/billinginvoices/convert.go
```diff
@@ -0,0 +1,435 @@
...
+		if err != nil {
+			return out, err
+		}
+
+		if err := out.FromBillingStandardInvoice(stdAPI); err != nil {
+			return out, fmt.Errorf("setting standard invoice union: %w", err)
+		}
+
+	default:
+		return out, fmt.Errorf("unsupported invoice type %q", inv.Type())
```
turip: we can consider nillable validation error here.

turip: Or return 404, but we should not return an internal server error.

api/v3/handlers/billinginvoices/convert.go
```diff
@@ -0,0 +1,435 @@
...
+	if err != nil {
+		return nil, "", fmt.Errorf("mapping price: %w", err)
+	}
+
+	rc := &api.BillingInvoiceLineRateCard{
+		Price:      price,
+		FeatureKey: lo.EmptyableToPtr(line.UsageBased.FeatureKey),
+		Discounts:  toAPIRateCardDiscounts(line.RateCardDiscounts),
+		TaxConfig:  addons.ToAPIBillingRateCardTaxConfig(line.TaxConfig.ToProductCatalog()),
+	}
```
turip, line 325: `BillingInvoiceDetailedLine.managed_by` is required in the generated v3 schema, but this mapper never sets `ManagedBy`, so any invoice with detailed child lines will serialize `"managed_by": ""`, which is outside the `manual` / `subscription` / `system` enum. Either derive this from the parent standard line when mapping detailed lines, or remove `managed_by` from the detailed-line API model if child lines do not own lifecycle independently.

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +25,941 @@ namespace Invoices;
...
+   * Aggregated financial totals for the invoice.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Totals")
+  totals: Billing.BillingTotals;
+
+  /**
+   * The service period covered by this invoice.
+   *
+   * Optional because some invoice types (e.g. immediate flat fees) do not have a
```
turip: The field is required, and the billing domain always sets a service period for standard invoices. For flat fees the period can be empty (`from == to`) rather than absent, so this doc should describe that shape instead of saying the field is optional.

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +25,941 @@ namespace Invoices;
...
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Draft until")
+  draft_until?: Shared.DateTime;
+
+  /**
+   * Timestamp when usage quantities were last snapshotted for this invoice.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Quantity snapshoted at")
```
turip: Typo in the human-facing summary: "snapshoted" -> "snapshotted". If the wire field name intentionally follows the existing domain/v1 spelling, the summary can still use the correct spelling.

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +25,941 @@ namespace Invoices;
...
+ */
+@friendlyName("BillingInvoiceDetailedLine")
+model BillingInvoiceDetailedLine {
+  ...Shared.Resource;
+
+  /**
+   * The lifecycle manager for this detailed line.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Managed by")
```
turip: Typo: "qunatity" -> "quantity".

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +25,941 @@ namespace Invoices;
...
+}
+
+/**
+ * Snapshot of the customer's information at the time the invoice was issued.
+ *
+ * Extends BillingParty with billing-specific usage attribution data.
+ */
+@friendlyName("BillingInvoiceCustomer")
+model BillingInvoiceCustomer {
+  ...Invoices.BillingParty;
```
turip: I think `BillingParty` is doing two jobs here and ends up exposing fields neither side really supports.

For the supplier side, the actual domain object is `billing.SupplierContact`: `ID`, `Name`, `Address`, and `TaxCode`. Persisted billing profiles / invoice snapshots only carry supplier name, supplier tax code, and supplier address fields; `key` is not part of the supplier model. I would suggest renaming/narrowing this API model to something supplier-specific, e.g. `BillingSupplier`, and omitting unsupported party fields like `key` (and only keeping `id` if we actually intend to populate it).

For the customer side, I do not think `BillingInvoiceCustomer` should spread this party model. The real invoice customer snapshot is `billing.InvoiceCustomer`: customer ID, key, name, billing address, and usage attribution. It does not have `tax_id`, so inheriting `BillingParty` makes the response schema advertise a field we never populate. I also would not reuse the full `Customers.Customer` model directly because that is the live resource shape with labels, description, timestamps, primary email, currency, etc., which are not part of the invoice snapshot.

Something like this would better match the current domain objects:

```tsp
@friendlyName("BillingSupplier")
model BillingSupplier {
  @visibility(Lifecycle.Read)
  id?: string;

  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
  name?: string;

  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
  tax_id?: BillingSupplierTaxIdentity;

  @visibility(Lifecycle.Read, Lifecycle.Create, Lifecycle.Update)
  address?: Shared.Address;
}

@friendlyName("BillingInvoiceCustomer")
model BillingInvoiceCustomer {
  @visibility(Lifecycle.Read)
  id: Shared.ULID;

  @visibility(Lifecycle.Read)
  key?: Shared.ExternalResourceKey;

  @visibility(Lifecycle.Read)
  name: string;

  @visibility(Lifecycle.Read)
  billing_address?: Customers.Address;

  @visibility(Lifecycle.Read)
  usage_attribution: Customers.CustomerUsageAttribution;
}
```

The exact naming/address nesting can follow the style you want for invoices, but I would reuse smaller existing pieces like `Customers.Address` / `CustomerUsageAttribution` where useful, while avoiding both `BillingParty` as a shared base and the full live `Customers.Customer` shape.

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +25,941 @@ namespace Invoices;
...
+   * A flat fee line item charged at a fixed amount.
+   */
+  FlatFee: "flat_fee",
+
+  /**
+   * A usage-based line item charged based on metered consumption.
+   */
+  UsageBased: "usage_based",
+}
+
```
turip, line 637: This is the old line type. Now all standard invoice lines are standard_line typed, so use that distinction.

Add one more commented out type for credit_purchase.

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +25,941 @@ namespace Invoices;
...
+  @visibility(Lifecycle.Read)
+  @summary("Discounts")
+  discounts?: BillingInvoiceLineDiscounts;
+
+  /**
+   * Credit allocations applied to this line item.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Credit allocations")
+  credit_allocations?: BillingInvoiceLineCreditAllocation[];
```
turip: `credits_applied` seems like a better name.

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +25,941 @@ namespace Invoices;
...
+  @visibility(Lifecycle.Read)
+  @summary("Quantity")
+  quantity: Shared.Numeric;
+}
+
+/**
+ * A credit allocation applied to an invoice line item.
+ */
+@friendlyName("BillingInvoiceLineCreditAllocation")
+model BillingInvoiceLineCreditAllocation {
```
turip:
```suggestion
model BillingInvoiceLineCreditsApplied {
```

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +25,941 @@ namespace Invoices;
...
+
+/**
+ * A top-level line item on an invoice.
+ *
+ * Each line represents a single charge, typically associated with a rate card from
+ * a subscription. Detailed (child) lines are nested under `detailed_lines` when
+ * present.
+ */
+@friendlyName("BillingInvoiceLine")
+model BillingInvoiceLine {
```
turip:
```suggestion
model BillingInvoiceStandardLine {
```

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +25,941 @@ namespace Invoices;
...
+  rate_card?: BillingInvoiceLineRateCard;
+
+  /**
+   * Detailed sub-lines that this line has been broken down into.
+   *
+   * Present when `status` is `detailed` or `split`.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Detailed lines")
+  detailed_lines?: BillingInvoiceDetailedLine[];
```
turip: This should not be optional, it can be an empty array, but should not be optionion impho.

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +25,941 @@ namespace Invoices;
...
+  @visibility(Lifecycle.Read)
+  @summary("Subscription reference")
+  subscription?: Subscriptions.SubscriptionReference;
+
+  /**
+   * The rate card configuration snapshot used to price this line item.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Rate card")
+  rate_card?: BillingInvoiceLineRateCard;
```
turip: required field

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +25,941 @@ namespace Invoices;
...
+  @visibility(Lifecycle.Read)
+  @summary("Totals")
+  totals: Billing.BillingTotals;
+
+  /**
+   * The cost category of this detailed line.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Category")
+  category?: BillingInvoiceDetailedLineCostCategory;
```
turip: required, let's default to regular if empty on the source invoice.

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +25,941 @@ namespace Invoices;
...
+  @visibility(Lifecycle.Read)
+  @summary("Managed by")
+  managed_by: BillingInvoiceLineManagedBy;
+
+  /**
+   * Three-letter ISO 4217 currency code.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Currency")
+  currency: Shared.CurrencyCode;
```
turip: drop same as invoice's currency.

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +25,941 @@ namespace Invoices;
...
+  @visibility(Lifecycle.Read)
+  @summary("Discounts")
+  discounts?: BillingInvoiceLineDiscounts;
+
+  /**
+   * Credit allocations applied to this detailed line.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Credit allocations")
+  credit_allocations?: BillingInvoiceLineCreditAllocation[];
```
turip: CreditsApplied

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +25,941 @@ namespace Invoices;
...
+
+  /**
+   * The qunatity snapshot used to calculate charges for this detailed line.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Quantity")
+  quantity: Shared.Numeric;
+
+  /**
+   * The price definition used to calculate charges for this detailed line.
```
turip: rephase (steal from others if needed), this is the price per unit

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +25,941 @@ namespace Invoices;
...
+
+  /**
+   * External identifiers for this detailed line.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("External IDs")
+  external_ids?: BillingInvoiceLineExternalIds;
+
+  /**
+   * The qunatity snapshot used to calculate charges for this detailed line.
```
turip: this is not a snapshot in some cases.

api/spec/packages/aip/src/invoices/invoice.tsp
borosr: What do you think about this description? 
```
For flat fee the service period can be empty which means `from` will be equals to `to`. 
In other cases those fields will be filled with the actual service period.
```

api/v3/handlers/billinginvoices/convert.go
borosr: Sorry, I haven't seen your second comment so I changed it to GenericValidationError. Is that okay or change it to 404?

api/v3/handlers/billinginvoices/convert.go
borosr: Added to the detailed line

api/v3/handlers/billinginvoices/convert.go
```diff
@@ -0,0 +1,435 @@
...
+		payment = &p
+	}
+
+	return api.BillingInvoiceWorkflowSettings{
+		SourceBillingProfileId: w.SourceBillingProfileID,
+		Workflow: api.BillingInvoiceWorkflow{
+			Invoicing: invoicing,
+			Payment:   payment,
+		},
+	}, nil
```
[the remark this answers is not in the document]
borosr: @turip is that good that way?

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,989 @@ namespace Invoices;
...
+@friendlyName("BillingInvoiceDetailedLine")
+model BillingInvoiceDetailedLine {
+  ...Shared.Resource;
+
+  /**
+   * The lifecycle manager for this detailed line.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Managed by")
+  managed_by: BillingInvoiceLineManagedBy;
```
turip: This should be dropped from `BillingInvoiceDetailedLine`. The implementation currently sets detailed-line `ManagedBy` by copying the parent standard line value (`api/v3/handlers/billinginvoices/convert.go:334`), which means detailed lines do not actually own lifecycle management independently. I think we should keep `managed_by` on the parent invoice line and remove it from the detailed-line TypeSpec/schema instead of exposing a duplicated parent value on every child line.

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,989 @@ namespace Invoices;
...
+
+  /**
+   * Line items on this invoice.
+   *
+   * Always returned on single-resource GET; omitted on list endpoints unless
+   * explicitly expanded.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Lines")
+  lines?: BillingInvoiceStandardLine[];
```
turip: this should be a discriminated type that can hold creditpurchase and standard lines.

api/v3/handlers/billinginvoices/convert.go
```diff
@@ -0,0 +1,431 @@
...
+
+	default:
+		return out, models.NewGenericValidationError(fmt.Errorf("unsupported invoice type %q", inv.Type()))
+	}
+
+	return out, nil
+}
+
+func toAPIStandardInvoice(inv billing.Invoice) (api.BillingStandardInvoice, error) {
+	std, err := inv.AsStandardInvoice()
```
turip: nit you can do the convert on the callsite and here use the specific invoice type as argument, so that the union handling is not part of the actual mapping.

api/v3/handlers/billinginvoices/convert.go
```diff
@@ -0,0 +1,431 @@
...
+
+	// Sort lines for consistent output — matches v1 behavior.
+	std.SortLines()
+
+	// Emulate collectionAt: preserve the v1 behavior of defaulting to CreatedAt for
+	// flat-fee-only invoices where the domain leaves CollectionAt nil.
+	collectionAt := std.CollectionAt
+	if collectionAt == nil {
+		collectionAt = lo.ToPtr(std.CreatedAt)
+	}
```
turip:
```
lo.FromPtrOr(std.CollectionAt, std.CreatedAt)
```

also inline to the struct

api/v3/handlers/billinginvoices/convert.go
```diff
@@ -0,0 +1,431 @@
...
+}
+
+func mapRateCard(line *billing.StandardLine) (api.BillingInvoiceLineRateCard, error) {
+	if line.UsageBased == nil {
+		return api.BillingInvoiceLineRateCard{}, nil
+	}
+
+	if line.UsageBased.Price == nil {
+		return api.BillingInvoiceLineRateCard{}, nil
+	}
```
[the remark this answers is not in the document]
turip: @borosr let's apply this. Those values must be set for any invoice.

api/v3/handlers/billinginvoices/convert.go
```diff
@@ -0,0 +1,431 @@
...
+		return nil, nil
+	}
+
+	return lo.MapErr(dls, func(dl billing.DetailedLine, _ int) (api.BillingInvoiceDetailedLine, error) {
+		mapped, err := mapDetailedLine(dl, managedBy)
+		if err != nil {
+			return api.BillingInvoiceDetailedLine{}, fmt.Errorf("mapping detailed line[%s]: %w", dl.ID, err)
+		}
+		return mapped, nil
+	})
```
turip, line 341: Let's sort the detailed lines. There's a helper function for that on the detailed lines AFAIK.

turip (commented):
nice

api/v3/handlers/billinginvoices/convert.go
borosr: Thanks, did a line.SortDetailedLines() before the mapping

api/v3/handlers/billinginvoices/convert.go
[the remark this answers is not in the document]
borosr: Thanks, applied!

api/spec/packages/aip-client-javascript/src/models/types.ts
```diff
@@ -2637,6 +2803,70 @@ export interface Party {
...
+  /** Aggregated financial totals for the invoice. */
+  totals: Totals
+  /** The service period covered by this invoice. For flat fee the service period can be empty which means `from` will be equals to `to`. In other cases those fields will be filled with the actual service period. */
+  service_period: ClosedPeriod
+  /** Validation issues found during invoice processing. Present only when there are one or more validation findings. An empty list is omitted. */
+  validation_issues?: InvoiceValidationIssue[]
+  /** External identifiers assigned to this invoice by third-party systems. */
+  external_ids?: InvoiceExternalIds
+  /** Discriminator field identifying this as a credit note invoice. */
+  type: 'standard'
```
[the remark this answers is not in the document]
borosr: This is not scope of this ticket

api/spec/packages/aip/src/invoices/invoice.tsp
```diff
@@ -17,3 +24,928 @@ namespace Invoices;
...
+  severity: InvoiceValidationIssueSeverity;
+
+  /**
+   * JSON path to the field that caused this validation issue, if applicable.
+   *
+   * For example: `lines/0/rate_card/price`.
+   */
+  @visibility(Lifecycle.Read)
+  @summary("Field path")
+  field?: string;
```
tothandras, line 619: This probably won't work across v1 and v3. Should we comment this out?

e2e/billinginvoices_v3_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+		now := time.Now().UTC()
+		price := api.RateCardUsageBasedPrice{}
+		require.NoError(t, price.FromFlatPriceWithPaymentTerm(api.FlatPriceWithPaymentTerm{
+			Amount:      api.Numeric("10.00"),
+			Type:        api.FlatPriceWithPaymentTermTypeFlat,
+			PaymentTerm: lo.ToPtr(api.PricePaymentTermInAdvance),
+		}))
+
+		lineResp, err := v1.CreatePendingInvoiceLineWithResponse(t.Context(), customerID, api.InvoicePendingLineCreateInput{
```
tothandras, line 138: test v3. gathering invoices should be filtered out

tothandras (changes requested):

api/spec/packages/aip/src/invoices/invoice.tsp
borosr: We are mapping the billing.ValidationIssue.Path into this field, should I delete it instead?

tothandras (approved):

PR #4549, opened 2026-06-22
feat(custom-currencies): add custom currencies support to ledger

mark-vass-konghq:


tools/migrate/migrations/20260620111541_widen_currency_codes.up.sql
```diff
@@ -0,0 +1,27 @@
...
+-- modify "charge_credit_purchases" table
+ALTER TABLE "charge_credit_purchases" ALTER COLUMN "currency" TYPE character varying(24);
+-- modify "charge_flat_fee_run_detailed_lines" table
+ALTER TABLE "charge_flat_fee_run_detailed_lines" ALTER COLUMN "currency" TYPE character varying(24);
+-- modify "charge_flat_fees" table
+ALTER TABLE "charge_flat_fees" ALTER COLUMN "currency" TYPE character varying(24);
+-- modify "charge_usage_based" table
+ALTER TABLE "charge_usage_based" ALTER COLUMN "currency" TYPE character varying(24);
+-- modify "charge_usage_based_run_detailed_line" table
+ALTER TABLE "charge_usage_based_run_detailed_line" ALTER COLUMN "currency" TYPE character varying(24);
```
turip: let's only change those parts of the db where this makes sense for this change if possible.

e.g. billing will not use non-fiat currencies as currency.

turip: I mean that's my recommendation.

PR #4568, opened 2026-06-24
Add unitconfig to ratecard in domain layer - OM-393

rolosp:
## Overview

Added `UnitConfig` to `RateCardMeta` in preparation for the next phase, persisting it. This change should be pretty low risk as:

- it is additive only even with the change being on the `RateCardMeta`
- no surface moved as persisting it will be the next ticket and there is no api wiring or typespec change
- no behavior change if the unit config field is nil

## Summary by CodeRabbit

* **New Features**
  * Added optional per-rate-card unit conversion configuration (multiply/divide) with configurable rounding mode, precision, and display unit.
  * Rate card calculations can now produce both converted and invoiced quantities, applying rounding/precision when configured.

* **Bug Fixes**
  * Improved unit conversion handling for cloning, equality comparisons, and validation error reporting (including unit-config field paths).

* **Tests**
  * Added coverage for unit conversion apply/validate/equality/clone, plus rate card metadata behavior and deep-copy semantics.

### DIFF
--- a/openmeter/productcatalog/ratecard.go
+++ b/openmeter/productcatalog/ratecard.go
@@ -247,20 +247,24 @@ func (r RateCardMeta) Validate() error {
 
 	if r.TaxConfig != nil {
 		if err := r.TaxConfig.Validate(); err != nil {
-			errs = append(errs, fmt.Errorf("invalid tax config: %w",
+			errs = append(errs, fmt.Errorf(
+				"invalid tax config: %w",
 				models.ErrorWithFieldPrefix(
 					models.NewFieldSelectorGroup(models.NewFieldSelector("taxConfig")),
-					err),
+					err,
+				),
 			))
 		}
 	}
 
 	if r.Price != nil {
 		if err := r.Price.Validate(); err != nil {
-			errs = append(errs, fmt.Errorf("invalid price: %w",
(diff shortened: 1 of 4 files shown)


openmeter/productcatalog/ratecard.go
```diff
@@ -269,6 +282,16 @@ func (r RateCardMeta) Validate() error {
 		errs = append(errs, err)
 	}
 
+	if r.UnitConfig != nil {
+		if err := r.UnitConfig.Validate(); err != nil {
+			errs = append(errs, fmt.Errorf("invalid unit config: %w",
+				models.ErrorWithFieldPrefix(
+					models.NewFieldSelectorGroup(models.NewFieldSelector("unitConfig")),
```
tothandras: If we expose these errors on the API for v3, how can we make sure the field selectors are correct for both v1 and v3?

rolosp: I don't have a general solution for this, but as we only expose this on v3 for authoring I just renamed the field selector. But yes the same problem comes up with tax code too. 

```go
models.NewFieldSelectorGroup(models.NewFieldSelector("taxConfig")),
```

chrisgacsal (approved):

PR #4604, opened 2026-06-29
feat: unit config authoring on v3 api - OM-396

rolosp:
## Overview

Wires unit_config through the v3 plan create/update path, all gated behind the dark unitConfig.enabled flag, plus a fix for the OM-395 adapter mappers that dropped  unit_config on expanded plan/add-on relations. 

### Note to reviewers

- the v1 read-side coexistence gap (a v3-authored unit_config silently dropped when listed on v1) is a known and is deferred to a later ticket

## Summary by CodeRabbit

* **New Features**
  * Added explicit unit configuration support for v3 rate-card conversions, including new bidirectional unit-config conversion helpers.
* **Bug Fixes**
  * Preserved stored unit configuration when converting and mapping plan/addon rate-cards into product-catalog models.
  * Enforced that unit configuration is only valid with usage-based or tiered price types; invalid requests now return a dedicated “unit config requires usage-based price” validation error.
  * Improved conversion-factor handling by surfacing errors for non-numeric values and aligning precision/rounding behavior.
* **Tests**
  * Added/extended unit-config mapping, round-trip, and validation coverage across v3 conversions and adapter mappings.

### DIFF
--- a/openmeter/productcatalog/ratecard.go
+++ b/openmeter/productcatalog/ratecard.go
@@ -296,6 +296,16 @@ func (r RateCardMeta) Validate() error {
 				),
 			))
 		}
+
+		// A unit_config is a per-quantity conversion applied before rating, so it is
+		// only valid on a unit or tiered (graduated/volume) price. It is rejected on
+		// flat prices (not per-quantity) and on package/dynamic prices (which already
+		// carry their own conversion and would otherwise convert twice).
+		priceAllowsUnitConfig := r.Price != nil &&
+			(r.Price.Type() == UnitPriceType || r.Price.Type() == TieredPriceType)
+		if !priceAllowsUnitConfig {
+			errs = append(errs, ErrRateCardUnitConfigRequiresUsageBasedPrice)
+		}
 	}
 
 	return models.NewNillableGenericValidationError(errors.Join(errs...))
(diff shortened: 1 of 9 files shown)


rolosp:
As the feature is still work in progress and behind a config flag, for now I relied mostly on unit test coverage but for sanity check I created an e2e test with hurl and tested that the v3 unit config authoring works e2e.

openmeter/productcatalog/ratecard.go
```diff
@@ -296,6 +296,16 @@ func (r RateCardMeta) Validate() error {
...
+
+		// A unit_config is a per-quantity conversion applied before rating, so it is
+		// only valid on a unit or tiered (graduated/volume) price. It is rejected on
+		// flat prices (not per-quantity) and on package/dynamic prices (which already
+		// carry their own conversion and would otherwise convert twice).
+		priceAllowsUnitConfig := r.Price != nil &&
+			(r.Price.Type() == UnitPriceType || r.Price.Type() == TieredPriceType)
+		if !priceAllowsUnitConfig {
+			errs = append(errs, ErrRateCardUnitConfigRequiresUsageBasedPrice)
+		}
```
turip, line 308: I can imagine that this might be used at other places, so it might worth just adding a SupportsUnitConfig() (bool) to the Price interface, that delegates it to the downstream pricers.

turip (approved):

openmeter/productcatalog/ratecard.go
rolosp: good idea, i'll add it to the next pr

PR #4614, opened 2026-06-30
feat(ledger): add entry charge provenance

GAlexIHU:
## Summary

Adds V2 charge provenance to ledger entries without introducing new provenance resources.

The ledger now stores two nullable entry facts:

- `source_charge_id`: the creditpurchase/promotional charge that funded credit
- `spend_charge_id`: the flat-fee or usage-based charge that consumed or accrued value

These are entry-level facts, not route dimensions. They are encoded into `identity_key` so same-route entries remain split by provenance without creating high-cardinality subaccounts.

## Why

We need revenue and credit reporting to answer questions like:

- how much credit from creditpurchase charge CP1 remains available?
- how much credit from CP1 was consumed by spend charge SP1?
- how much recognized earnings came from CP1, SP1, or the CP1 x SP1 pair?

The existing route/subaccount model preserves accounting dimensions, but coalescing by route loses source x spend attribution in collection, correction, and recognition flows.

## What Changed

- Added `source_charge_id` and `spend_charge_id` to `ledger_entries`, Ent schema/code, historical mappings, and migrations.
- Added entry identity parsing/validation so provenance fields and `identity_key` must agree.
- Added grouped balance buckets via `BalanceBucketQuery` / `GetBalanceBuckets`, with `GroupByValues` for provenance-style dimensions.
- Updated creditpurchase issuance, authorization, settlement, advance attribution, and cost-basis translation to stamp/preserve source provenance.
- Updated flat-fee and usage-based accrual/collection flows to stamp spend provenance and preserve selected source provenance.
- Updated FBO collection to select from source-level balance buckets instead of only subaccount totals.
- Updated breakage release/reopen paths to preserve source provenance where appropriate.
- Updated correction paths to preserve facts when reversing or reclassifying source/spend buckets.
- Updated revenue recognition to preserve source/spend facts for credit-backed accrued value.

## Important Scope Choices

- No new `source_credit`, `obligation`, `collection_leg`, or recognition-leg resources.
- `source_charge_id` is always the creditpurchase/promotional source charge when set.
- `spend_charge_id` is the flat-fee or usage-based charge when set.
- Invoice-backed accrued value is not recognized by the recognizer yet; recognition stays credit-lineage gated until invoice-backed recognition and correction are implemented together.
- Breakage planning remains plan/subaccount based. Release/reopen entries preserve source provenance, but source-aware breakage planning is still future work if needed.
- Legacy entries remain nullable; historical exact backfill is not attempted in this PR.

## Validation

Local `make test` was started after the rebase but intentionally interrupted; CI should run the full suite for the PR.

## Summary by CodeRabbit

* **New Features**
  * Added historical balance bucket querying with provenance grouping (source/spend charge).
  * Persisted per-entry schema version plus optional source/spend charge provenance.
  * Breakage records can now store optional source charge provenance.
* **Bug Fixes**
  * Improved charge-provenance propagation across collections, credit purchases, flat fees, usage-based flows, breakage release/reopen, corrections, and earnings recognition.
  * Added stricter identity-key validation and provenance-aware grouping/routing behavior.
* **Tests**
  * Expanded integration coverage for bucket grouping, backfill/provenance preservation, and correction/reopen scenarios.


openmeter/ledger/chargeadapter/creditpurchase.go
```diff
@@ -415,101 +460,185 @@ func (h *creditPurchaseHandler) advanceAttributions(
...
 		if !balance.IsNegative() {
 			continue
 		}
 
-		attributed := balance.Neg()
-		if attributed.GreaterThan(remaining) {
-			attributed = remaining
-		}
+		eligibleAdvanceReceivables = append(eligibleAdvanceReceivables, advanceReceivable)
+		advanceReceivablesByMatchKey[advanceReceivable.matchKey] = advanceReceivable
```
turip: [P2] `advanceBackfillMatchKey` only contains the spend charge, so two open receivable buckets for the same spend but different feature routes overwrite each other here. The later attribution still counts both amounts in `totalAdvanceReceivable`, but only one bucket/address/features survives in these maps, and the fallback loop can reuse the same remaining amount for duplicate keys. Include the route/address in the match key or accumulate per spend bucket so backfill preserves each open receivable slice.

turip (commented):
Found one issue worth tightening before merge.

openmeter/ledger/chargeadapter/creditpurchase.go
```diff
@@ -415,101 +460,185 @@ func (h *creditPurchaseHandler) advanceAttributions(
...
+		advanceRemainingByMatchKey[advanceReceivable.matchKey] = balance.Neg()
```
tothandras: **P1 — advance-receivable buckets collapse on `spendChargeID` alone, can overdraw a route**

`advanceReceivablesByMatchKey` and `advanceRemainingByMatchKey` are keyed only by `advanceBackfillMatchKey{spendChargeID}`, while the accrued side uses the richer `accruedBackfillBucketKey` (spend + tax). When two open advance-receivable buckets share a `spendChargeID`, the last one wins in both maps (here at L479-480), but `totalAdvanceReceivable` (L481) sums **all** of them.

This is reachable today via the legacy path: every pre-PR advance receivable has `spend_charge_id = NULL → "null"`, and a receivable account is route-multiplexed by `{Currency, TaxCode, Features, CostBasis, AuthStatus}` (see `accounts.go` + `balancebuckets_query.go` grain of `(sub_account_id, spend_charge_id)`). A customer with advances on 2+ feature/tax routes therefore yields 2+ buckets all keyed `"null"`. A later credit purchase can then post the full backfill against a single route (the surviving map entry) — overdrawing it while leaving the other route open, and attributing route/features from the wrong bucket (L516, L541).

Same root cause drives the over-attribution variant: `allocateAccruedAttribution`/`totalUnattributedAccruedBalance` (L674-686, L705-711) gate accrued buckets only on the *existence* of a matching advance key and bound by accrued amount, never by that key's advance amount — so the accrued sum for a key can exceed its advance and push `advanceRemainingByMatchKey[K]` negative (L522).

**Fix direction:** give the advance match key the same dimensionality as the bucket grain (route/feature/tax), so each bucket gets its own remaining/attribution slot. Worth a multi-route legacy test.

Confirmed independently by Greptile, a cardinality trace, and Codex.

openmeter/ledger/transactions/accrual.go
```diff
@@ -300,15 +323,24 @@ func (t TransferCustomerFBOAdvanceToAccruedTemplate) correct(scope CorrectionInp
...
+	var accruedIdentity ledger.EntryIdentityParts
 
 	for _, entry := range scope.OriginalTransaction.Entries() {
 		switch {
 		case entry.PostingAddress().AccountType() == ledger.AccountTypeCustomerFBO && entry.Amount().IsNegative():
 			fboAddress = entry.PostingAddress()
 			fboAmount = fboAmount.Add(entry.Amount().Abs())
+			fboIdentity = ledger.EntryIdentityParts{
+				SourceChargeID: entry.SourceChargeID(),
+			}
```
tothandras, line 336: **P1 — FBO reversal leg drops `SpendChargeID`, drifting per-spend FBO balance buckets**

The forward `resolve()` posts the FBO advance leg with `identity: {SpendChargeID: t.SpendChargeID}` (see L408-410), so the forward entry lives in the **spend-charge** FBO bucket. But this correction leg builds:

```go
fboIdentity = ledger.EntryIdentityParts{
    SourceChargeID: entry.SourceChargeID(), // always nil for this template
}
```

It reads the always-nil source and **omits `SpendChargeID` entirely**, while the accrued leg right below (L340-343) correctly preserves both. Net effect: a correction reverses out of the spend-charge FBO bucket but posts the positive reversal into the source-less/spend-less bucket. Transaction totals still balance, but per-spend FBO balance buckets drift — and those buckets are exactly what `advanceReceivableBalances` reads (grouped by `spend_charge_id`) during a later credit purchase, so the drift propagates into attribution.

This is the precise source/spend-drop this PR exists to prevent.

**Fix:** mirror the accrued leg —
```go
fboIdentity = ledger.EntryIdentityParts{
    SpendChargeID: entry.SpendChargeID(),
}
```
Add a correction test asserting the reversal lands in the same spend bucket as the forward leg.

Found by both the Claude adversarial pass and Codex.

openmeter/ledger/collector/collection_fbo.go
```diff
@@ -201,26 +178,29 @@ func (c *accrualCollector) listCustomerFBOSources(
 			continue
 		}
 
+		consumedSources := consumeFBOBalanceBucketSources(sources, plan.FBOSubAccountID, available)
```
tothandras: **P2 — FBO source consumption order is DB-plan dependent (nondeterministic attribution)**

`consumeFBOBalanceBucketSources` drains the `sources` slice in the order it was built, which is the row order returned by `GetBalanceBuckets`. That query has no `ORDER BY` (`balancebuckets_query.go` builds a grouped `SELECT` with no ordering), and the priority/expiry/cursor sort (`fboCollectionSource.Compare`, this file L77/L82-103) runs only **after** consumption. So when a breakage plan's open amount is smaller than a sub-account's available balance spread across multiple source buckets, which source bucket gets consumed/released is Postgres-plan dependent and can flip across DB versions with no code or data change.

Related: when multiple open breakage plans share an FBO sub-account, this loop drains `availableBySubAccountID` in `ListPlans` order (L165-201), not by credit priority/expiry — so a lower-priority/later-expiring plan returned first can reserve the bucket ahead of the plan the post-consumption sort would have preferred.

Note the existing TODO at L83-85 ("Version this collection-order contract before changing it") and the PR's scope note that source-aware breakage planning is future work.

**Lowest-risk fix:** add a deterministic `ORDER BY` to the balance-bucket query (e.g. by `source_charge_id`, then a stable cursor) so source attribution is stable across DB plans. The priority-aware multi-plan ordering can stay as documented future work given the L83 versioning contract.

Found by Greptile (P2) and Codex.

openmeter/ledger/chargeadapter/creditpurchase.go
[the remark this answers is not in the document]
tothandras: **Confirmed valid** — verified reachable via the legacy nil-spend path: pre-PR advance receivables all have `spend_charge_id = NULL → "null"`, and a receivable account is route-multiplexed by {Currency, TaxCode, Features, CostBasis, AuthStatus}, so a customer with advances on 2+ routes yields 2+ buckets colliding on one match key. Detailed analysis + fix direction (align the advance match key dimensionality with the bucket grain) posted at https://github.com/openmeterio/openmeter/pull/4614#discussion_r3499295302

openmeter/ledger/collector/collection_fbo.go
```diff
@@ -201,26 +178,29 @@ func (c *accrualCollector) listCustomerFBOSources(
...
 		planCopy := plan
 		expiresAt := plan.ExpiresAt
 		route := plan.FBOAddress.Route().Route()
 		if len(route.Features) > 0 && !lo.Contains(route.Features, featureKey) {
 			continue
 		}
```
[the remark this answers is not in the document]
tothandras: **Partially valid — fix is incomplete as suggested.** The drain-before-`continue` is real, but note the suggested patch only moves the check above `consumeFBOBalanceBucketSources` (L181) and leaves the `availableBySubAccountID` drain at L175 above the check too. The feature check should move above **both** mutations. In practice the L185 guard is currently unreachable-after-drain (the route filter in `listCustomerFBOBalanceBucketSources` — `MatchFeature`/`Features=Some(nil)` — already excludes any sub-account whose plan would fail L185), so this is defensive rather than actively firing. Recommend reordering anyway since it costs nothing and the coupling is fragile.

openmeter/ledger/collector/collection_fbo.go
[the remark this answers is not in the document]
tothandras: **Confirmed valid.** Same root surfaced by the cross-model adversarial pass: `consumeFBOBalanceBucketSources` drains `sources` in `GetBalanceBuckets` row order (no `ORDER BY` in the bucket query), and the priority/expiry sort runs only after consumption — so source attribution is also DB-plan nondeterministic, not just plan-order dependent. Detail + suggested fix (deterministic ORDER BY on the bucket query) at https://github.com/openmeterio/openmeter/pull/4614#discussion_r3499298338. Noting the L83-85 'version this ordering contract' TODO and the PR's future-work scope for source-aware breakage planning.

openmeter/ledger/chargeadapter/creditpurchase.go
GAlexIHU: spendCharge is of higher granularity than the dimensions, for any SpendCharge dimensions like feautre, tax, etc... are deterministic

openmeter/ledger/transactions/accrual.go
GAlexIHU: this is wrong. when reversing accrual (regardless whether from FBO or receivable) its nonsense to preserve spend provenance, as the correction means that amount becomes "unspent". this is not a two way street, we only want to preserve the source_charge provenance)

turip (approved):
lgfm

PR #4616, opened 2026-06-30
feat: apply unit config in rating via mutator - OM-397

rolosp:
## Overview

Apply unit_config in rating (charges path)

Wires `unit_config` into the billing rating engine so a usage-based rate card's conversion (operation × factor, then rounding) is applied to the metered quantity before pricing — previously authored and persisted (OM-395, OM-396) but inert at billing time. Dark behind `unitConfig.enabled`; with the flag off, rating is byte-identical to today.

This covers the **charges path** end-to-end. Standard/legacy invoice lines return no config yet — that's the next ticket (persist `applied_unit_config` on invoice lines), where the convert+round also gets applied in the charges line-mapper.

### Changes
- New `UnitConfig` pre-calculation mutator: converts + rounds on the cumulative endpoints (`round(convert(pre+qty)) − round(convert(pre))`) so progressive splits
  round on the cumulative boundary, then hands converted units to the existing
  unit/graduated/volume pricers (unchanged, conversion-unaware). Uses `UnitConfig.Apply`.
- Registered ahead of `DiscountUsage` (convert → round → discount), gated by the flag.
- `unitConfig.enabled` threaded into the rating service via DI, defaulting off.
- `GetUnitConfig()` on the rating line accessor; `RateableIntent` returns the
  snapshotted charge-intent config.

### Testing
- Mutator unit tests (multiply/divide, all rounding modes, cumulative boundary, idempotency).
- Pricer registration test (flag on/off ordering, flat price unaffected).
- End-to-end rating + DB-backed charges integration tests: flag-on bills the converted
  quantity, flag-off bills raw.

## Summary by CodeRabbit

* **New Features**
  * Added deploy-wide enablement for `unit_config` during billing rating, including unit conversion with configurable rounding for eligible usage-based pricing.
  * Usage-based charge intents now capture `unit_config` from rate cards so rating uses the intended units.
* **Bug Fixes**
  * Rating now rejects charges when `unit_config` is present but the feature is disabled, instead of falling back to raw quantities.
  * Split/progressive usage calculations now apply conversion+rounding consistently to the correct endpoints.
* **Tests**
  * Added/updated end-to-end and unit tests covering enabled/disabled flows and conversion/rounding behavior.

### DIFF
--- a/openmeter/billing/charges/service/unitconfig_rating_test.go
+++ b/openmeter/billing/charges/service/unitconfig_rating_test.go
@@ -50,15 +50,15 @@ func (s *unitConfigRatingEnabledSuite) TestRatesConvertedQuantity() {
 
 	stdLine := invoices[0].Lines.OrEmpty()[0]
 
-	// MeteredQuantity records the raw metered value and stays 7400: the unit_config
-	// conversion changes the priced amount (asserted below), not the recorded metered
-	// quantity.
-	//
-	// TODO: once the charges line-mapper converts+rounds the displayed UsageBased.Quantity,
-	// extend this test to also assert the converted displayed quantity.
-	// MeteredQuantity stays raw (7400) as the audit value.
+	// The line is internally consistent: MeteredQuantity is the raw audit value (7400),
+	// while the customer-facing billable Quantity is the converted ceil(7400/1000) = 8,
+	// matching the priced amount. The line-mapper converts the displayed quantity through
+	// the same unit_config as rating.
 	s.Require().NotNil(stdLine.UsageBased.MeteredQuantity)
 	s.Equal(float64(7400), lo.FromPtr(stdLine.UsageBased.MeteredQuantity).InexactFloat64())
+
+	s.Require().NotNil(stdLine.UsageBased.Quantity)
+	s.Equal(float64(8), lo.FromPtr(stdLine.UsageBased.Quantity).InexactFloat64())
 
 	s.RequireTotals(billingtest.ExpectedTotals{
 		Amount: 8,
--- a/openmeter/billing/charges/service/unitconfig_rating_test.go
+++ b/openmeter/billing/charges/service/unitconfig_rating_test.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/openmeterio/openmeter/openmeter/billing"
 	"github.com/openmeterio/openmeter/openmeter/billing/charges"
+	"github.com/openmeterio/openmeter/openmeter/billing/rating/service/mutator"
 	"github.com/openmeterio/openmeter/openmeter/productcatalog"
 	"github.com/openmeterio/openmeter/pkg/clock"
 	"github.com/openmeterio/openmeter/pkg/datetime"
@@ -42,7 +43,27 @@ func (s *unitConfigRatingEnabledSuite) TearDownTest() {
 
 func (s *unitConfigRatingEnabledSuite) TestRatesConvertedQuantity() {
 	// flag on: 7400 raw / 1000, ceiling => 8 billed units * $1 = $8.
-	s.runUnitConfigChargesScenario(8)
+	invoices, err := s.invoiceUnitConfigChargesScenario()
+	s.Require().NoError(err)
+	s.Require().Len(invoices, 1)
+	s.Require().Len(invoices[0].Lines.OrEmpty(), 1)
+
+	stdLine := invoices[0].Lines.OrEmpty()[0]
+
+	// MeteredQuantity records the raw metered value and stays 7400: the unit_config
+	// conversion changes the priced amount (asserted below), not the recorded metered
+	// quantity.
+	//
+	// TODO: once the charges line-mapper converts+rounds the displayed UsageBased.Quantity,
+	// extend this test to also assert the converted displayed quantity.
+	// MeteredQuantity stays raw (7400) as the audit value.
+	s.Require().NotNil(stdLine.UsageBased.MeteredQuantity)
+	s.Equal(float64(7400), lo.FromPtr(stdLine.UsageBased.MeteredQuantity).InexactFloat64())
+
+	s.RequireTotals(billingtest.ExpectedTotals{
+		Amount: 8,
+		Total:  8,
+	}, stdLine.Totals)
 }
 
 func TestUsageBasedUnitConfigRatingDisabled(t *testing.T) {
@@ -54,8 +75,8 @@ type unitConfigRatingDisabledSuite struct {
 }
 
 func (s *unitConfigRatingDisabledSuite) SetupSuite() {
-	// UnitConfigEnabled defaults to false: the intent still carries a unit_config,
-	// but the mutator is not registered, so rating must bill the raw quantity.
+	// UnitConfigEnabled defaults to false: the intent still carries a unit_config, so
+	// ForbidUnitConfig must reject rating rather than silently bill the raw quantity.
 	s.BaseSuite.SetupSuite()
 }
 
@@ -63,15 +84,19 @@ func (s *unitConfigRatingDisabledSuite) TearDownTest() {
 	s.BaseSuite.TearDownTest()
 }
 
-func (s *unitConfigRatingDisabledSuite) TestRatesRawQuantityWhenFlagOff() {
-	// flag off: unit_config ignored, 7400 raw units * $1 = $7400 (parity with today).
-	s.runUnitConfigChargesScenario(7400)
+func (s *unitConfigRatingDisabledSuite) TestErrorsWhenFlagOff() {
+	// flag off + a unit_config on the charge: ForbidUnitConfig surfaces the
+	// inconsistency instead of silently billing the raw quantity.
+	_, err := s.invoiceUnitConfigChargesScenario()
+	s.Require().Error(err)
+	s.Require().ErrorIs(err, mutator.ErrUnitConfigDisabled)
 }
 
-// runUnitConfigChargesScenario creates a usage-based charge carrying a divide-by-1000
-// ceiling unit_config, seeds 7400 raw units, invoices mid-period, and asserts the
-// rated line amount matches expectedAmount.
-func (s *BaseSuite) runUnitConfigChargesScenario(expectedAmount float64) {
+// invoiceUnitConfigChargesScenario creates a usage-based charge carrying a divide-by-1000
+// ceiling unit_config, seeds 7400 raw units, and invoices mid-period. It returns the
+// InvoicePendingLines result so each suite asserts its own outcome (converted amount
+// when the flag is on, a ForbidUnitConfig error when it is off).
+func (s *BaseSuite) invoiceUnitConfigChargesScenario() ([]billing.StandardInvoice, error) {
 	s.T().Helper()
 
 	ctx := s.T().Context()
@@ -83,7 +108,8 @@ func (s *BaseSuite) runUnitConfigChargesScenario(expectedAmount float64) {
 	cust := s.CreateTestCustomer(ns, "test-subject")
 	s.NotEmpty(cust.ID)
 
-	_ = s.ProvisionBillingProfile(ctx, ns, customInvoicing.App.GetID(),
+	_ = s.ProvisionBillingProfile(
+		ctx, ns, customInvoicing.App.GetID(),
 		billingtest.WithProgressiveBilling(),
 		billingtest.WithCollectionInterval(datetime.MustParseDuration(s.T(), "P2D")),
 		billingtest.WithManualApproval(),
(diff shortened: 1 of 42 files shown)


openmeter/billing/stdinvoiceline.go
```diff
@@ -588,6 +588,22 @@ func (i StandardLine) GetRateCardDiscounts() Discounts {
...
+// computes the Pre-populated cumulative case for it), but the source field on this line
+// is a separate, later ticket.
+//
+// TODO(unit-config seam ④ / W4 — invoice-line applied_unit_config snapshot): when
+// UsageBasedLine gains the write-once AppliedUnitConfig snapshot, return it here instead
+// of nil. Until then a unit_config rate card must not be billed through the legacy path,
+// or rating silently bills raw quantities (overbilling by the conversion factor).
+func (i StandardLine) GetUnitConfig() *productcatalog.UnitConfig {
+	return nil
+}
```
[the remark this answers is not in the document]
rolosp: Not reachable today. Nothing feeds a unit_config into a standard line (v1 manual lines can't express it; subscriptions rate via the charges path), so nil is correct, not a live overbill. Applying/failing-closed needs the applied_unit_config column, which is the next ticket the TODO here points at it.

openmeter/billing/rating/service/pricer.go
```diff
@@ -70,11 +70,18 @@ func getPricerFor(line rating.PriceAccessor, opts rating.GenerateDetailedLinesOp
 		postCalculationMutators = append(postCalculationMutators, &mutator.Credits{})
 	}
 
+	// UnitConfig converts the raw metered quantity into billed units and must run
+	// before DiscountUsage so the usage discount applies to the converted, rounded
+	// quantity (convert → round → discount).
+	preCalculationMutators := make([]mutator.PreCalculationMutator, 0, 2)
+	if unitConfigEnabled {
+		preCalculationMutators = append(preCalculationMutators, &mutator.UnitConfig{})
+	}
```
turip, line 83: Call me paranoid, but instead of this, I would do something like this:

```
	if unitConfigEnabled {
		preCalculationMutators = append(preCalculationMutators, &mutator.UnitConfig{})
	} else {
        preCalculationMutators = append(preCalculationMutators, &mutator.ForbidUnitConfig{})
    }
```

Forbid unit config is a noop, but returns error when the unit config is set on the rateable. This way we are not accidentally ignoring unit configs.

openmeter/billing/rating/service/mutator/unitconfig.go
```diff
@@ -0,0 +1,54 @@
...
+		return l, nil
+	}
+
+	usage, err := l.GetUsage()
+	if err != nil {
+		return l, err
+	}
+
+	_, start := unitConfig.Apply(usage.PreLinePeriodQuantity)
+	_, end := unitConfig.Apply(usage.PreLinePeriodQuantity.Add(usage.Quantity))
```
turip: just convert the quantity, don't do calculations here. It is not necessarily true that quantity = perlineperiodquantity + periodquantity (e.g. for exotic meters, and charges rating)

test/app/testenv.go
```diff
@@ -274,7 +274,7 @@ func InitBillingService(t *testing.T, ctx context.Context, in InitBillingService
 	})
 	require.NoError(t, err)
 
-	billingRatingService := billingratingservice.New()
+	billingRatingService := billingratingservice.New(billingratingservice.Config{})
```
turip: enable in all test envs, so that you are validating that there's no regression due to this being enabled.

openmeter/billing/charges/usagebased/rating.go
```diff
@@ -53,6 +53,14 @@ func (r RateableIntent) GetRateCardDiscounts() billing.Discounts {
 	return r.Discounts.Clone()
 }
 
+func (r RateableIntent) GetUnitConfig() *productcatalog.UnitConfig {
```
turip, line 56: [P1] Subscription charges drop unit_config before rating

This new rating path only sees unit_config through RateableIntent.GetUnitConfig(), but normal subscription-created usage-based charges still never populate Intent.UnitConfig: newUsageBasedChargeIntent in openmeter/billing/worker/subscriptionsync/service/reconciler/patchchargeusagebased.go copies price and discounts from rateCardMeta but leaves UnitConfig behind. With unitConfig.enabled on, a plan/subscription rate card using unit_config will still invoice raw metered quantity, while only direct charge-intent tests exercise the converted path. Please snapshot rateCardMeta.UnitConfig into the usage-based charge intent, with a subscription round-trip/invoice test.

openmeter/billing/rating/service/mutator/unitconfig.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	// Defensive: the authoring validator forbids a unit_config on price types that
+	// cannot convert (flat/package/dynamic), so reaching the mutator with an
+	// unsupported price means inconsistent data. Skip conversion rather than
+	// mis-price, leaving the pricer to bill the raw quantity as it does today.
+	price := l.GetPrice()
+	if price == nil || !price.SupportsUnitConfig() {
+		return l, nil
+	}
```
turip, line 49: this should be an error. Don't ignore inconsistencies in billing or rating, instead make them surfaced.

openmeter/billing/rating/service/testutil/ubptest.go
```diff
@@ -46,6 +46,26 @@ type CalculationTestCase struct {
...
+// StandardLine type itself never carries.
+type unitConfigLineAccessor struct {
+	*billing.StandardLine
+
+	unitConfig *productcatalog.UnitConfig
+}
+
+func (l unitConfigLineAccessor) GetUnitConfig() *productcatalog.UnitConfig {
+	return l.unitConfig
 }
```
turip, line 70: I would prefer if you add this to the line even if it's not persisted, or add a TODO here to use the line's unit config once it's there.

Given you have done it for charges, you might aswell do this for standard lines.

openmeter/billing/rating/service/pricer.go
rolosp: added `ForbidUnitConfig`

openmeter/billing/rating/service/mutator/unitconfig.go
rolosp: done

test/app/testenv.go
rolosp: done

openmeter/billing/charges/usagebased/rating.go
rolosp: done,  newUsageBasedChargeIntent now snapshots rateCardMeta.UnitConfig onto the usage-based charge intent

openmeter/billing/rating/service/mutator/unitconfig.go
rolosp: done

openmeter/billing/rating/service/testutil/ubptest.go
rolosp: Added a TODO comment and deferred the standard-line wiring

openmeter/billing/rating/service/mutator/unitconfig.go
```diff
@@ -0,0 +1,63 @@
...
+	if price == nil {
+		return l, fmt.Errorf("line has no price: %w", ErrUnitConfigUnsupportedPrice)
+	}
+	if !price.SupportsUnitConfig() {
+		return l, fmt.Errorf("price type %q: %w", price.Type(), ErrUnitConfigUnsupportedPrice)
+	}
+
+	usage, err := l.GetUsage()
+	if err != nil {
+		return l, err
```
borosr: Please wrap the error

borosr (dismissed):
lgtm

openmeter/billing/stdinvoiceline.go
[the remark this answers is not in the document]
tothandras: Confirmed — valid finding. This PR intentionally lands unit_config conversion on the charges path only; `StandardLine.GetUnitConfig()` returns nil by design until the `AppliedUnitConfig` invoice-line snapshot lands (`TODO(unit-config seam ④ / W4)` at `stdinvoiceline.go:598`, next ticket: persist `applied_unit_config` on invoice lines).

The silent-raw-billing window is real when a usage-based line routes to the generic engine (`LineEngineTypeInvoice`) instead of the charges engine. We're gating on the operational invariant that `unitConfig.enabled` is only turned on where charges routing is guaranteed on (credits + credit-then-invoice enabled, `Credits()` namespace gate on), so a unit_config line always reaches `RateableIntent.GetUnitConfig()` and never the nil-returning standard-line accessor.

A mutator-layer fail-closed guard can't cover this: the standard line carries no unit_config signal to distinguish "no config" from "config dropped" (unit_config lives on the rate card, not the `Price`). So the durable fix is the tracked invoice-line snapshot, not a guard. Tracking under the W4 invoice-line applied_unit_config ticket.

tothandras (dismissed):

openmeter/billing/charges/service/unitconfig_rating_test.go
```diff
@@ -0,0 +1,174 @@
...
+	// extend this test to also assert the converted displayed quantity.
+	// MeteredQuantity stays raw (7400) as the audit value.
+	s.Require().NotNil(stdLine.UsageBased.MeteredQuantity)
+	s.Equal(float64(7400), lo.FromPtr(stdLine.UsageBased.MeteredQuantity).InexactFloat64())
+
+	s.RequireTotals(billingtest.ExpectedTotals{
+		Amount: 8,
+		Total:  8,
+	}, stdLine.Totals)
+}
```
mark-vass-konghq, line 67:
```suggestion
   s.Require().NotNil(stdLine.UsageBased.Quantity)
   s.Equal(float64(8), lo.FromPtr(stdLine.UsageBased.Quantity).InexactFloat64())
}
```

mark-vass-konghq: MeteredQuantity and Quantity mean different things. MeteredQuantity should stay as raw usage, e.g. 7400, while Quantity should be the billable converted amount, e.g. ceil(7400 / 1000) = 8.

  The PR makes rating/totals use the converted quantity correctly, so totals become $8, but the invoice line mapper still exposes UsageBased.Quantity = 7400. That makes the invoice/API line internally inconsistent:
  raw metered usage is 7400, billable total is based on 8, but customer-facing quantity still says 7400.

  The expected result is:

  MeteredQuantity: 7400
  Quantity:        8
  Totals.Amount:   8

  A good regression test should assert both: MeteredQuantity == 7400 and Quantity == 8.

mark-vass-konghq (changes requested):

openmeter/billing/rating/service/mutator/unitconfig.go
rolosp: done

openmeter/billing/charges/service/unitconfig_rating_test.go
rolosp: It was a known issue and was deferred to a later ticket originally, but it is pretty surgical and makes sense to do it here so I've pushed the changes.

openmeter/billing/charges/usagebased/service/linemapper.go
```diff
@@ -10,10 +10,11 @@ import (
 	"github.com/openmeterio/openmeter/openmeter/billing/charges/usagebased"
 	billingrating "github.com/openmeterio/openmeter/openmeter/billing/rating"
 	"github.com/openmeterio/openmeter/openmeter/billing/rating/service/mutator"
+	"github.com/openmeterio/openmeter/openmeter/productcatalog"
 	"github.com/openmeterio/openmeter/pkg/currencyx"
 )
 
-func populateUsageBasedStandardLineFromRun(stdLine *billing.StandardLine, run usagebased.RealizationRun, runs usagebased.RealizationRuns) error {
+func populateUsageBasedStandardLineFromRun(stdLine *billing.StandardLine, run usagebased.RealizationRun, runs usagebased.RealizationRuns, unitConfig *productcatalog.UnitConfig) error {
```
mark-vass-konghq: Super nit, but these could be more readable in the future and does not have to pass too much parameters to the function itself
```suggestion
type populateStandardLineFromRunInput struct {
	Run        usagebased.RealizationRun
	Runs       usagebased.RealizationRuns
	UnitConfig *productcatalog.UnitConfig
}

func populateStandardLineFromRun(stdLine *billing.StandardLine, input populateStandardLineFromRunInput) error {
```

mark-vass-konghq (approved):

openmeter/billing/charges/usagebased/service/linemapper.go
rolosp: done

PR #4637, opened 2026-07-02
refactor: [OM-387] governance entitlement performance improvement 2

gergely-kurucz-konghq:
## Resolve only requested features in governance access queries

> Stacked on #4635

### What
`entitlement.Service.GetAccess` now accepts an optional feature filter; the governance query passes it so only the requested features' entitlements are resolved, instead of every entitlement the customer has.

### Why
`GetAccess` fanned out over **all** of a customer's entitlements — computing a balance (and a ClickHouse usage query) for each — even when the caller only needed a few. So a query asking about 10 features on a customer with 1000 entitlements did ~1000 entitlements' worth of work and discarded 990. Cost scaled with the customer's *total* entitlement count, not the request — a latent cliff for orgs with many features. (Every prior benchmark hid this by querying all seeded features.)

### How
- `GetAccess(ctx, ns, customerID, featureKeys ...string)` — variadic; **empty = all** (unchanged for the customer-access API and the all-features governance path). When set, entitlements are filtered to those feature keys **before** the balance fan-out, so balances and ClickHouse queries shrink to the requested count.
- Governance's `resolveAccess` passes the query's `FeatureKeys`.
- Added an `e2e` benchmark variant (`GOV_BENCH_OVERCOMPUTE`) that decouples seeded entitlements (N) from queried features (M) to expose/guard this, plus `make bench-governance-overcompute[-trace]`.

Verified via traces (1 customer, 50 seeded, 10 queried): per-customer `GetEntitlementBalance` and ClickHouse usage queries **50 → 10**, latency **68ms → 20.5ms**; ~100× at 10-of-1000. Non-filtering callers unchanged; non-bench tests green.

## Summary by CodeRabbit

* **New Features**
  * Added support for requesting only specific governance features, reducing unnecessary entitlement lookups.
  * Introduced new benchmark targets to exercise and trace over-compute behavior.
  * Added request-scoped memoization to speed up repeated lookups within a single operation.

* **Bug Fixes**
  * Governance access checks now return only the features requested by a query.
  * Improved reuse of customer and entitlement data during concurrent processing.

* **Tests**
  * Added coverage for memoization behavior, including concurrency and error handling.

### DIFF
--- a/openmeter/entitlement/service/service.go
+++ b/openmeter/entitlement/service/service.go
@@ -346,6 +346,12 @@ func (c *service) GetAccess(ctx context.Context, namespace string, customerId st
 		return entitlement.Access{}, nil
 	}
 
+	// This fans out over all of the customer's entitlements; each metered entitlement's
+	// DescribeOwner would otherwise re-fetch the same customer once per entitlement. Memoize the
+	// customer for the duration of this (read-only) call so the fan-out shares one lookup.
+	// Installed before errgroup.WithContext so the derived ctx carries the memo into every goroutine.
+	ctx = meteredentitlement.WithCustomerCache(ctx)
+
 	var result sync.Map
 
 	g, ctx := errgroup.WithContext(ctx)
--- a/openmeter/entitlement/service/service.go
+++ b/openmeter/entitlement/service/service.go
@@ -334,12 +334,24 @@ func (c *service) expandCustomers(ctx context.Context, entitlements []entitlemen
 	return custs, nil
 }
 
-func (c *service) GetAccess(ctx context.Context, namespace string, customerId string) (entitlement.Access, error) {
+func (c *service) GetAccess(ctx context.Context, namespace string, customerId string, featureKeys ...string) (entitlement.Access, error) {
 	now := clock.Now()
 
 	entitlements, err := c.GetEntitlementsOfCustomer(ctx, namespace, customerId, now)
 	if err != nil {
 		return entitlement.Access{}, err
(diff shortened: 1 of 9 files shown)


openmeter/entitlement/service/service.go
```diff
@@ -334,18 +334,36 @@ func (c *service) expandCustomers(ctx context.Context, entitlements []entitlemen
...
 	entitlements, err := c.GetEntitlementsOfCustomer(ctx, namespace, customerId, now)
 	if err != nil {
 		return entitlement.Access{}, err
 	}
 
+	// When the caller only needs a subset of features, drop the rest before the fan-out so we
+	// don't compute (and issue ClickHouse usage queries for) balances that would be discarded.
+	// Empty featureKeys means "all" — the fetch is a single query regardless, so filtering its
+	// result in memory is enough; the per-entitlement value calc is the cost this avoids.
+	if len(featureKeys) > 0 {
```
chrisgacsal, line 349: I would extend the `GetEntitlementsOfCustomer` method to allow fitlering by feature keys. It will require side-loading features in the adapter by it is way more efficient then overfetching from db and filtering in memory.

chrisgacsal: It seems the `ListEntitlements` on the adapter does have support for filtering by feature keys/ids so it would require a minimal effort.

chrisgacsal (approved):

chrisgacsal (changes requested):
See above.

PR #4642, opened 2026-07-03
feat: snapshot unit config on invoice lines - OM-398

rolosp:
## Overview

Snapshots `applied_unit_config` on usage-based invoice lines so re-rating converts from raw metered quantity using the config applied at billing time, even after the rate card changes. Adds one nullable jsonb column; converted value is recomputed on read.

Written on both billing paths (charges line-mapper + legacy line-engine via `AsNewStandardLine`), so `unitConfig` works even when lines route to the legacy engine. `StandardLine.GetUnitConfig()` now returns the snapshot.

## Summary by CodeRabbit

* **New Features**
  * Billing now supports persisting a billing-time `unit_config` snapshot (`NULL` when unset) and exposes it on usage-based invoice lines, including legacy gathering lines. Standard lines can surface the snapshot via `GetUnitConfig()`.
* **Bug Fixes**
  * Upserts no longer overwrite existing `unit_config` when omitted; conflict re-sync refreshes the snapshot per line.
  * Rating-time `unit_config` mutation now validates inputs to fail gracefully on corrupt data.
  * Invoice line upserts now also refresh `description` and `ratecard_discounts` (plus additional related fields).
* **Tests**
  * Added coverage for snapshot round-trip, propagation (including deep-copy), and legacy unit-config rating behavior.


openmeter/billing/adapter/gatheringlines.go
```diff
@@ -161,6 +161,19 @@ func (a *adapter) updateGatheringLines(ctx context.Context, lines billing.Gather
...
 				SetID(line.UBPConfigID)
 
+			// Persist the unit_config snapshot on the legacy gathering line. Like Price
+			// and FeatureKey above, it is written at line creation and not re-synced
+			// afterward: the subscription-sync update path (getPatchesForUpdateUsageBasedLine)
+			// only mutates service period / invoice-at / deleted-at, so an existing config
+			// row is never re-upserted with a different (or removed) unit_config. Left unset
+			// when nil (no unit_config rate card) so non-unit_config lines write NULL on
+			// create, matching the standard-line convention; because the value is never
+			// changed non-nil->nil on an existing row, ResolveWithNewValues preserving an
```
turip: This is fine, but I would prefer that we do:

UpdateUnitConfig() in the upsert items, as that reflects a codebug if we are not updating it
We shuold also do the same for all missing optional fields.

(Sorry, I know that this is housekeeping as we had this bad practice already, but this behavior will make updates special, and as price is mutable so unit config is via manual invoice edits or charges patching)

openmeter/billing/adapter/gatheringlines.go
```diff
@@ -319,6 +332,7 @@ func (a *adapter) mapGatheringInvoiceLineFromDB(schemaLevel int, dbLine *db.Bill
 			UBPConfigID: ubpLine.ID,
 			FeatureKey:  lo.FromPtr(ubpLine.FeatureKey),
 			Price:       lo.FromPtr(ubpLine.Price),
+			UnitConfig:  ubpLine.AppliedUnitConfig,
```
turip: Why not UnitConfig? this is a diverging pattern.

openmeter/billing/adapter/stdinvoicelines.go
```diff
@@ -518,6 +518,15 @@ func (a *adapter) upsertUsageBasedConfig(ctx context.Context, lineDiffs entitydi
 				SetNillableMeteredQuantity(line.UsageBased.MeteredQuantity).
 				SetNillableMeteredPreLinePeriodQuantity(line.UsageBased.MeteredPreLinePeriodQuantity)
 
+			// applied_unit_config is a billing-time snapshot with a write-once contract
+			// following the price precedent: the guarantee is behavioral — finalized lines
+			// are never re-upserted through this path, and no caller carries a different
+			// config for an existing line — not enforced at the SQL level. Left unset when
+			// nil so a conflicting re-upsert cannot clear an already-captured snapshot.
```
turip: charges/lines can be updated via invoice edit apis.

tools/migrate/migrations/20260703084504_applied-unit-config-snapshot.up.sql
```diff
@@ -0,0 +1,2 @@
+-- modify "billing_invoice_usage_based_line_configs" table
+ALTER TABLE "billing_invoice_usage_based_line_configs" ADD COLUMN "applied_unit_config" jsonb NULL;
```
turip: let's call it unit config imho

turip (approved):

PR #4731, opened 2026-07-16
feat(ledger): add custom currencies support (credit_only)

mark-vass-konghq:
## Summary

This PR adds the ledger foundation for custom-currency `credit_only` flows.

Custom credits can now be issued, allocated, corrected, and routed without
applying fiat precision or coalescing different managed currencies that share
the same code.

## What changed

- Added custom-currency support to ledger routes and persisted route metadata.
- Switched ledger routing and collection filters to composite currency identity
  (`CurrencyReference` / managed custom-currency ID).
- Preserved optional fiat source currency and cost basis for funded custom
  credits.
- Kept promotional custom credits source-less and without a cost basis.
- Added custom-currency handling to credit purchase lifecycle events:
  - credit purchase initiation;
  - payment authorization;
  - payment settlement;
  - promotional credit issuance.
- Preserved feature filters across the credit purchase lifecycle.
- Ensured collection separates custom currencies with the same code but
  different managed IDs.
- Added custom-currency support to allocation, correction, breakage, and
  credit-void ledger paths.
- Made payment and invoice amount creation the single owner of fiat rounding
  and passed the exact `FiatAmount` through lifecycle events, so payment records
  and ledger entries use the same amount without ledger-side recalculation.

## Credit-only behavior

For a custom-currency `credit_only` charge:

- credit balances remain denominated in the custom currency;
- uncovered usage becomes custom-currency exposure;
- promotional credits do not create a fiat source or cost basis;
- fiat-funded credits retain their source fiat currency and cost basis;
- the ledger does not convert custom credit consumption to fiat;
- invoice and subscription-level end-to-end behavior is not introduced by this
  PR.

## Tests

Added and updated coverage for:

- custom and fiat route validation;
- custom-currency precision handling;
- composite currency routing;
- same-code managed custom currencies;
- source-less and fiat-funded custom credits;
- credit purchase initiation, authorization, and settlement;
- feature-filtered custom-currency routes;
- promotional credit issuance and correction;
- fractional remainder/backfill allocation;
- exact fiat amount propagation across external-payment, invoice,
  authorization, and settlement flows;
- custom-currency credit-only allocation and correction.


mark-vass-konghq:
@greptileai
@coderabbit

openmeter/ent/schema/ledger_account.go
```diff
@@ -119,6 +120,28 @@ func (LedgerSubAccountRoute) Fields() []ent.Field {
 		field.String("routing_key").Immutable(),
 		// Literal routing values (denormalized from routing_key for query filtering; not FKs).
 		field.String("currency").Immutable(),
+		field.String("exchange_source_currency").
```
GAlexIHU: nit: lets not call `exchange_source_currency`, the "exchange" only makes sense for CC and even then it's not necessarily an exchange, let's call it cost_basis_currency or revenue_source_currency or source_fiat revenue_fiat, something of the like... (should be meaningful for FIAT as well)

openmeter/ledger/chargeadapter/creditpurchase.go
```diff
@@ -118,11 +150,18 @@ func (h *creditPurchaseHandler) OnCreditPurchasePaymentAuthorized(ctx context.Co
 		}
 	}
 
-	transactionGroup, err := h.ledger.CommitGroup(ctx, transactions.GroupInputs(
-		charge.Namespace,
-		annotations,
-		inputs...,
-	))
+	transactionGroupInput := transactions.WithIdempotencyKey(
```
GAlexIHU: this is not needed & i think its conceptually dangerous, remove it please

- groups roughly map to business events
- charges lifecycle defines/protects payment flow correctness, its not our business
- conceptually, why would the ledger _decline_ transactions? thats a business problem

openmeter/ledger/historical/idempotency.go
```diff
@@ -0,0 +1,190 @@
+package historical
```
GAlexIHU: why did you think this is needed?

openmeter/ledger/historical/idempotency.go
```diff
@@ -0,0 +1,190 @@
...
+	ledger.AnnotationSubscriptionItemID,
+	ledger.AnnotationFeatureID,
+	ledger.AnnotationCollectionType,
+	ledger.AnnotationCollectionSourceOrder,
+	ledger.AnnotationBreakageKind,
+	ledger.AnnotationBreakageRecordID,
+	ledger.AnnotationBreakagePlanID,
+}
+
+type transactionGroupFingerprintPayload struct {
```
GAlexIHU: i think your fingerprinting only works if the adapter is invoked with the exact same input. that scenario wouldn't even happen in production, when it would be called with semantically same inputs we'd just receive a conflict. this is also a good sign as to why this ledger lvl idempotency doesnt make sense... also, why would two identical groups be invalid (conceptually from ledger side)?

openmeter/ledger/chargeadapter/creditpurchase_customcurrency_test.go
```diff
@@ -0,0 +1,666 @@
...
+	authorizationInput := chargecreditpurchase.PaymentEventInput{
+		Charge:  charge,
+		EventAt: charge.CreatedAt.Add(15 * time.Minute),
+	}
+	authRef, err := env.handler.OnCreditPurchasePaymentAuthorized(t.Context(), authorizationInput)
+	require.NoError(t, err)
+	require.NotEmpty(t, authRef.TransactionGroupID)
+
+	// and: the same authorization event is replayed.
+	replayedAuthRef, err := env.handler.OnCreditPurchasePaymentAuthorized(t.Context(), authorizationInput)
```
GAlexIHU: this is not a meaningful testcase, what made you concerned with replay?

openmeter/ent/schema/ledger_transaction_group.go
```diff
@@ -22,6 +24,34 @@ func (LedgerTransactionGroup) Mixin() []ent.Mixin {
...
+		field.String("idempotency_scope").
+			Optional().
+			Nillable().
+			Immutable(),
+		field.String("idempotency_key").
+			Optional().
+			Nillable().
+			Immutable().
+			MaxLen(256),
+		field.String("input_fingerprint").
```
GAlexIHU: please get rid of these (see other comments for justifications)

GAlexIHU (changes requested):
please get rid "idempotency"

The ledger should correctly book whatever is requested, not invent business mechanisms. Why ever would 2 identical transactions be invalid in sequence? Imagine conceptually how bad it would be if you couldn't buy the same thing twice IRL

openmeter/ledger/chargeadapter/creditpurchase.go
```diff
@@ -92,21 +102,43 @@ func (h *creditPurchaseHandler) OnCreditPurchasePaymentAuthorized(ctx context.Co
...
+	if err != nil {
+		return ledgertransaction.GroupReference{}, fmt.Errorf("get settlement payment amount: %w", err)
+	}
+
+	var templates []transactions.TransactionTemplate
+	if charge.Intent.Currency.IsCustom() {
+		// Re-denominate what the customer owes from the custom-currency IOU
+		// into the fiat amount actually being paid, before authorizing payment
+		// against it. Authorize/Settle only ever move real (fiat) money.
+		templates = append(templates, transactions.ConvertCurrencyTemplate{
```
GAlexIHU, line 102: conversion should preserve everything present on the receivable account. you're dropping feature filters here, e.g.

```
# 100 ACME credits for api-calls costing $50:

+100 ACME  receivable [no feature]   ← conversion
-100 ACME  receivable [api-calls]    ← purchase

+$50 USD   receivable [api-calls]    ← authorization
-$50 USD   receivable [no feature]   ← conversion
```

openmeter/ledger/transactions/fx.go
```diff
@@ -2,61 +2,175 @@ package transactions
...
+
+	if err := ledger.ValidateCustomCurrency(t.TargetCurrency, t.TargetCustomCurrency); err != nil {
+		errs = append(errs, fmt.Errorf("target custom currency: %w", err))
 	}
 
-	if t.CreditPriority != nil {
-		if err := ledger.ValidateCreditPriority(*t.CreditPriority); err != nil {
-			return fmt.Errorf("credit priority: %w", err)
+	if sourceAmountErr == nil && targetAmountErr == nil && costBasisErr == nil && sourceCurrencyDefinition != nil {
+		expectedSourceAmount := t.TargetAmount.Mul(t.CostBasis).RoundBank(int32(sourceCurrencyDefinition.Subunits))
```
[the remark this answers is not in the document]
mark-vass-konghq: Fixed by changing `settlementPaymentAmount` to use `RoundBank` with the
  settlement fiat currency precision, matching the FX template validation and
  partial-correction paths. The global `RoundToPrecision` behavior was left
  unchanged.

  Added table-driven rounding coverage for values below, at, and above the
  midpoint, including a midpoint produced by multiplication:

  - 1.004 USD -> 1.00
  - 1.005 USD -> 1.00
  - 1.006 USD -> 1.01
  - 2 x 1.0025 EUR -> 2.00

GAlexIHU:
i think we should simplify the route currency representation before merging this.

right now `Route` has both `Currency` and `CustomCurrency`, meaning every caller has to know that the actual identity is the combination of those two fields. the invariance bug is a good example of how easy it is to accidentally use only the code.

we already have the right domain abstraction for this: `currencies.CurrencyReference`.

```go
type Route struct {
	Currency currencies.CurrencyReference

	CostBasis         *alpacadecimal.Decimal
	CostBasisCurrency *currencyx.Code // fiat only

	// other dimensions...
}
```

a hydrated reference already contains everything the ledger needs:

- code
- managed custom currency ID
- the resolved currency definition, including precision

the currencies pkg could own a stable serialization for ledger dimensions:

```go
func (r CurrencyReference) MarshalText() ([]byte, error)
func ParseCurrencyReference([]byte) (CurrencyReference, error)
```

roughly:

```text
USD
custom:v1:CREDITS:<managed-id>:2
```

a hydrated ref serializes into the ledger dimension. parsing that value reconstructs a hydrated ref from the embedded snapshot. no currency service lookup, lifecycle loading, or cost-basis loading involved.

this would let us keep one `currency` string dimension in storage and remove the ledger-local `CustomCurrencyIdentity` plus the separate custom ID / precision / version columns.

the result is:

- one typed currency field on the route
- identity comparison through `CurrencyReference.Equal`
- precision available locally
- immutable historical behavior
- no paired fields callers can accidentally treat independently

the only related cleanup is that `CurrencyReference.IsResolved()` currently conflates “currency definition is hydrated” with “cost-basis history was expanded”. the ledger only needs the first, so those states should probably be separated.

openmeter/ledger/transactions/fx.go
```diff
@@ -2,61 +2,175 @@ package transactions
...
+// correct reverses scope.Amount (denominated in the target/custom currency)
+// of an original currency conversion. scope.CostBasis must match the cost
+// basis the original conversion booked at; it is validated, not
+// reverse-engineered, because the rate is immutable for a given purchase.
+// A correction that reverses the whole original target amount reuses the
+// exact original source amount instead of recomputing it, so it absorbs any
+// rounding difference the original booking's precision left behind. Partial
+// corrections recompute the proportional source amount from the supplied
+// cost basis.
+func (t ConvertCurrencyTemplate) correct(scope CorrectionInput) ([]ledger.TransactionInput, error) {
```
GAlexIHU: i don't think we should implement this yet. credit purchases currently have no refund/payment-correction flow, so production code cannot reach it with the required exact fiat amount.

lets keep it explicitly unsupported until the charge layer provides both exact amounts.

openmeter/ledger/chargeadapter/creditpurchase.go
```diff
@@ -93,21 +89,44 @@ func (h *creditPurchaseHandler) OnCreditPurchasePaymentAuthorized(ctx context.Co
 	annotations := chargeAnnotationsForCreditPurchaseCharge(charge)
 	featureFilters := charge.Intent.FeatureFilters.Normalize()
 
+	settlementAmount, settlementCurrency, err := settlementPaymentAmount(charge.Intent.Currency, charge.Intent.Settlement, charge.Intent.CreditAmount, costBasis)
```
GAlexIHU: can we pass the exact `FiatAmount` through `PaymentEventInput` instead of recalculating it here?

charges already calculates and persists this value for both external and invoice settlements. the ledger should book that same amount so rounding has one owner and the payment realization always matches the ledger.

openmeter/ledger/customerbalance/service.go
```diff
@@ -190,14 +191,14 @@ func (i GetBalanceCurrenciesInput) pendingGrantAsOf() time.Time {
 
 func (i GetBalanceServiceInput) bookedRoute() ledger.RouteFilter {
 	route := i.featureRoute()
-	route.Currency = i.Currency
+	route.Currency = currencies.NewCurrencyReference(i.Currency)
```
GAlexIHU, line 196: can we add an explicit fiat-only guard here?

lets keep custom currency support in customerbalance out of scope for this PR

openmeter/billing/charges/creditpurchase/service/realizations/service.go
```diff
@@ -119,14 +119,18 @@ func (s *Service) AuthorizeExternalPayment(ctx context.Context, charge creditpur
...
 
-	fiatAmount := fiatCurrency.RoundToPrecision(
-		charge.Intent.CreditAmount.Mul(externalSettlement.CostBasis),
-	)
+	fiatAmount := charge.Intent.CreditAmount
+	if charge.Intent.Currency.IsCustom() {
+		fiatAmount = fiatCurrency.RoundToPrecision(
+			charge.Intent.CreditAmount.Mul(externalSettlement.CostBasis),
+		)
+	}
```
GAlexIHU: no, costbasis is still meaningful for fiat currencies, now FIAT purchases disregard costbasis, just remove the `IsCustom` branching

openmeter/currencies/currency_test.go
```diff
@@ -121,6 +124,80 @@ func TestCurrencyReference(t *testing.T) {
...
+		require.True(t, parsed.IsResolved())
+	})
+
+	t.Run("custom", func(t *testing.T) {
+		custom := currencytestutils.NewCustomCurrency(t, "CREDITS", 2)
+		reference := custom.Reference()
+
+		serialized, err := reference.MarshalText()
+		require.NoError(t, err)
+		require.Equal(t, "custom:v1:CREDITS:"+custom.ID+":2", string(serialized))
```
GAlexIHU: char ":" is unfortunately allowed for custom currency keying, lets use | as the delimiter, that is guarded against

GAlexIHU (changes requested):
Nice job! Lets fix these two small ones and then it should be good to merge

openmeter/billing/charges/creditpurchase/settlement.go
```diff
@@ -327,3 +327,27 @@ func (s Settlement) GetCostBasis() (alpacadecimal.Decimal, error) {
 		return alpacadecimal.Zero, fmt.Errorf("invalid settlement type: %s", s.t)
 	}
 }
+
+// GetCurrency returns the fiat currency real money settles in. Promotional
+// settlements never move real money, so they return an empty currency rather
+// than an error.
+func (s Settlement) GetCurrency() (currencyx.FiatCode, error) {
```
GAlexIHU, line 334: nit: shouldn't this return `*currencyx.FiatCode`, error instead? (using `lo.ToPtr`)

openmeter/currencies/testutils/currency/currency.go
```diff
@@ -17,3 +19,22 @@ func NewFiatCurrency[T ~string](t testing.TB, code T) currencies.Currency {
 
 	return currency
 }
+
+// NewCustomCurrency builds a custom currency value for tests that only need
+// the resolved calculator (code, precision) and a stable fixture managed ID,
+// not a persisted custom_currencies row.
+func NewCustomCurrency[T ~string](t testing.TB, code T, precision uint32) currencies.Currency {
```
GAlexIHU, line 26: why do we have a subpackage for these two methods? also why's that subpackage called `currency` (it can get unnecessarily confusing)... can these live in the parent testutils please?

openmeter/currencies/currency.go
```diff
@@ -79,6 +107,121 @@ func (r CurrencyReference) String() string {
...
+	return json.Marshal(struct {
+		Code             currencyx.Code `json:"code"`
+		CustomCurrencyID *string        `json:"custom_currency_id,omitempty"`
+	}{
+		Code:             r.Code,
+		CustomCurrencyID: r.CustomCurrencyID,
+	})
+}
+
+func ParseCurrencyReference(value []byte) (CurrencyReference, error) {
```
GAlexIHU, line 173: very very nit: this convenience is nice, but as we have `encoding.TextMarshaller` it could we nice if we implemented `encoding.TextUnmarshaller` and this just wrapped that (still this API is nicer at place of use)

we don't do XML, v3 AIP filters honor textmarshallers, there wouldn't be other interplay (+ we already implement marshaller)

GAlexIHU (approved):

PR #4813, opened 2026-07-29
fix(productcatalog): require billingCadence when unmarshaling usage based rate cards

### LINKED ISSUE #4812: Usage-based rate card with missing billing_cadence silently zero-defaults instead of erroring, surfaces as bare 500 on Kong Konnect
### OpenMeter Version

Observed on Kong Konnect (eu/us/au/global). Source references below are against `main` at 19b4fd5.

### Expected Behavior

`POST /v3/openmeter/plans` with a `usage_based` rate card that omits `billing_cadence` should return a `400` validation error, e.g. `rate_cards[0].billing_cadence: required for non-flat prices`.

### Actual Behavior

It returns a bare `500 Internal Server Error` with no detail body:

```json
{"type":"about:blank","title":"Internal Server Error","status":500,"instance":"urn:request:..."}
```

Adding `billing_cadence` to the same rate card makes the identical request succeed (`201`, plan created as `draft`). Reproduced live across `eu`/`us`/`au`/`global` base URLs, for `unit` pricing and for `graduated`/`volume` pricing with a `unit_price` tier component, independent of Stripe billing profile connection.

### Root cause — found

`FromAPIBillingRateCard` (`api/v3/handlers/plans/convert.go`) already rejects a usage-based rate card without a billing cadence, covering exactly the three price types that fail in production:

```go
case "unit", "graduated", "volume":
    if billingCadence == nil {
        return nil, fmt.Errorf("billing cadence is required for usage-based rate card %q", rc.Key)
    }
```

The check is correct. The error *type* is the bug. `create.go` and `update.go` propagate it unchanged, and `apierrors.GenericErrorEncoder` only handles `*apierrors.BaseAPIError`, two not-found types, and validation issues carrying an HTTP status attribute. A plain `fmt.Errorf` matches none of them, so the encoder returns `false`, nothing writes a response, and the request falls through as a `500` with an empty body.

`PUT` is affected the same way, since `update.go` propagates the converter error identically.

PR #4814 fixes this.

**Correction to earlier versions of this issue:** I first pointed at a silent zero-default in `Plan.UnmarshalJSON` (`openmeter/productcatalog/plan/serializer.go`), then said the 500 must live in Kong's closed layer. Both were wrong — I had not looked in `api/v3/` at all. The serializer issue is real but unrelated to this 500; it is filed separately as #4813.

### Steps To Reproduce

```bash
curl -X POST "https://eu.api.konghq.com/v3/openmeter/plans" \
  -H "Authorization: Bearer $KONNECT_TOKEN" \
  -H "Content-Type: application/json" \
  -d '{
    "name": "Repro", "key": "repro_unit",
    "currency": "USD", "billing_cadence": "P1M",
    "phases": [{
      "name": "default", "key": "default",
      "rate_cards": [{
        "name": "Tokens", "key": "tokens",
        "feature": {"id": "<any valid feature id>"},
        "price": {"type": "unit", "amount": "0.001"}
      }]
    }]
  }'
# -> 500
```

Adding `"billing_cadence": "P1M"` to the rate card returns `201`. Same for `graduated` with a `unit_price` tier.

### Additional Information

- A separate, unrelated defect turned up while tracing this: `Plan.UnmarshalJSON` in `openmeter/productcatalog/plan/serializer.go` silently zero-defaults `billingCadence` to `P0D` for usage based rate cards instead of erroring, producing a value `Validate()` always rejects. It is not on the path that causes this 500. I opened #4813 for it and closed it again to avoid leaving a speculative PR open; happy to revive it if you want it fixed on its own merits.
- Also reported to Kong support (support@konghq.com) on 2026-07-27 as a Konnect DX issue. Sample failing request instance IDs (org `Amr-Test`): `urn:request:bcd36764-3ee7-9e88-b225-b961869b64c0` (unit), `urn:request:4735454f-013c-9afd-90fa-e3ffbd56b679` (graduated).
- Separately, the [Pricing Models docs page](https://developer.konghq.com/metering-and-billing/pricing-models/) doesn't state that `billing_cadence` is required at the rate-card level for non-flat prices when creating plans via the API. The Konnect UI presumably always sets it, so only API-first users hit this. The Go SDK's generated comment ("Only valid for flat prices") is currently the only place the constraint is discoverable.

rixtrayker:
## Overview

`Plan.UnmarshalJSON` treats `billingCadence` as optional for usage based rate cards, mirroring the flat fee branch right above it:

```go
case productcatalog.UsageBasedRateCardType:
    urc := &productcatalog.UsageBasedRateCard{RateCardMeta: meta}
    if rcData.BillingCadence != nil {
        // ... parse and assign
    }
    rc = urc
```

But the two types differ. `FlatFeeRateCard.BillingCadence` is a `*datetime.ISODuration`, where nil legitimately means a one-time charge. `UsageBasedRateCard.BillingCadence` is a non-pointer `datetime.ISODuration`, so skipping the assignment doesn't leave it unset — it leaves the zero value `P0D` in place, and unmarshaling returns no error.

The result is a rate card that is invalid by construction: `UsageBasedRateCard.Validate()` rejects a zero cadence with `ErrBillingCadenceInvalidValue`, so the value can never be legal, yet the decode step happily produces it and defers the problem to whatever runs next.

This changes the usage based branch to return an error when the field is absent, instead of defaulting.

## Why this matters

Failing at decode keeps the invalid value from propagating. Callers that unmarshal a `Plan` and don't immediately call `Validate()` currently receive a `P0D` cadence with no indication anything is wrong, and how that surfaces depends on what touches it next.

This was found while investigating a bare `500` (no error body) from `POST /v3/openmeter/plans` on Kong Konnect when a `unit`-priced rate card omitted `billing_cadence` — adding the field made the identical request return `201`. Filed as #4812.

To be precise about scope: I could not reproduce that 500 from this repo. The v3 mapping path Konnect appears to use, `AsUsageBasedRateCard` in `openmeter/productcatalog/http/mapping.go`, takes `billingCadence` as a non-pointer `string` and calls `Parse()` unconditionally, which errors on `""` and is wrapped into a validation error by the `CreatePlan` handler — a correct `400`. So that path already behaves. The defect fixed here is the one I could demonstrate directly against this code: unmarshaling the payload below succeeds and yields `BillingCadence: P0D`.

I'm not claiming this patch is the fix for the Konnect 500 — that layer isn't in this repo. This is a real correctness bug on its own terms, and the same missing-required-field shape.

## Repro before this change

```go
var p plan.Plan
err := json.Unmarshal([]byte(`{
  "name": "Repro", "key": "repro_unit",
  "currency": "USD", "billing_cadence": "P1M",
  "phases": [{
    "name": "default", "key": "default",
    "rateCards": [{
      "type": "usage_based",
      "name": "Tokens", "key": "tokens",
      "featureKey": "tokens",
      "price": {"type": "unit", "amount": "0.001"}
    }]
  }]
}`), &p)

// err == nil
// p.Phases[0].RateCards[0].GetBillingCadence() == P0D
// ...but RateCards[0].Validate() reports
//   "billing cadence must be positive and 1 hour long duration at least"
```

After: unmarshal fails with `missing billing cadence for rate card "tokens": required for usage based rate cards`.

## Tests

Two cases added to the existing `TestPlanSerializationErrors` table:

- missing `billingCadence` on a usage based rate card now errors
- missing `billingCadence` on a flat fee rate card still succeeds, pinning that it stays optional

`go build ./...` and `go test ./openmeter/productcatalog/...` both pass.

## Note on the error type

The new error is a plain `fmt.Errorf`, consistent with the surrounding cases in this function (`unsupported rate card type`, `invalid billing cadence`). If you'd prefer these decode failures carry `models.NewGenericValidationError` so they map to a `400` rather than a `500` when this path is reached from an HTTP handler, I'm happy to change all of them in this PR or a follow-up — it seemed out of scope to alter the existing ones unprompted.

Closes #4812 if you consider the underlying defect the same; otherwise happy to leave that issue open for the Konnect-side behaviour.

## Summary by CodeRabbit

* **Bug Fixes**
  * Usage-based rate cards now require a valid billing cadence during plan import.
  * Clear validation errors are returned when the billing cadence is missing or invalid.
  * Flat-fee rate cards can still be imported without a billing cadence.

<h3>Greptile Summary</h3>

Require `billingCadence` when unmarshaling usage-based plan rate cards.
- Return a descriptive error instead of constructing a rate card with a zero cadence.
- Add coverage for the required usage-based cadence and the optional flat-fee cadence.

The PR appears safe to merge with no actionable defects identified.

The stricter deserialization behavior matches the usage-based rate-card invariant, preserves the distinct flat-fee semantics, and is covered by focused regression tests.

| Filename | Overview |
|----------|----------|
| openmeter/productcatalog/plan/serializer.go | Rejects usage-based rate cards without billing cadence while preserving flat-fee behavior. |
| openmeter/productcatalog/plan/serializer_test.go | Covers missing cadence for both usage-based and flat-fee rate cards. |

<sub>Reviews (1): Last reviewed commit: ["fix(productcatalog): require billingCade..."](https://github.com/openmeterio/openmeter/commit/2e45076c823f36c67c5f9186727c5f6a96c5e8ba) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48485778)</sub>

### DIFF
--- a/openmeter/productcatalog/plan/serializer.go
+++ b/openmeter/productcatalog/plan/serializer.go
@@ -199,13 +199,18 @@ func (p *Plan) UnmarshalJSON(data []byte) error {
 
 			case productcatalog.UsageBasedRateCardType:
 				urc := &productcatalog.UsageBasedRateCard{RateCardMeta: meta}
-				if rcData.BillingCadence != nil {
-					period, err := datetime.ISODurationString(*rcData.BillingCadence).Parse()
-					if err != nil {
-						return fmt.Errorf("invalid billing cadence for rate card %q: %w", rcData.Key, err)
-					}
-					urc.BillingCadence = period
+				// Unlike flat fee rate cards, billing cadence is required for
+				// usage based ones. Without this check a missing field would
+				// silently leave the zero value (P0D) in place.
+				if rcData.BillingCadence == nil {
+					return fmt.Errorf("missing billing cadence for rate card %q: required for usage based rate cards", rcData.Key)
 				}
+
+				period, err := datetime.ISODurationString(*rcData.BillingCadence).Parse()
+				if err != nil {
+					return fmt.Errorf("invalid billing cadence for rate card %q: %w", rcData.Key, err)
+				}
+				urc.BillingCadence = period
 				rc = urc
 
 			default:
--- a/openmeter/productcatalog/plan/serializer_test.go
+++ b/openmeter/productcatalog/plan/serializer_test.go
@@ -204,6 +204,45 @@ func TestPlanSerializationErrors(t *testing.T) {
 			wantErr:  true,
 			errMatch: "invalid billing cadence for rate card \"rate-card-1\"",
 		},
+		{
+			name: "missing billing cadence on usage based rate card",
+			json: `{
+				"namespace": "test",
+				"id": "plan-1",
+				"name": "Test Plan",
+				"phases": [{
+					"key": "phase-1",
+					"name": "Test Phase",
+					"rateCards": [{
+						"type": "usage_based",
+						"key": "rate-card-1",
+						"name": "Test Rate Card",
+						"featureKey": "tokens",
+						"price": {"type": "unit", "amount": "0.001"}
+					}]
+				}]
+			}`,
+			wantErr:  true,
+			errMatch: "missing billing cadence for rate card \"rate-card-1\"",
+		},
+		{
+			name: "missing billing cadence on flat fee rate card is allowed",
+			json: `{
+				"namespace": "test",
+				"id": "plan-1",
+				"name": "Test Plan",
+				"phases": [{
+					"key": "phase-1",
+					"name": "Test Phase",
+					"rateCards": [{
+						"type": "flat_fee",
+						"key": "rate-card-1",
+						"name": "Test Rate Card"
+					}]
+				}]
+			}`,
+			wantErr: false,
+		},
 	}
 
 	for _, tt := range tests {


rixtrayker:
Closing this one.

While tracing the 500 in #4812 further I found the actual cause in `api/v3/handlers/plans/` — a correct validation check returning a plain `fmt.Errorf` that the v3 error encoder doesn't recognise, so it falls through as a bodyless 500. That's fixed in #4814, verified against the live API.

The defect this PR addressed is real but separate: `Plan.UnmarshalJSON` silently zero-defaults `billingCadence` to `P0D` for usage-based rate cards rather than erroring, producing a value `Validate()` always rejects. It isn't on the path that caused the reported 500, and I'd rather not leave a speculative PR open on your tracker. Happy to reopen if you'd like it fixed on its own merits — the branch stays on my fork.

PR #4814, opened 2026-07-29
fix(v3/plans): return 400 instead of bare 500 when usage-based rate card omits billing_cadence

rixtrayker:
Fixes the bare `500` reported in #4812.

## The bug

`FromAPIBillingRateCard` already rejects a usage-based rate card without a billing cadence, and the check is correct — it covers exactly the three price types that fail in production:

```go
case "unit", "graduated", "volume":
    if billingCadence == nil {
        return nil, fmt.Errorf("billing cadence is required for usage-based rate card %q", rc.Key)
    }
```

The problem is the error type. `create.go` and `update.go` propagate the converter's error unchanged, and `apierrors.GenericErrorEncoder` handles only:

- `*apierrors.BaseAPIError`
- `*feature.FeatureNotFoundError` / `*meter.MeterNotFoundError`
- validation issues carrying an HTTP status attribute (`HandleIssueIfHTTPStatusKnown`)

A plain `fmt.Errorf` matches none of these. The encoder returns `false`, nothing writes a response, and the request falls through as a `500` with an empty body — so a straightforward missing-field mistake is indistinguishable from a server fault.

Note that `create.go` already uses `models.NewGenericValidationError` for the `unit_config` gate a few lines above, so the intent to return `400` here is clear; this path just misses it.

## The fix

Introduce a typed `ErrRateCardBillingCadenceRequired` and map it to a `400` at the handler boundary, where `ctx` is available for `apierrors.NewBadRequestError`.

The mapping uses `errors.As` rather than a type assertion because the error is wrapped twice with `%w` on the way up (rate card → phase → request). I checked that first: a plain type assertion does not survive the chain.

I put the mapping in a small helper shared by `create.go` and `update.go` instead of threading `ctx` through the four converter functions, which would have been a much larger diff for the same result. Happy to do it the other way if you'd prefer the converters own their status codes.

## Before / after

Against the live API (Konnect, `eu`), the payload from #4812:

```
POST /v3/openmeter/plans   ->  500  {"title":"Internal Server Error", ...}
```

with `billing_cadence` added to the rate card:

```
POST /v3/openmeter/plans   ->  201  (draft plan)
```

Confirmed the `500` for `unit`, `graduated` and `volume`. After this change the same request produces:

```json
{
  "type": "https://kongapi.info/konnect/bad-request",
  "status": 400,
  "title": "Bad Request",
  "detail": "Bad Request: phases[].rate_cards[].billing_cadence [required]: billing cadence is required for usage-based rate card \"sessions\"",
  "invalid_parameters": [{
    "field": "phases[].rate_cards[].billing_cadence",
    "rule": "required",
    "reason": "billing cadence is required for usage-based rate card \"sessions\"",
    "source": "body"
  }]
}
```

which matches the shape the endpoint already returns for schema violations.

## Tests

`api/v3/handlers/plans/errors_test.go`:

- all three usage-based price types produce the typed error, it survives `%w` wrapping, and the encoder maps it to `400`
- a flat-priced rate card without a cadence still converts cleanly, pinning that flat/free are unaffected
- unrelated errors pass through untouched, so other failures keep their own status mapping

`go build ./...` and `go test ./api/v3/... ./openmeter/productcatalog/...` pass.

## Relationship to #4813

#4813 fixes a separate silent zero-default in `Plan.UnmarshalJSON` in the productcatalog package. That one is real but is *not* the cause of this 500 — I initially thought it was and said so in #4812 before tracing this path, which I've since corrected there. The two are independent; this PR is the one that fixes the reported behaviour. #4813 can be taken, dropped, or merged separately on its own merits.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for usage-based rate cards missing billing cadence.
  * Create and update plan requests now return a clear `400 Bad Request` identifying `phases[].rate_cards[].billing_cadence`, including “required” rule details.
  * Flat rate cards without billing cadence remain supported.
* **Tests**
  * Added conversion and API-level coverage for billing cadence validation across supported usage-based pricing models and plan operations.

<h3>Greptile Summary</h3>

This PR converts missing billing cadence errors for usage-based plan rate cards into structured HTTP 400 responses.
- Introduces a typed conversion error for unit, graduated, and volume prices without `billing_cadence`.
- Maps that error at both create and update handler boundaries.
- Adds regression coverage for create, update, wrapped errors, all affected price types, and unaffected flat prices.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| api/v3/handlers/plans/convert.go | Replaces the untyped missing-cadence conversion error with a dedicated error that remains discoverable through wrapping. |
| api/v3/handlers/plans/create.go | Maps the typed conversion failure to a structured bad-request response during plan creation. |
| api/v3/handlers/plans/update.go | Applies the same structured bad-request mapping during plan updates. |
| api/v3/handlers/plans/convert_billingcadence_test.go | Covers affected price types, create and update handlers, wrapped-error matching, response structure, and the valid flat-price case. |

<sub>Reviews (4): Last reviewed commit: ["test(v3/plans): assert the full problem ..."](https://github.com/openmeterio/openmeter/commit/125f163338f244fb23e250a2cc57990492689b8e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48495244)</sub>


rixtrayker:
Thanks — addressed both points in 69c0e91.

**Handler-level coverage.** You were right that the test was weak: it called `asBadRequestIfConversionError` directly, so it would have stayed green if a handler stopped calling it. It now drives `CreatePlan` and `UpdatePlan` over HTTP via `httptest` and asserts the 400.

I checked the new test actually catches the regression rather than assuming it: reverting the `create.go` mapping makes it fail with the original production symptom —

```
expected: 400
actual  : 500
body: {"type":"about:blank","title":"Internal Server Error","status":500}
```

`UpdatePlan` is covered too, since it shares the conversion path and had the same defect.

**`t.Context()`.** Switched throughout.

The service is deliberately nil in the test handler — these requests must fail during conversion, so a non-nil service would only mask a regression that let one through.

`gofmt`, `go vet` and `go test ./api/v3/...` are clean.

rixtrayker:
Pushed f32da68 — I'd written the first version without reading `AGENTS.md` and `docs/development/type-conversions.md`, and it broke three of your conventions. Corrected:

- **Removed `errors.go`.** The naming doc assigns `api/v3/handlers/` to `convert.go` for API↔domain translation, so `ErrRateCardBillingCadenceRequired` now sits next to the converter that returns it.
- **Inlined the mapping** at both handlers with `apierrors.NewBadRequestError`, matching `customers/list.go`. The helper I'd extracted was a single-use pass-through called `asBadRequestIfConversionError` — `AGENTS.md` rejects pass-through wrappers, and `As…` is listed under forbidden naming patterns.
- **Dropped the two helper-only tests**; the handler-level tests cover that mapping through the real request path. Also inlined a trivial `jsonReader` wrapper, added given/when/then comments, and renamed the file to `convert_billingcadence_test.go`.

Re-ran the mutation check after restructuring — disabling the `create.go` branch still fails with the original bodyless `{"title":"Internal Server Error","status":500}`.

On scope: I looked at whether this should generalize, and I don't think it should in this PR. Making the encoder default unrecognised errors to 400 would be wrong — an untyped error genuinely is a server fault, and broadening the default would mask real internal failures. The narrower observation is that `api/v3/handlers/*/convert.go` returns a lot of untyped errors, some of which are input rejections that would 500 the same way (`addons/convert.go` looks like the closest sibling, since addons share rate cards). That seems like your call on prioritisation rather than something to attach here — happy to open a separate issue if it's useful.

rixtrayker:
Good catch — addressed in 125f163, and it found a real regression.

The HTTP tests now decode the problem response and assert the field path, `rule`, `source` and reason, rather than just the 400 plus a substring match. As you noted, the old assertion would have passed on any unrelated 400 that happened to mention `billing_cadence`.

Strengthening it immediately caught something the weak assertion couldn't see: when I inlined the mapping in the previous commit, I dropped `Rule: "required"` from both handlers. So the endpoint was returning an empty `rule` in the `invalid_parameters` entry. Restored, matching the convention in `customers/billing` and `subscriptions/create`.

Verified the new assertions actually bite by mutating the rule to `"invalid"` — both the create and update tests now fail on it, where before they passed.

Two of the three sites you cited (`errors_test.go` L106-107 and L156-159) no longer exist: that file and the helper-only tests were removed in f32da68 when I moved the mapping inline to follow the naming conventions in `docs/development/type-conversions.md`. Only the create and update handler tests remain, and both are covered now.

rolosp:
Hi @rixtrayker! Please check my response on the issue #4812 and how we opted to fix it in #4816.  Thanks again for reporting it! I will close this PR as we went with a different approach.

PR #4827, opened 2026-08-01
perf: make balance calculation linear in entitlement reset count

tothandras:
## Problem

`UsagePeriod.GetResetTimelineInclusive` was quadratic in an entitlement's usage-reset count, which grows without bound with account age. Balance calculation walks it several times per run, so old high-cadence entitlements blew the 30s message processing timeout. In one 16h production window the balance worker dead-lettered 294 messages, all `failed to describe owner <id>: context deadline exceeded` — the deadline always tripped before `engine.Run`, i.e. the whole budget went to metadata resolution rather than the balance math.

The failures were concentrated on ~16 entitlements in 2 namespaces that failed every hour, permanently. They are not slow; they are superlinear in a quantity that grows with wall-clock time.

Two independent causes, both loop-shaped:

- `Timeline.GetTimes()` allocates a fresh slice on every call, and was evaluated inside loop *conditions* — `usageperiod.go` (3 sites) and `engine/run.go` via `GetClosedPeriods()` (2 sites).
- `GetUsagePeriodInputAt`'s loop variable was captured by the returned closure, forcing a heap allocation **per iteration** of an O(n) scan that is itself called O(n) times. This was the dominant term: 88% of allocations.

The cost lands as **CPU, not memory**. The churn is transient garbage reclaimed continuously, so it never becomes RSS — it shows up as GC mark workers burning cores.

## Changes

**`perf(credit): iterations optimization`**
- `Timeline.Len()` — allocation-free, for use in loop conditions.
- `Timeline.LastIndexNotAfter()` — binary search over the ascending-sorted invariant `NewTimeline` guarantees, replacing the reverse linear scan in `GetUsagePeriodInputAt` and eliminating the closure capture.
- Hoist `GetClosedPeriods()` out of two loop conditions in `engine.Run`.

**`perf(credit): drop redundant owner fetch from the usage querier`**
`QueryUsage` called both `GetDefaultParams` and `DescribeOwner`, but at both (only) construction sites `GetDefaultParams` was `DescribeOwner(...).DefaultQueryParams` — literally so in the snapshot service, and closed over the same prefetched owner in `buildEngineForOwner`. Redundant since #2406 introduced it. On the snapshot usage-backfill path in `GetLatestValidAt` this halves the owner lookups, each of which is four uncached Postgres queries.

**`perf(credit): let InvalidateAfter use the balance snapshot index`**
`balancesnapshot_namespace_owner_id_at` is partial on `deleted_at IS NULL`, and Postgres only uses a partial index when the query predicate implies the index predicate. `InvalidateAfter` omitted it, so the planner fell back to the namespace-only index. Runs on every grant create and void.

## Numbers

Reset timeline resolution, median of 3 runs at `-benchtime=10x`, same machine, baseline measured in a worktree at the parent commit:

| resets | time | bytes/op | allocs/op |
|---|---|---|---|
| 100 | 1.28 ms → 0.19 ms | 3.3 MB → 81 KB | 12,135 → 1,430 |
| 1,000 | 104 ms → 1.8 ms | 309 MB → 869 KB | 1,021,132 → 14,040 |
| 8,760 (1 year hourly) | 7,484 ms → 16.6 ms | 23.5 GB → 7.8 MB | 76,926,253 → 122,745 |

That is **451× faster** and **~3,000× less allocation** at one year of hourly resets.

The scaling is the substantive result. Going from 1,000 to 8,760 resets (8.76× more): before, time grew **71.8×** ≈ 8.76² — quadratic. After, **9.1×** ≈ 8.76 — linear.

CPU profile at 1,000 resets:

| | before | after |
|---|---|---|
| total CPU samples | 3,940 ms (**163% CPU**) | 70 ms (34.6% CPU) |
| `gcBgMarkWorker` / `gcDrain` | 540 ms | below threshold |

A single-goroutine workload pulling 163% CPU is GC mark workers running on other cores — which is why this presents as CPU pressure with flat memory.

The benchmark used to produce these figures was intentionally not committed.

## Correctness

`LastIndexNotAfter` is covered by a differential test asserting it matches the previous reverse scan for every probe across every timeline size, including duplicate-timestamp ties (which must resolve to the last of the tied group to preserve the old semantics).

Green and unchanged: `openmeter/credit/...`, `openmeter/entitlement/...`, `pkg/timeutil/...`, `test/credits`, `test/entitlement/regression`, `test/billing`, `test/subscription`. `golangci-lint` reports 0 issues.

Note the OM-400 unit_config regime check in `GetLastValidSnapshotAt` is untouched — only the number of owner fetches changed, never the comparison.

## Follow-ups (not in this PR)

- The two `failed to describe owner %s` error strings in `helper.go` and `balance.go` are byte-identical, so logs alone cannot distinguish them.
- `snapshotEngineResult` takes `SELECT ... FOR UPDATE NOWAIT` in a transaction that commits before `saveSnapshot` runs, so the lock guards nothing while costing round-trips on every read.
- `ListInScopeEntitlements` materialises an entire namespace's entitlements before processing; `ProcessEntitlements` then walks them serially.
- Postgres `pool_max_conns` is unset, defaulting to `max(4, NumCPU)`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented already-deleted balance snapshots from being selected for invalidation.
  * Usage queries now consistently apply each owner’s configured default parameters.
  * Improved recurrence handling for usage periods, including boundary and reset scenarios.

* **Performance**
  * Optimized usage-period recurrence lookups to reduce unnecessary computation and memory usage.

* **Reliability**
  * Improved timeline handling for empty timelines, duplicate timestamps, and exact time boundaries.

<h3>Greptile Summary</h3>

The PR makes entitlement balance calculation linear in the reset count while reducing redundant owner lookups and enabling the balance-snapshot partial index.
- Adds allocation-free timeline length and upper-bound lookup operations.
- Reuses closed periods during engine execution instead of repeatedly materializing them.
- Uses the already-described owner’s default query parameters.
- Restricts snapshot invalidation to active snapshots.

The PR appears safe to merge.

No blocking failure remains; the previously reported comparator-contract issue is resolved by the monotonic predicate passed to sort.Search.

| Filename | Overview |
|----------|----------|
| pkg/timeutil/timeline.go | Adds allocation-free length access and a contract-correct upper-bound search over constructor-sorted timeline entries. |
| pkg/timeutil/timeline_test.go | Covers empty timelines, inclusive boundaries, duplicate timestamps, and differential equivalence with the former reverse scan. |
| openmeter/entitlement/usageperiod.go | Replaces allocating timeline reads and linear recurrence lookup while preserving effective-recurrence semantics. |
| openmeter/credit/engine/run.go | Materializes closed periods once and consistently uses that slice for iteration and final-period detection. |
| openmeter/credit/balance/usage.go | Derives streaming defaults from the owner already fetched for meter metadata. |
| openmeter/credit/balance/service.go | Removes the redundant owner-backed default-parameter callback from snapshot-service wiring. |
| openmeter/credit/helper.go | Simplifies owner-specific engine wiring after default query parameters moved to the existing owner description. |
| openmeter/credit/adapter/balance_snapshot.go | Limits invalidation to undeleted snapshots so the applicable partial database index can be used. |

<sub>Reviews (3): Last reviewed commit: ["style: comments"](https://github.com/openmeterio/openmeter/commit/3d523836141240861748e1e9c4481a6993ec150d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49324368)</sub>

**Context used:**

- Knowledge Base — [Credit Grants, Balances, and Entitlements](https://app.greptile.com/openmeter/-/custom-context/knowledge-base/openmeterio/openmeter/-/docs/credit-entitlement.md)

### DIFF
--- a/openmeter/credit/adapter/balance_snapshot.go
+++ b/openmeter/credit/adapter/balance_snapshot.go
@@ -28,7 +28,12 @@ func NewPostgresBalanceSnapshotRepo(db *db.Client) *balanceSnapshotRepo {
 func (b *balanceSnapshotRepo) InvalidateAfter(ctx context.Context, owner models.NamespacedID, at time.Time) error {
 	return entutils.TransactingRepoWithNoValue(ctx, b, func(ctx context.Context, rep *balanceSnapshotRepo) error {
 		return rep.db.BalanceSnapshot.Update().
-			Where(db_balancesnapshot.OwnerID(owner.ID), db_balancesnapshot.Namespace(owner.Namespace), db_balancesnapshot.AtGT(at)).
+			Where(
+				db_balancesnapshot.OwnerID(owner.ID),
+				db_balancesnapshot.Namespace(owner.Namespace),
+				db_balancesnapshot.AtGT(at),
+				db_balancesnapshot.DeletedAtIsNil(),
+			).
 			SetDeletedAt(clock.Now()).
 			Exec(ctx)
 	})
--- a/openmeter/credit/balance/service.go
+++ b/openmeter/credit/balance/service.go
@@ -34,15 +34,8 @@ func NewSnapshotService(conf SnapshotServiceConfig) SnapshotService {
 		SnapshotServiceConfig: conf,
 		// We build a custom UsageQuerier for our usecase here
 		UsageQuerier: NewUsageQuerier(UsageQuerierConfig{
-			StreamingConnector: conf.StreamingConnector,
-			DescribeOwner:      conf.OwnerConnector.DescribeOwner,
-			GetDefaultParams: func(ctx context.Context, ownerID models.NamespacedID) (streaming.QueryParams, error) {
-				owner, err := conf.OwnerConnector.DescribeOwner(ctx, ownerID)
-				if err != nil {
-					return streaming.QueryParams{}, err
-				}
-				return owner.DefaultQueryParams, nil
-			},
+			StreamingConnector:    conf.StreamingConnector,
+			DescribeOwner:         conf.OwnerConnector.DescribeOwner,
 			GetUsagePeriodStartAt: conf.OwnerConnector.GetUsagePeriodStartAt,
 		}),
 	}
--- a/openmeter/credit/balance/usage.go
+++ b/openmeter/credit/balance/usage.go
@@ -22,7 +22,6 @@ type UsageQuerier interface {
 type UsageQuerierConfig struct {
 	StreamingConnector    streaming.Connector
 	DescribeOwner         func(ctx context.Context, id models.NamespacedID) (grant.Owner, error)
-	GetDefaultParams      func(ctx context.Context, ownerID models.NamespacedID) (streaming.QueryParams, error)
 	GetUsagePeriodStartAt func(ctx context.Context, ownerID models.NamespacedID, at time.Time) (time.Time, error)
 }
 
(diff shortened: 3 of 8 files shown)


GAlexIHU (approved):
thx!

PR #4921, opened 2026-08-12
fix(entitlements): honor query params when listing customer entitlement grants

liormechlovich-kong:
## Problem

A customer reported that `GET /api/v2/customers/{customerIdOrKey}/entitlements/{entitlementIdOrFeatureKey}/grants` always returns 100 items regardless of `page`, `pageSize`, `limit`, or `offset`.

Confirmed: the handler's request decoder never copied the parsed query parameters into its request struct, so `request.Params` stayed a zero value and **every** documented query parameter was silently dropped:

- `page` → always 1
- `pageSize` → always 100
- `limit` / `offset` → ignored
- `order` / `orderBy` → ignored
- `includeDeleted` → always false

`totalCount` still reported the real total, so the response looked like paging should work while the rest of the grants were unreachable.

## Change

- Wire `params.Params` through the decoder.
- Select the pagination mode so the existing contract is preserved: `page`/`pageSize` wins when supplied, the deprecated `limit`/`offset` mode is entered only when one of those is actually sent, and a request carrying neither keeps the paginated default response (`page=1`, `pageSize=100`, real `totalCount`) this endpoint has always returned.
- `meteredentitlement.ListEntitlementGrantsParams` gains `IncludeDeleted`, `Limit`, and `Offset`, forwarded to the grant repository. `Validate()` aggregates field errors and rejects a request that bounds the result set with neither mode.
- `orderBy` now goes through `strcase.CamelToSnake` before becoming a `grant.OrderBy`. The raw API value (`createdAt`) matched no branch in the repository's ordering switch, so ordering was dropped even when the parameter was read.

### Two latent bugs this diff made reachable

Both were harmless while every request collapsed to page 1 with ordering forced to the default, and both bite once pagination and ordering actually work:

- **`orderBy=id` emitted no `ORDER BY`.** `GrantOrderBy` documents `id`, but `grant.OrderBy` had no such member, so the ordering switch fell through. Two `LIMIT/OFFSET` queries with no ordering can return the same grant on both pages or on neither. Added `grant.OrderByID` and the matching `db_grant.ByID` ordering. This also means `/api/v1/grants` stops rejecting `orderBy=id`, which its own enum documents.
- **The limit/offset branch never set `TotalCount`.** Clients that had been sending the previously-ignored `limit` would start receiving `totalCount: 0` beside a non-empty `items` array. The branch now counts matches before applying the window, which also settles `/api/v2/grants` — its response schema requires `totalCount` but it reported `0` in its default mode.

Note: `/api/v1/grants` returns a plain array when no page is given, so it now pays for a `COUNT` it discards. Happy to thread an `IncludeTotalCount` flag through `grant.ListParams` instead if reviewers prefer.

## Tests

- `openmeter/entitlement/service/grants_test.go` (new, Postgres-backed): page window with total count, limit/offset window with total count, `orderBy=id` ordering, unbounded query rejected, negative offset rejected.
- `e2e/entitlement_test.go`: `TestCustomerEntitlementGrantsPagination` drives the endpoint over HTTP for the empty-parameter default, `pageSize`-only, `limit`-only, explicit `page`, `page` winning over conflicting `limit`/`offset`, `limit`+`offset`, descending order, and a `page=0` rejection. The grants are created with amounts and effective dates running counter to their creation order, so an assertion on `createdAt` ordering cannot be satisfied by ordering on another column.

Verified locally: `go build ./...`, `go vet ./...`, `go vet -C e2e ./...`, `golangci-lint` on the changed packages, and `go test ./openmeter/entitlement/... ./openmeter/credit/...` against Postgres — all green. The e2e case has not been executed anywhere; it needs the live stack and no test workflow has triggered on this fork PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<h3>Greptile Summary</h3>

The PR wires customer-entitlement grant query parameters through the HTTP and service layers and completes pagination metadata and ID ordering support.
- Preserves page-based defaults while enabling deprecated limit/offset requests.
- Forwards deletion, ordering, and pagination options to the grant repository.
- Counts all matching grants before applying limit/offset and adds deterministic ID ordering.
- Adds service and E2E coverage for pagination modes, validation, counts, and ordering.

The PR appears safe to merge because no blocking failure remains.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| openmeter/entitlement/driver/v2/customer_metered.go | Propagates parsed query parameters and selects page or legacy limit/offset pagination while normalizing ordering fields. |
| openmeter/entitlement/metered/entitlement_grant.go | Extends and validates entitlement grant list parameters before forwarding them to the repository. |
| openmeter/credit/adapter/grant.go | Adds ID ordering and populates total counts for limit/offset queries. |
| openmeter/credit/grant/repo.go | Adds ID to the supported grant ordering contract. |
| e2e/entitlement_test.go | Adds HTTP coverage for default, page-based, legacy, conflicting, ordered, and invalid pagination requests. |
| openmeter/entitlement/service/grants_test.go | Adds PostgreSQL-backed coverage for grant windows, total counts, ordering, and validation. |

<a href="https://app.greptile.com/ide/claude-code?prompt=Greploop%20openmeterio%2Fopenmeter%20PR%20%234921%3A%20work%20through%20Greptile's%20open%20review%20comments%2C%20then%20keep%20reviewing%20and%20fixing%20until%20it%20comes%20back%20clean%20at%205%2F5%20with%20zero%20unresolved%20comments.%0AStart%20by%20reading%20the%20comments%20off%20the%20PR%20itself.%20On%20GitHub%2C%20use%20paginated%20%60gh%20api%20graphql%60%20to%20query%20%60pullRequest.reviewThreads%60%20with%20each%20thread's%20%60isResolved%60%20value%20and%20inline%20%60comments%60.%20%60gh%20pr%20view%20--comments%60%20only%20includes%20conversation%20comments%2C%20so%20do%20not%20use%20it%20as%20the%20findings%20source.%20They%20are%20not%20listed%20in%20this%20prompt%20on%20purpose%3A%20the%20PR%20is%20current%2C%20a%20pasted%20copy%20would%20not%20be.%20Skip%20anything%20already%20resolved%2C%20and%20if%20you%20judge%20a%20comment%20wrong%2C%20say%20so%20rather%20than%20changing%20code%20to%20satisfy%20it.%0A%0APrefer%20the%20Greptile%20CLI%2C%20which%20reviews%20the%20working%20tree%20with%20no%20push%20and%20no%20CI%20run.%20Fall%20back%20to%20PUSH%20LOOP%20only%20where%20a%20step%20below%20says%20to.%0A1.%20Run%20%60command%20-v%20greptile%60.%20Missing%3A%20go%20to%20PUSH%20LOOP%2C%20and%20pitch%20the%20CLI%20at%20the%20end.%0A2.%20Run%20%60greptile%20whoami%60.%20It%20exits%200%20even%20when%20signed%20out%2C%20so%20read%20the%20output%3A%20if%20it%20says%20%22Not%20signed%20in%22%2C%20ask%20the%20user%20to%20run%20%60greptile%20login%60%20%28it%20opens%20a%20browser%2C%20do%20not%20run%20it%20yourself%29%20and%20go%20to%20PUSH%20LOOP%20with%20no%20pitch.%0A3.%20Look%20for%20%60greploop%2FSKILL.md%60%20under%20%60.agents%2Fskills%2F%60%20or%20%60.claude%2Fskills%2F%60%2C%20both%20in%20the%20repo%20root%20and%20in%20the%20user's%20home%20directory.%20Found%3A%20follow%20that%20skill%2C%20never%20pushing%20between%20rounds%2C%20and%20you%20are%20done.%0A4.%20Missing%3A%20run%20%60greptile%20skills%20list%60.%20If%20it%20fails%20as%20an%20unknown%20command%2C%20this%20CLI%20predates%20the%20skill%2C%20so%20tell%20the%20user%20to%20run%20%60greptile%20update%60%20and%20go%20to%20PUSH%20LOOP%20with%20no%20pitch.%20Otherwise%20ask%20the%20user%20before%20running%20%60greptile%20skills%20install%20greploop%60%2C%20then%20follow%20the%20skill%20it%20writes.%0A%0APUSH%20LOOP%20%28only%20when%20a%20step%20above%20sends%20you%20here%29%3A%0A%0AGREPLOOP%20ENABLED%3A%20After%20fixing%20all%20issues%2C%20commit%20and%20push%20your%20changes.%20Greptile%20will%20automatically%20re-review%20the%20PR%20via%20webhook.%20Wait%20for%20the%20review%20to%20complete%2C%20then%20read%20the%20new%20review%20comments%20and%20confidence%20score.%20If%20the%20confidence%20score%20is%20below%205%2F5%20or%20there%20are%20unresolved%20comments%2C%20fix%20those%20issues%20and%20push%20again.%20Repeat%20until%20the%20confidence%20score%20is%205%2F5%20with%20zero%20unresolved%20comments.%20Maximum%205%20iterations.%20Do%20NOT%20open%20a%20new%20PR%20%E2%80%94%20keep%20pushing%20to%20the%20same%20branch.%0A%0AIF%20YOU%20ENDED%20UP%20ON%20PUSH%20LOOP%20BECAUSE%20THE%20CLI%20WAS%20MISSING%3A%20once%20the%20loop%20is%20done%2C%20print%20the%20block%20below%20and%20nothing%20else%20about%20the%20CLI.%20Reproduce%20it%20character%20for%20character%20inside%20a%20fenced%20code%20block%2C%20so%20the%20borders%20line%20up.%20Do%20not%20reword%20it%2C%20translate%20it%2C%20summarize%20it%2C%20add%20to%20it%2C%20or%20drop%20the%20box.%20Print%20it%20once%2C%20and%20do%20not%20bring%20the%20CLI%20up%20again%20later%20in%20the%20session.%0A%0A%E2%95%AD%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%95%AE%0A%E2%94%82%20%20You%20can%20run%20greploops%20faster%20locally%20with%20our%20CLI.%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%E2%94%82%0A%E2%94%82%20%20Install%20it%20at%20https%3A%2F%2Fwww.greptile.com%2Fcli%2C%20or%20I%20can%20install%20it%20for%20you.%20%20%E2%94%82%0A%E2%95%B0%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%94%80%E2%95%AF%0A%0AIf%20they%20take%20you%20up%20on%20it%2C%20install%20with%20%60npm%20install%20-g%20greptile%60%20%28or%20%60brew%20install%20greptileai%2Ftap%2Fgreptile%60%29%2C%20then%20%60greptile%20skills%20install%20greploop%60.%20Leave%20%60greptile%20login%60%20to%20them%2C%20it%20opens%20a%20browser.&repo=openmeterio%2Fopenmeter&pr=4921&platform=github">Fix All in Greploop</a>

<sub>Reviews (7): Last reviewed commit: ["fix(credit): order grants by id and coun..."](https://github.com/openmeterio/openmeter/commit/2dce89e5917f9749f11baaa5ee31d6d283d5e0b3) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=52612454)</sub>

**Context used:**

- Context used - AGENTS.md ([source](https://github.com/openmeterio/openmeter/blob/main/AGENTS.md))

### DIFF
--- a/e2e/entitlement_test.go
+++ b/e2e/entitlement_test.go
@@ -756,9 +756,6 @@ func TestEntitlementWithLatestAggregation(t *testing.T) {
 	})
 }
 
-// TestCustomerEntitlementGrantsPagination pins that the list endpoint honors the caller supplied
-// pagination. The handler used to drop every query parameter, capping callers at the first 100
-// grants with no way to reach the rest.
 func TestCustomerEntitlementGrantsPagination(t *testing.T) {
 	client := initClient(t)
 
@@ -812,7 +809,39 @@ func TestCustomerEntitlementGrantsPagination(t *testing.T) {
 		require.Equal(t, http.StatusCreated, resp.StatusCode(), "body: %s", resp.Body)
 	}
 
-	t.Run("Should honor pageSize", func(t *testing.T) {
+	t.Run("Should default to the first page", func(t *testing.T) {
+		// when no pagination parameters are sent at all
+		resp, err := client.ListCustomerEntitlementGrantsV2WithResponse(ctx, cust.Id, featureKey, &api.ListCustomerEntitlementGrantsV2Params{})
+		require.NoError(t, err)
+		require.Equal(t, http.StatusOK, resp.StatusCode(), "body: %s", resp.Body)
+		require.NotNil(t, resp.JSON200)
+
+		// then the documented page defaults apply, rather than the limit/offset mode that
+		// reports no page metadata
+		require.Len(t, resp.JSON200.Items, grantCount)
+		require.Equal(t, 1, resp.JSON200.Page)
+		require.Equal(t, 100, resp.JSON200.PageSize)
+		require.Equal(t, grantCount, resp.JSON200.TotalCount)
+	})
+
+	t.Run("Should default the offset when only a limit is sent", func(t *testing.T) {
+		// when only the deprecated limit is sent, the offset defaults to zero
+		resp, err := client.ListCustomerEntitlementGrantsV2WithResponse(ctx, cust.Id, featureKey, &api.ListCustomerEntitlementGrantsV2Params{
+			Limit:   lo.ToPtr(2),
+			OrderBy: lo.ToPtr(api.GrantOrderByCreatedAt),
+			Order:   lo.ToPtr(api.SortOrderASC),
+		})
+		require.NoError(t, err)
+		require.Equal(t, http.StatusOK, resp.StatusCode(), "body: %s", resp.Body)
+		require.NotNil(t, resp.JSON200)
+
+		// then the window starts at the first grant
+		require.Len(t, resp.JSON200.Items, 2)
+		require.Equal(t, 1.0, resp.JSON200.Items[0].Amount)
+		require.Equal(t, 2.0, resp.JSON200.Items[1].Amount)
+	})
+
+	t.Run("Should default the page when only a pageSize is sent", func(t *testing.T) {
 		// when a page smaller than the grant count is requested
 		resp, err := client.ListCustomerEntitlementGrantsV2WithResponse(ctx, cust.Id, featureKey, &api.ListCustomerEntitlementGrantsV2Params{
 			PageSize: lo.ToPtr(2),
--- a/e2e/entitlement_test.go
+++ b/e2e/entitlement_test.go
@@ -916,12 +916,13 @@ func TestCustomerEntitlementGrantsPagination(t *testing.T) {
 		require.Equal(t, http.StatusOK, resp.StatusCode(), "body: %s", resp.Body)
 		require.NotNil(t, resp.JSON200)
 
-		// then the remaining grants are returned in order; this mode runs no count query, so
-		// the page metadata stays zeroed, matching /api/v2/grants
+		// then the remaining grants are returned in order. The page fields stay zeroed because
+		// this mode has no page, but the count is real, so a caller walking the offset can
+		// still tell when it has seen every grant.
 		require.Equal(t, []float64{10, 20}, amounts(resp.JSON200.Items))
 		require.Zero(t, resp.JSON200.Page)
 		require.Zero(t, resp.JSON200.PageSize)
-		require.Zero(t, resp.JSON200.TotalCount)
+		require.Equal(t, grantCount, resp.JSON200.TotalCount)
 	})
 
 	t.Run("Should honor a descending order", func(t *testing.T) {
--- a/e2e/entitlement_test.go
+++ b/e2e/entitlement_test.go
@@ -795,18 +795,25 @@ func TestCustomerEntitlementGrantsPagination(t *testing.T) {
 		require.Equal(t, http.StatusCreated, resp.StatusCode(), "body: %s", resp.Body)
 	}
 
-	// given three grants created in ascending amount order, so that ordering by createdAt makes
-	// the returned window identifiable by amount
-	grantCount := 3
-	firstEffectiveAt := time.Now().Truncate(time.Minute)
+	// given three grants whose amounts identify them, and whose amounts and effective dates both
+	// run counter to their creation order. Ordering by createdAt therefore yields a sequence that
+	// ordering by amount or by effectiveAt could not produce, so the assertions below fail if
+	// orderBy is ignored or resolved to the wrong column.
+	amountsInCreationOrder := []float64{30, 10, 20}
+	grantCount := len(amountsInCreationOrder)
+	lastEffectiveAt := time.Now().Truncate(time.Minute).Add(time.Duration(grantCount) * time.Minute)
 
-	for i := 0; i < grantCount; i++ {
+	for i, amount := range amountsInCreationOrder {
 		resp, err := client.CreateCustomerEntitlementGrantV2WithResponse(ctx, cust.Id, featureKey, api.CreateCustomerEntitlementGrantV2JSONRequestBody{
-			Amount:      float64(i + 1),
-			EffectiveAt: firstEffectiveAt.Add(time.Duration(i) * time.Minute),
+			Amount:      amount,
+			EffectiveAt: lastEffectiveAt.Add(-time.Duration(i) * time.Minute),
 		})
 		require.NoError(t, err)
 		require.Equal(t, http.StatusCreated, resp.StatusCode(), "body: %s", resp.Body)
+	}
+
+	amounts := func(items []api.EntitlementGrantV2) []float64 {
+		return lo.Map(items, func(item api.EntitlementGrantV2, _ int) float64 { return item.Amount })
 	}
 
 	t.Run("Should default to the first page", func(t *testing.T) {
@@ -817,8 +824,8 @@ func TestCustomerEntitlementGrantsPagination(t *testing.T) {
 		require.NotNil(t, resp.JSON200)
 
 		// then the documented page defaults apply, rather than the limit/offset mode that
-		// reports no page metadata
-		require.Len(t, resp.JSON200.Items, grantCount)
+		// reports no page metadata, and the grants come back in the default createdAt order
+		require.Equal(t, amountsInCreationOrder, amounts(resp.JSON200.Items))
 		require.Equal(t, 1, resp.JSON200.Page)
 		require.Equal(t, 100, resp.JSON200.PageSize)
 		require.Equal(t, grantCount, resp.JSON200.TotalCount)
@@ -835,10 +842,8 @@ func TestCustomerEntitlementGrantsPagination(t *testing.T) {
 		require.Equal(t, http.StatusOK, resp.StatusCode(), "body: %s", resp.Body)
 		require.NotNil(t, resp.JSON200)
 
-		// then the window starts at the first grant
-		require.Len(t, resp.JSON200.Items, 2)
-		require.Equal(t, 1.0, resp.JSON200.Items[0].Amount)
-		require.Equal(t, 2.0, resp.JSON200.Items[1].Amount)
+		// then the window starts at the first created grant
+		require.Equal(t, []float64{30, 10}, amounts(resp.JSON200.Items))
 	})
 
 	t.Run("Should default the page when only a pageSize is sent", func(t *testing.T) {
@@ -853,9 +858,7 @@ func TestCustomerEntitlementGrantsPagination(t *testing.T) {
 		require.NotNil(t, resp.JSON200)
 
 		// then the page is capped at the requested size while the count covers every grant
-		require.Len(t, resp.JSON200.Items, 2)
-		require.Equal(t, 1.0, resp.JSON200.Items[0].Amount)
-		require.Equal(t, 2.0, resp.JSON200.Items[1].Amount)
+		require.Equal(t, []float64{30, 10}, amounts(resp.JSON200.Items))
 		require.Equal(t, 1, resp.JSON200.Page)
 		require.Equal(t, 2, resp.JSON200.PageSize)
 		require.Equal(t, grantCount, resp.JSON200.TotalCount)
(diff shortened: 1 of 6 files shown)


liormechlovich-kong:
Addressed the review feedback in 880f9486c:

- **`Validate()` should aggregate field errors** (CodeRabbit) — done. It now collects the page/limit/offset, customer, and entitlement errors and returns `models.NewNillableGenericValidationError(errors.Join(errs...))`. Worth noting for reviewers: the wrapped `*pagination.InvalidError` is still reachable via `errors.As` through the join, so the entitlement error encoder keeps mapping an invalid `page` to 400 rather than falling through to the generic branch.
- **Negative offset coverage** (CodeRabbit) — added a service subtest with a positive `Limit` and `Offset = -1`.
- **Legacy pagination contract in the e2e limit/offset case** (CodeRabbit) — now asserts the returned amounts in order (so a dropped `offset`, `order`, or `orderBy` fails the test) and that `page`, `pageSize`, and `totalCount` come back as `0`, which is what that mode returns since it runs no count query. Also pinned the ordering explicitly in the `page`/`pageSize` subtests and asserted their amounts.
- **`t.Context()`** (Greptile + CodeRabbit) — done.
- **`given`/`when`/`then` intent comments** (Greptile) — added to the new e2e and service subtests.

Re-verified locally: `go build ./...`, `go vet ./...`, `go vet -C e2e ./...`, `golangci-lint run openmeter/entitlement/...`, and `go test ./openmeter/entitlement/... ./openmeter/credit/...` against Postgres — all green. The e2e case still needs the live stack, so CI is its first real run.

openmeter/entitlement/driver/v2/customer_metered.go
```diff
@@ -68,16 +70,26 @@ func (h *entitlementHandler) ListCustomerEntitlementGrants() ListCustomerEntitle
...
-			})
+			}
+
+			// page/pageSize takes precedence over the deprecated limit/offset mode.
+			if request.Params.Page != nil || request.Params.PageSize != nil {
+				listParams.Page = pagination.NewPage(
+					lo.FromPtrOr(request.Params.Page, commonhttp.DefaultPage),
+					lo.FromPtrOr(request.Params.PageSize, commonhttp.DefaultPageSize),
+				)
+			} else {
```
GAlexIHU: defaulting is lost here, it used to be a default page response, now it defaults limit/offset with common page defaults.  also,`commonhttp.DefaultPageSize` might not be 100.

e2e/entitlement_test.go
```diff
@@ -755,3 +755,133 @@ func TestEntitlementWithLatestAggregation(t *testing.T) {
 		}, 2*time.Minute, time.Second)
 	})
 }
+
+// TestCustomerEntitlementGrantsPagination pins that the list endpoint honors the caller supplied
+// pagination. The handler used to drop every query parameter, capping callers at the first 100
+// grants with no way to reach the rest.
```
GAlexIHU: comment is untrue & unneeded

e2e/entitlement_test.go
```diff
@@ -755,3 +755,133 @@ func TestEntitlementWithLatestAggregation(t *testing.T) {
...
+		// when a page index below the documented minimum is requested
+		resp, err := client.ListCustomerEntitlementGrantsV2WithResponse(ctx, cust.Id, featureKey, &api.ListCustomerEntitlementGrantsV2Params{
+			Page:     lo.ToPtr(0),
+			PageSize: lo.ToPtr(2),
+		})
+		require.NoError(t, err)
+
+		// then the request is rejected rather than silently falling back to the first page
+		require.Equal(t, http.StatusBadRequest, resp.StatusCode(), "body: %s", resp.Body)
+	})
```
GAlexIHU, line 952: lets add an explicit case for expected defaulting

GAlexIHU (changes requested):
thx for the contrib!

liormechlovich-kong:
Thanks for the review — all three addressed in `bd66c719c`.

**Lost defaulting.** You're right, and it was a real regression: a request with no pagination parameters fell into the limit/offset branch and came back with `page`, `pageSize`, and `totalCount` all zeroed, where this endpoint has always returned a real paginated response. The deprecated mode is now entered only when `limit` or `offset` is actually sent, with `page`/`pageSize` still winning when both are supplied:

```go
if request.Params.Page == nil && request.Params.PageSize == nil &&
    (request.Params.Limit != nil || request.Params.Offset != nil) {
    // deprecated limit/offset mode
} else {
    // page mode with defaults
}
```

**`commonhttp.DefaultPageSize` might not be 100.** It is today, but the coupling was the problem — that constant is free to change without anyone noticing this endpoint drifted from the defaults its OpenAPI spec publishes. The four defaults are now stated locally next to the handler with a comment explaining why they are not borrowed. Happy to switch back to the shared constants if you'd rather keep one source of truth.

**Explicit defaulting case.** Added `"Should default to the first page"`, which sends an empty parameter object and asserts `page=1`, `pageSize=100`, and the real `totalCount`. Also added `"Should default the offset when only a limit is sent"`, and renamed the old `"Should honor pageSize"` case to `"Should default the page when only a pageSize is sent"` since that is what it actually pins.

**Test comment.** Removed.

Note the branch had `main` merged into it, so I rebased this commit on top of `c6481bcd7` rather than force-pushing over it. Re-verified after the rebase: `go build ./...`, `go vet` on both modules, `golangci-lint run openmeter/entitlement/...`, and `go test ./openmeter/entitlement/... ./openmeter/credit/...` against Postgres — all green. The e2e case still has not run anywhere; it needs the live stack, and no test workflow has triggered on this fork PR yet.

GAlexIHU (dismissed):

liormechlovich-kong:
CodeRabbit raised two more on the last commit, both fair — fixed in `1a5bcacb2`.

**The ordering fixture proved nothing.** Amounts, effective dates, and creation order all ascended together, so `orderBy=createdAt` + `ASC` produced `1, 2, 3` whether the handler ordered by `createdAt`, ordered by `effectiveAt`, or ignored `orderBy` entirely. The grants now carry amounts `30, 10, 20` in creation order with effective dates running the opposite way, so each candidate column yields a distinct sequence and only the requested one matches. Added a `DESC` case as well, since nothing covered the direction.

**Nothing covered page precedence.** Every case sent one mode or the other, so a handler that picked the legacy mode whenever `limit` or `offset` appeared would still have passed — which is close to the regression you caught. New case sends `page=2&pageSize=2` together with `limit=1&offset=0` and asserts the page-2 window plus non-zero page metadata.

Verification unchanged: build, vet on both modules, lint, and the entitlement + credit suites against Postgres are green. The e2e file still has not run anywhere — it needs the live stack, and no test workflow has triggered on this fork PR, so those assertions are unproven until someone approves a run.

liormechlovich-kong:
I ran a review pass over the whole diff and it surfaced two problems that the earlier rounds missed — both are cases where this PR *makes an existing latent bug reachable*. Fixed in `e001e5906`.

**`orderBy=id` produced an unordered paginated query.** `GrantOrderBy` documents `id` alongside `createdAt` and `updatedAt`, and the OpenAPI middleware lets it through, but `grant.OrderBy` has no `id` member — so `strcase.CamelToSnake("id")` matched no branch of the adapter's ordering switch and the query went out with no `ORDER BY`. That was harmless while every request collapsed to page 1. Now that pages are reachable, `?orderBy=id&page=1` followed by `page=2` issues two `LIMIT/OFFSET` queries with no ordering, and Postgres is free to return a grant on both pages or on neither. I added the `id` ordering (`grant.OrderByID` + `db_grant.ByID`) rather than rejecting a documented value. Side effect: `/api/v1/grants` currently 400s on `orderBy=id` because its validation reads `StrValues()` — it now accepts it, which I take to be the intended contract.

**`limit` callers would have started seeing `totalCount: 0`.** The adapter's limit/offset branch never set `TotalCount`. A client that had been sending the previously-ignored `limit=1000` used to get page 1 with a real count; after this PR it would get its window alongside `totalCount: 0`, and a `while offset < totalCount` loop would conclude there are no grants. The branch now counts matches before applying the window. This also settles `/api/v2/grants`, whose response schema requires `totalCount` but which reports `0` in its default mode.

**One tradeoff worth your call:** `/api/v1/grants` returns a plain array when no page is given, so it now pays for a `COUNT` it discards. I left it unconditional rather than threading an `IncludeTotalCount` flag through `grant.ListParams` — say the word if you'd rather have the flag.

Verified: build, vet on both modules, lint, and `go test ./openmeter/entitlement/... ./openmeter/credit/...` against Postgres, all green, including a new service case pinning `orderBy=id`. The e2e file remains unexecuted — it needs the live stack and no workflow has triggered on this fork PR.

GAlexIHU (approved):

PR #5031, opened 2026-08-31
feat: v3 edit subscription

rolosp:
## What
Add a patch-based **edit** endpoint to the v3 subscription API — `POST /subscriptions/{subscriptionId}/edit` — bringing v3 to parity with v1's subscription editing and unblocking the UI migration off v1. The endpoint applies an ordered batch of customizations to a running subscription and returns the full `BillingSubscription` expanded view, so callers see the post-edit phases → items pricing tree in the same response. All six operations are supported: `add_item`, `remove_item`, `add_phase`, `remove_phase`, `stretch_phase`, and `unschedule_edit`, each carried by a `type`-discriminated `SubscriptionEditOperation` union, with an optional top-level `timing` (immediate / next_billing_cycle). Includes the TypeSpec + regenerated OpenAPI/SDKs/Go types, the op→patch converter, and the handler/route wiring.

## Why
v3 could read the full subscription but not modify it — every edit still had to go through v1, which blocks the UI from moving off v1 entirely. `edit` is distinct from the existing `change` (a full plan swap): it applies a granular patch list — add/remove an item, add/remove/stretch a phase, or unschedule a pending edit — which is exactly what the UI's inline subscription editing needs. This lands that contract on v3 and wires it end to end.

## How
The domain engine already exists — `subscriptionworkflow.Service.EditRunning(ctx, id, []subscription.Patch, timing)` — so this is API-surface porting, not new domain logic. `edit.tsp` defines the `SubscriptionEdit` body (a `customizations[]` batch + optional `timing`), the discriminated operation union, and `SubscriptionPhaseCreate`; the `edit` op is added to the shared `SubscriptionsOperations` so both openmeter/konnect specs auto-propagate. A single converter, `FromAPIBillingSubscriptionEditOperation`, maps each operation to a `subscription.Patch`; `add_item` reuses `plans.FromAPIBillingRateCard` (the full rate-card mapping path), and durations parse through `datetime.ISODurationString`. `add_phase` creates an empty phase (no nested rate cards, matching v1). The handler mirrors `change.go` — default immediate timing, calls `EditRunning`, returns `ToAPIBillingSubscription` — and the route is registered behind the existing `featuregate` middleware (unit-config flags into ctx), consistent with `ChangeSubscription`. The change is additive/non-breaking: a brand-new endpoint using the already-shipped v3 subscription view.

Tests cover the full stack: a unit test exercising the op→patch converter for all six operations, an integration test that maps an `add_item` rate card and applies it through the real `EditRunning` workflow against Postgres, and an HTTP e2e test (`TestV3SubscriptionEdit`) that pins add_item / remove_item / not-found over the wire.

## Summary by CodeRabbit

- **New Features**
  - Added support for editing running subscriptions.
  - Apply ordered changes to items and phases, including adding, removing, stretching, and unscheduling edits.
  - Supports immediate or next-cycle changes.
  - Added JavaScript and Go client support and API documentation.
- **Validation**
  - Validates edit requests and supported configurations.
  - Subscriptions with add-ons cannot be edited.
- **Bug Fixes**
  - Adding an undated phase before an existing phase now returns a clear validation error instead of failing unexpectedly.

<h3>Greptile Summary</h3>

The PR adds the v3 subscription edit endpoint and generated client contracts for ordered item and phase customizations.
- Converts all six edit operations into existing subscription patches.
- Wires the endpoint through the v3 handler, server, OpenAPI specification, and generated SDKs.
- Adds validation that safely rejects an undated phase inserted before an existing phase.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| api/v3/handlers/subscriptions/convert.go | Converts each API edit-operation variant into the corresponding domain patch while preserving the intended nullable phase-start semantics. |
| api/v3/handlers/subscriptions/edit.go | Implements the v3 edit handler, timing default, workflow invocation, response conversion, and typed error mapping. |
| openmeter/subscription/patch/addphase.go | Rejects a phase without duration before phase-shifting arithmetic can dereference the absent duration. |
| api/spec/packages/aip/src/subscriptions/edit.tsp | Defines the edit request, operation union, phase creation model, and timing contract used by generated API artifacts. |
| e2e/subscriptions_edit_v3_test.go | Covers representative edit operations, not-found handling, and the intentional null phase-start behavior. |

<details open><summary><h3>Sequence Diagram</h3></summary>

```mermaid
sequenceDiagram
  participant Client
  participant API as V3 Edit Handler
  participant Converter
  participant Workflow as EditRunning
  participant Patch as Subscription Patches
  Client->>API: "POST /subscriptions/{id}/edit"
  API->>Converter: Convert ordered customizations
  Converter-->>API: []subscription.Patch
  API->>Workflow: EditRunning(id, patches, timing)
  Workflow->>Patch: Apply patches in order
  Patch-->>Workflow: Updated subscription specification
  Workflow-->>API: Edited subscription
  API-->>Client: Expanded BillingSubscription
```
</details>

<sub>Reviews (12): Last reviewed commit: ["test: the integration test didn&#39;t add mu..."](https://github.com/openmeterio/openmeter/commit/0a8576b9a304383759d739b6ee0f71e525f37e51) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58650658)</sub>


api/spec/packages/aip/src/subscriptions/edit.tsp
```diff
@@ -0,0 +1,247 @@
...
+
+  /**
+   * When the requested changes should take effect. Defaults to immediate.
+   */
+  @summary("Timing")
+  timing?: SubscriptionEditTiming = SubscriptionEditTimingEnum.Immediate;
+}
+
+/**
+ * The type of a subscription edit operation.
```
tothandras, line 32: Add enum types to docs

api/spec/packages/aip/src/subscriptions/edit.tsp
```diff
@@ -0,0 +1,244 @@
...
+   */
+  @summary("Description")
+  description?: string;
+
+  /**
+   * The ISO-8601 interval after the subscription start at which the phase begins.
+   * When null, the phase starts immediately after the subscription starts.
+   */
+  @summary("Start after")
+  start_after: Shared.ISO8601Duration | null;
```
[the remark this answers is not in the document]
rolosp: Deliberate v1 parity — v1's startAfter is duration | null (null = subscription start). On a running sub that's in the past, so the domain correctly returns 403 "cannot add phase in the past"; not a doc/runtime mismatch. Requiring non-null diverges from v1, and making null mean "at edit time" is a new semantic v1 never had. Keeping the contract. Added a Go e2e assertion (null start_after on an active sub → 403) to pin it.

api/v3/handlers/subscriptions/convert.go
```diff
@@ -287,3 +290,143 @@ func FromAPIBillingSubscriptionCreate(
...
+		apiP, err := op.AsBillingSubscriptionEditAddPhase()
+		if err != nil {
+			return nil, models.NewGenericValidationError(fmt.Errorf("failed to decode add_phase operation: %w", err))
+		}
+
+		// start_after is required but nullable: a null (or absent) value leaves the
+		// zero offset, i.e. the subscription start. The domain then rejects that as
+		// "in the past" on a running subscription, matching v1's null handling.
+		var startAfter datetime.ISODuration
+		if apiP.Phase.StartAfter.IsSpecified() && !apiP.Phase.StartAfter.IsNull() {
```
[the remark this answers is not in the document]
rolosp: Deliberate v1 parity — v1's startAfter is duration | null (null = subscription start). On a running subscription that offset is in the past, so the domain correctly returns 403 "cannot add phase in the past"; this is intended behavior, not a doc/runtime mismatch. Requiring a non-null offset diverges from v1's contract, and making null resolve at edit timing introduces a new semantic v1 never had. Keeping the contract as-is. Added a Go e2e assertion (null start_after on an active subscription → 403) to pin the behavior.

api/spec/packages/aip-client-javascript/src/models/schemas.ts
```diff
@@ -6064,6 +6180,22 @@ export const subscription = z
   })
   .describe('Subscription.')
 
+export const subscriptionEdit = z
+  .object({
+    customizations: z
+      .array(subscriptionEditOperation)
+      .max(100)
```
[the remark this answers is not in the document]
rolosp: generated code

api/v3/handlers/subscriptions/edit.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/openmeterio/openmeter/pkg/framework/transport/httptransport"
+	"github.com/openmeterio/openmeter/pkg/models"
+)
+
+type (
+	EditSubscriptionRequest struct {
+		ID             models.NamespacedID
+		Customizations []subscription.Patch
+		Timing         subscription.Timing
+	}
```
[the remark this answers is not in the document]
rolosp: Skipping — this would make edit.go inconsistent with every sibling handler in the package (change/create/cancel/get/list all decode inline with the same parse/resolve-namespace/guard/convert shape; change.go is the direct analog). The type switching the guideline targets is already extracted into the named FromAPIBillingSubscriptionEditOperation helper, and the only inline validation is a single non-empty customizations guard, which doesn't warrant an Input struct plus Validate() method. The callback just orchestrates named converters, so it stays consistent with the established pattern.

api/v3/handlers/subscriptions/convert.go
```diff
@@ -275,3 +278,136 @@ func FromAPIBillingSubscriptionCreate(
...
+			return nil, models.NewGenericValidationError(fmt.Errorf("failed to decode add_phase operation: %w", err))
+		}
+
+		var startAfter datetime.ISODuration
+		if apiP.Phase.StartAfter != nil {
+			startAfter, err = datetime.ISODurationString(*apiP.Phase.StartAfter).Parse()
+			if err != nil {
+				return nil, models.NewGenericValidationError(fmt.Errorf("failed to parse add_phase start_after: %w", err))
+			}
+		}
```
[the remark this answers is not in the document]
rolosp: fixed

api/v3/handlers/subscriptions/convert.go
```diff
@@ -275,3 +278,136 @@ func FromAPIBillingSubscriptionCreate(
...
+
+		var duration *datetime.ISODuration
+		if apiP.Phase.Duration != nil {
+			d, err := datetime.ISODurationString(*apiP.Phase.Duration).Parse()
+			if err != nil {
+				return nil, models.NewGenericValidationError(fmt.Errorf("failed to parse add_phase duration: %w", err))
+			}
+			duration = &d
+		}
```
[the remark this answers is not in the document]
rolosp: fixed

api/v3/handlers/subscriptions/convert.go
```diff
@@ -287,3 +290,143 @@ func FromAPIBillingSubscriptionCreate(
...
+		if apiP.Phase.StartAfter.IsSpecified() && !apiP.Phase.StartAfter.IsNull() {
+			sa, err := apiP.Phase.StartAfter.Get()
+			if err != nil {
+				return nil, models.NewGenericValidationError(fmt.Errorf("failed to read add_phase start_after: %w", err))
+			}
+			startAfter, err = datetime.ISODurationString(sa).Parse()
+			if err != nil {
+				return nil, models.NewGenericValidationError(fmt.Errorf("failed to parse add_phase start_after: %w", err))
+			}
+		}
```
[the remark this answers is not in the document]
rolosp: null to the subscription-start offset matches v1; on a running subscription that's in the past, so the 403 "cannot add phase in the past" is the correct response, not a bug. Rejecting null or resolving it at edit time were both considered and declined to keep the v1 contract. Behavior is covered by a Go e2e test (null start_after on an active sub → 403)

api/spec/packages/aip/src/subscriptions/edit.tsp
```diff
@@ -0,0 +1,244 @@
...
+enum RemovePhaseShifting {
+  Next: "next",
+  Prev: "prev",
+}
+
+/**
+ * The definition of a new subscription phase to add via an edit.
+ */
+@friendlyName("BillingSubscriptionPhaseCreate")
+model SubscriptionPhaseCreate {
```
gergely-kurucz-konghq, line 212: The legacy model contains this field:
```
  /**
   * The discounts on the plan.
   */
  @summary("Discounts")
  discounts?: Discounts;
```

It is missing from here. Is this intentional?

api/spec/packages/aip/src/subscriptions/operations.tsp
```diff
@@ -165,6 +166,22 @@ interface SubscriptionsOperations {
...
+   */
+  @post
+  @operationId("edit-subscription")
+  @summary("Edit subscription")
+  @route("/{subscriptionId}/edit")
+  edit(@path subscriptionId: Shared.ULID, @body body: SubscriptionEdit):
+    | Shared.UpdateResponse<Subscription>
+    | Common.NotFound
+    | Common.Conflict
+    | Common.ErrorResponses;
```
gergely-kurucz-konghq, line 184: In the legacy v1 definitions, this endpoint is using `PATCH /{subscriptionId}` semantics. Is the change to `POST /{subscriptionId}/edit` intentional?

api/v3/handlers/subscriptions/edit.go
```diff
@@ -0,0 +1,118 @@
...
+						},
+					},
+				)
+			}
+
+			patches := make([]subscription.Patch, 0, len(body.Customizations))
+			for idx, op := range body.Customizations {
+				p, err := FromAPIBillingSubscriptionEditOperation(op)
+				if err != nil {
+					return EditSubscriptionRequest{}, fmt.Errorf("customization at index %d: %w", idx, err)
```
gergely-kurucz-konghq: Error message is less descriptive than in the legacy counterpart. No "failed/required/missing" explanation. Consider improving this for consistency with general error message wording.

api/v3/handlers/subscriptions/edit.go
```diff
@@ -0,0 +1,118 @@
...
+				return EditSubscriptionResponse{}, err
+			}
+
+			return ToAPIBillingSubscription(view)
+		},
+		commonhttp.JSONResponseEncoderWithStatus[EditSubscriptionResponse](http.StatusOK),
+		httptransport.AppendOptions(
+			h.options,
+			httptransport.WithOperationName("edit-subscription"),
+			httptransport.WithErrorEncoder(editSubscriptionErrorEncoder()),
```
gergely-kurucz-konghq, line 98: Not in scope for this PR: I see that this is a pattern in the v3 handlers, but we should replicate what we had in v1, namely: have a common error encoder handling all the different domain errors in a single place. Will create a ticket for this as discussed with @tothandras.

api/v3/handlers/subscriptions/convert_edit_test.go
```diff
@@ -0,0 +1,177 @@
...
+		p, err := FromAPIBillingSubscriptionEditOperation(op)
+
+		// then: the patch is an add-item targeting the phase, with the item key
+		// derived from the rate card
+		require.NoError(t, err)
+		add, ok := p.(patch.PatchAddItem)
+		require.True(t, ok, "expected PatchAddItem, got %T", p)
+		require.Equal(t, "trial", add.PhaseKey)
+		require.Equal(t, "setup", add.ItemKey)
+		require.Equal(t, subscription.PatchOperationAdd, add.Op())
```
gergely-kurucz-konghq, line 49: Consider adding this assertion of the mapped patch operation type to the other test cases as well.

api/v3/handlers/subscriptions/edit_integration_test.go
gergely-kurucz-konghq, line 1: This file does not seem to be an appropriate integration test. It reimplements/transcribes the edit handler logic. If we delete the edit handler, this test still passes. So it does not exercise the edit handler implementation, and therefore does not cover it appropriately.

There is the e2e test, which better covers what this integration test seemingly intends to cover.

Suggestion: either drop `edit_integration_test.go` as subsumed by the e2e tests, or reduce it to what only it can cheaply do — the handler-level cases e2e doesn't reach (empty `customizations` 400, omitted `Timing`, the 409/403/400 encoder mapping) — using the `handler.EditSubscription().With(id).ServeHTTP(...)` pattern from `api/v3/handlers/currencies/handler_test.go:30`.

e2e/subscriptions_edit_v3_test.go
```diff
@@ -0,0 +1,270 @@
...
+		updated, err := c.Subscriptions.Edit(t.Context(), subscriptionID, v3sdk.SubscriptionEdit{
+			// add_phase then add_item: the phase must exist before it can be populated.
+			Customizations: []v3sdk.SubscriptionEditOperation{addPhase, addItem},
+			Timing:         &immediate,
+		})
+		c.requireStatus(http.StatusOK, err)
+		require.NotNil(t, updated)
+
+		assert.True(t, subscriptionHasItem(updated, newPhaseKey, item.Key),
+			"added phase %q with item %q must be present after add_phase", newPhaseKey, item.Key)
```
gergely-kurucz-konghq, line 230: Should we assert that the phase was added at the right position?

gergely-kurucz-konghq (changes requested):
Some clarification and some modification requested.

api/v3/server/routes.go
```diff
@@ -121,6 +121,14 @@ func (s *Server) ChangeSubscription(w http.ResponseWriter, r *http.Request, subs
 		)).With(subscriptionId).ServeHTTP(w, r)
 }
 
+func (s *Server) EditSubscription(w http.ResponseWriter, r *http.Request, subscriptionId api.ULID) {
+	s.subscriptionsHandler.EditSubscription().
+		Chain(featuregate.NewMiddleware[subscriptionhandler.EditSubscriptionRequest, subscriptionhandler.EditSubscriptionResponse](
+			s.NamespaceDecoder.GetNamespace,
+			s.FeatureGate,
+		)).With(subscriptionId).ServeHTTP(w, r)
+}
```
gergely-kurucz-konghq, line 126: It is not clear from the PR why this context decoration is needed here. Is this for future proofing for a subsequent change? Or is this not really needed here?

api/spec/packages/aip/src/subscriptions/edit.tsp
rolosp: I think the v1 version isn't used actually. The mapper never reads it and the discount is tided to rate card instead of the phase.

api/spec/packages/aip/src/subscriptions/operations.tsp
rolosp: This is closer to the AIP style of v3. `POST /{subscriptionId}/<verb>` style is a custom action, not a partial-resource replacement. I feel this is consistent with the AIP v3 design, but I'm open to suggestions.

api/v3/handlers/subscriptions/edit.go
rolosp: fixed

api/v3/handlers/subscriptions/edit.go
rolosp: understood, out of scope

api/v3/handlers/subscriptions/convert_edit_test.go
rolosp: fixed

api/v3/handlers/subscriptions/edit_integration_test.go
rolosp: dropped it

e2e/subscriptions_edit_v3_test.go
rolosp: extended it with the position check

api/v3/server/routes.go
rolosp: this isn't needed here, it doesn't do anything, it was wired because of the feature gate but in the end i realized that we don't need to gate this. I'll drop it

api/spec/packages/aip/src/subscriptions/operations.tsp
gergely-kurucz-konghq: Thanks for the explanation. It did not seem like a "classic" patch endpoint, since the input is not entity fields of the resource to add/remove/update/replace, but a list of "operations" representing those changes, so maybe more like an RPC. Just wanted to make sure. I am ok with it.

api/spec/packages/aip/src/subscriptions/edit.tsp
gergely-kurucz-konghq: Got it, just wanted to make sure the difference is intentional.

gergely-kurucz-konghq (approved):

PR #5034, opened 2026-08-31
fix(chart): configure bundled PostgreSQL SSL mode

### LINKED ISSUE #4951: Helm chart: bundled Postgres DSN has no sslmode param — migrations fail with "pq: SSL is not enabled on the server"
## Bug

When `postgresql.enabled=true`, the chart hardcodes the app's Postgres URL in `templates/configmap.yaml`:

```
postgres:
  autoMigrate: migration
  url: postgres://application:application@{{ include "openmeter.fullname" . }}-postgres:5432/application
```

The URL has no `sslmode` parameter. Since the migration-based `autoMigrate` strategy (golang-migrate) opens this DSN with `lib/pq`, whose default is `sslmode=require`, and the bundled bitnami PostgreSQL ships with TLS disabled, **every service crashloops at startup**:

```
failed to create migrator: failed to open database: pq: SSL is not enabled on the server
```

The runtime connections (pgx, default `sslmode=prefer`) fall back to plaintext, which is why this only surfaces once `autoMigrate: migration` became the default — upgrades from ent-automigrate chart versions (e.g. helm chart 1.0.0-beta.213 → 1.0.0-beta.232) break out of the box.

There is also no values-level escape hatch: the configmap template builds `$cfg` with `mergeOverwrite $config $valuesConfig`, so a user-supplied `config.postgres.url` (with `?sslmode=disable`) is overwritten by the generated one.

## Repro

1. `helm install openmeter oci://ghcr.io/openmeterio/helm-charts/openmeter --version 1.0.0-beta.232` (defaults, bundled Postgres)
2. Watch api/workers crashloop with the error above.

## Workaround

Enable TLS on the bundled Postgres so `sslmode=require` succeeds:

```yaml
postgresql:
  tls:
    enabled: true
    autoGenerated: true
```

## Suggested fix

Append `?sslmode=disable` to the generated DSN for the bundled (non-TLS) Postgres, or let `config.postgres.url` win over the generated value.

Chart/app version: 1.0.0-beta.232 (also present on current main). Environment: RKE2 + microk8s, helm 3-style OCI install.

anishfyi:
## Overview

Configure an explicit PostgreSQL SSL mode in the Helm-generated DSN for the bundled PostgreSQL dependency. The default non-TLS deployment now uses `sslmode=disable`, while deployments that enable the subchart's TLS setting use `sslmode=require`.

Fixes #4951

## Notes for reviewer

Validation:

- `make lint-helm`
- `helm template` with the default values renders `sslmode=disable`
- `helm template` with `postgresql.tls.enabled=true` and certificate values renders `sslmode=require`

This only changes the chart-generated DSN for the bundled PostgreSQL dependency. User-supplied PostgreSQL configuration remains unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * PostgreSQL connection URLs now automatically use the appropriate SSL mode based on the configured TLS setting.
  * TLS-enabled connections require SSL, while non-TLS connections explicitly disable it.

<h3>Greptile Summary</h3>

The chart-generated DSN for bundled PostgreSQL now explicitly selects its SSL mode from the subchart’s TLS setting.
- Renders `sslmode=disable` for the default non-TLS deployment.
- Renders `sslmode=require` when bundled PostgreSQL TLS is enabled.
- Leaves user-supplied PostgreSQL configuration unchanged.

The PR appears safe to merge.

No blocking failure remains.

| Filename | Overview |
|----------|----------|
| deploy/charts/openmeter/templates/configmap.yaml | Adds a safe default lookup for bundled PostgreSQL TLS and maps the supported Boolean setting to the corresponding DSN SSL mode. |

<sub>Reviews (3): Last reviewed commit: ["fix(chart): configure bundled PostgreSQL..."](https://github.com/openmeterio/openmeter/commit/2dcb754f28379ec3c0c203a9ee3a2d723eeb565e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=58684877)</sub>

### DIFF
--- a/deploy/charts/openmeter/templates/configmap.yaml
+++ b/deploy/charts/openmeter/templates/configmap.yaml
@@ -21,9 +21,10 @@ aggregation:
 
 {{/* Overwrite postgres config if self-hosted postgres is used */}}
 {{- if .Values.postgresql.enabled -}}
+{{- $postgresTLSEnabled := dig "tls" "enabled" false .Values.postgresql -}}
 postgres:
   autoMigrate: migration
-  url: postgres://application:application@{{ include "openmeter.fullname" . }}-postgres:5432/application
+  url: postgres://application:application@{{ include "openmeter.fullname" . }}-postgres:5432/application?sslmode={{ ternary "require" "disable" $postgresTLSEnabled }}
 {{- end }}
 
 {{/* Overwrite svix config if self-hosted svix is used */}}


turip (approved):

deploy/charts/openmeter/templates/configmap.yaml
```diff
@@ -21,9 +21,10 @@ aggregation:
 
 {{/* Overwrite postgres config if self-hosted postgres is used */}}
 {{- if .Values.postgresql.enabled -}}
+{{- $postgresTLSEnabled := dig "tls" "enabled" false .Values.postgresql -}}
 postgres:
   autoMigrate: migration
-  url: postgres://application:application@{{ include "openmeter.fullname" . }}-postgres:5432/application
+  url: postgres://application:application@{{ include "openmeter.fullname" . }}-postgres:5432/application?sslmode={{ ternary "require" "disable" $postgresTLSEnabled }}
```
turip, line 27: Looks good, but I would prefer if we add a .Values.postgres.sslMode instead into values with default value being prefer and use that value here.

turip (changes requested):

PR #5055, opened 2026-09-02
feat: support inline plans on subscriptions

rolosp:
## What

Add inline (custom) plan support to the v3 subscription API — `POST /subscriptions` and `POST /subscriptions/{subscriptionId}/change` now accept a `custom_plan` (rate cards defined in the request) as a sibling to the published-plan `plan` reference, bringing v3 to parity with v1's custom subscription create/change. Exactly one of `plan` / `custom_plan` is required. This is also the vehicle for attaching custom currencies to a subscription, which live at the rate-card level. Includes the TypeSpec + regenerated OpenAPI/SDKs/Go types, the inbound converter, handler dispatch, and a service-layer feature gate.

## Why

v3 create/change could only reference a published plan — the handler explicitly rejected anything else — so inline/custom plans (and the custom currencies that attach per rate card) had no path on v3, and every such flow still had to go through v1. v1 already has `CustomSubscriptionCreate` / `CustomSubscriptionChange`; this lands that contract on v3. Adding a sibling `custom_plan` rather than overloading `plan` keeps the existing plan-reference semantics untouched, mirroring v1's `plan` vs `customPlan` model.

## How

The domain already supports inline plans — `plansubscription.PlanInput.FromInput(*plan.CreatePlanInput)` flows through the same `service.Create`/`Change` path as the plan-ref case — so this is API-surface porting, not new domain logic. `subscription.tsp` adds `SubscriptionCustomPlan` (the plan create shape minus key/version), makes `plan` optional, and adds the `custom_plan` sibling; `SubscriptionChange` inherits it. A single converter, `FromAPIBillingSubscriptionCustomPlan`, maps the inline body to `plan.CreatePlanInput`, reusing the plans phase/rate-card converters (`currencyx.Code`, not `FiatCode`, is what admits custom currency). The handlers branch on presence (`custom_plan` → `FromInput`, `plan` → lookup + `FromRef`); the exactly-one rule (incl. reject-both) is enforced in the domain's `PlanInput.Validate()`. Custom currency is gated at the service layer (the credits featuregate flag) before currency resolution, so a disabled deployment returns a clear "not enabled" error. Currency-resolution errors already map to 400 through the existing encoder — no new error encoder needed. The change is additive/non-breaking at the HTTP layer.

Tests cover the full stack: unit tests for `PlanInput.Validate` (both/neither/each) and the inline converter (asserts a custom currency survives mapping), a Postgres integration test that the custom-currency gate fires before resolution, and an HTTP e2e test (`TestV3SubscriptionCreateCustomPlan`) pinning inline-create → 201, neither → 400, and change plan → inline → 200 over the wire.

## ⚠️ Breaking change (Go SDK, source-level)

Making `plan` optional flips `SubscriptionCreate.Plan` / `SubscriptionChange.Plan` from a value to `*SubscriptionChangePlan` in `api/v3/api.gen.go` and the `api/v3/client` SDK. The wire format is unchanged (`plan` is still an optional JSON object), but this is source-breaking for Go SDK consumers, who must now pass `Plan: &SubscriptionChangePlan{...}`. It's inherent to the sibling design — `plan` has to be omittable for the `plan` / `custom_plan` XOR. This PR already updates the in-repo `e2e` call sites.

## Summary by CodeRabbit

- **New Features**
  - Subscriptions can be created or changed with inline custom plans, including pricing phases, billing cadence, currency, and pro-rating settings.
  - Existing published plan references remain supported.
  - Custom currencies are supported for inline plans when enabled.

- **Validation**
  - Requests must provide exactly one plan source: a published plan reference or an inline custom plan.
  - Clear validation errors are returned when neither or both are supplied, or when custom currencies are unavailable.

- **API Updates**
  - Client models now include inline custom plan definitions for subscription requests.

<h3>Greptile Summary</h3>

The PR adds inline custom-plan support to v3 subscription creation and plan changes, including custom currencies and generated API clients.
- Adds the `custom_plan` request contract alongside optional published-plan references.
- Converts inline API plans into service-layer plan inputs.
- Resolves inline feature and currency references before subscription construction.
- Gates custom-currency use when credits support is disabled.
- Adds validation, converter, integration, and end-to-end coverage.

The PR appears safe to merge.

The previously reported blocking failure is fixed: both inline create and change resolve ID-only feature references before subscription specification and entitlement construction, and no blocking failure remains.

| Filename | Overview |
|----------|----------|
| api/v3/handlers/subscriptions/convert.go | Converts the new inline custom-plan API representation into the existing domain plan input. |
| openmeter/productcatalog/subscription/service/create.go | Prepares inline plans by gating custom currencies and resolving feature and currency references before creation. |
| openmeter/productcatalog/subscription/service/change.go | Applies the same inline-plan preparation before changing an existing subscription. |
| openmeter/productcatalog/subscription/service/plan.go | Canonicalizes feature references across all phases of an inline plan. |
| api/spec/packages/aip/src/subscriptions/subscription.tsp | Defines the optional published-plan reference and sibling inline custom-plan request shape. |

<details open><summary><h3>Flowchart</h3></summary>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart LR
  Request[Create or change request] --> Choice{Plan source}
  Choice -->|Published plan| Lookup[Resolve published plan]
  Choice -->|Custom plan| Convert[Convert inline plan]
  Convert --> Features[Resolve feature IDs and keys]
  Features --> Currencies[Gate and resolve currencies]
  Lookup --> Build[Build subscription specification]
  Currencies --> Build
  Build --> Apply[Create or change subscription]
```
</details>

<sub>Reviews (6): Last reviewed commit: ["test: fix test after subscription edit s..."](https://github.com/openmeterio/openmeter/commit/bb1dbc51c63a65ee0b8103e44cde24fd573fd6ad) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=59488299)</sub>

<details><summary><h4>Context used (3)</h4></summary>

- Knowledge Base — [Product catalog and subscriptions](https://app.greptile.com/openmeter/-/custom-context/knowledge-base/openmeterio/openmeter/-/docs/product-catalog-and-subscriptions.md)
- Knowledge Base — [Subscription lifecycle](https://app.greptile.com/openmeter/-/custom-context/knowledge-base/openmeterio/openmeter/-/docs/subscription-lifecycle.md)
- Knowledge Base — [API and runtime composition](https://app.greptile.com/openmeter/-/custom-context/knowledge-base/openmeterio/openmeter/-/docs/api-and-runtime-composition.md)
</details>


api/spec/packages/aip/src/subscriptions/subscription.tsp
```diff
@@ -113,9 +124,14 @@ model SubscriptionCreate {
...
-   * The plan reference of the subscription.
+   * A reference to a published plan the subscription is created from.
+   *
+   * Exactly one of `plan` or `custom_plan` must be provided. Use `plan` to base the
+   * subscription on an existing published plan; use `custom_plan` to define the plan
+   * inline.
    */
-  plan: {
+  @visibility(Lifecycle.Create)
+  plan?: {
```
gergely-kurucz-konghq, line 134: Wondering why are we not using a dedicated "PlanReference" type here? We don't have that in the v3 specs?

openmeter/productcatalog/subscription/service/create.go
```diff
@@ -2,31 +2,46 @@ package service
...
 	"github.com/openmeterio/openmeter/openmeter/subscription"
 	"github.com/openmeterio/openmeter/pkg/clock"
+	"github.com/openmeterio/openmeter/pkg/featuregate"
 	"github.com/openmeterio/openmeter/pkg/models"
 )
 
+// errCustomCurrenciesDisabled is returned when an inline plan uses a custom
+// currency but the credits feature (which gates custom currencies) is not enabled
+// on the deployment.
+var errCustomCurrenciesDisabled = errors.New("custom currencies are not enabled on this deployment of OpenMeter")
```
gergely-kurucz-konghq: This is defined here, but is used in `change.go` as well. Not a big issue, just from a code organization perspective a bit misleading co-location. Consider extracting it to a common place, in other parts of the codebase such sentinel errors are either added to a domain package local `errors.go` or put in `service.go`, or another appropriate parent file/directory.

In fact a similar error is defined at https://github.com/openmeterio/openmeter/blob/47e39f2b0d50814386082a2d4d39d194ae071283/api/v3/handlers/currencies/handler.go#L11, so this makes it a duplication.

api/v3/handlers/subscriptions/create.go
```diff
@@ -76,45 +76,36 @@ func (h *handler) CreateSubscription() CreateSubscriptionHandler {
...
-			// Convert the plan entity to a plan input
-			planInput := plansubscription.PlanInput{}
-			planInput.FromRef(&plansubscription.PlanRefInput{
-				Key:     planEntity.Key,
-				Version: &planEntity.Version,
-			})
-
 			// Convert the request to a create subscription workflow input
-			subscriptionName := fmt.Sprintf("%s v%d", planEntity.Key, planEntity.Version)
 			workflowInput, err := FromAPIBillingSubscriptionCreate(
```
gergely-kurucz-konghq, line 109: There is a suspicious todo comment in this convert function:
https://github.com/openmeterio/openmeter/blob/47e39f2b0d50814386082a2d4d39d194ae071283/api/v3/handlers/subscriptions/convert.go#L282-L285

Might be out of scope for this PR, but this suggests a disparity with the v1 implementation, so mentioning just in case.

However, in `change.go` this does not seem to be an issue, since the `timing` field is set in the handler code there. Might worth a look to ensure consistency.

gergely-kurucz-konghq (changes requested):
Minor things.

api/spec/packages/aip/src/subscriptions/subscription.tsp
rolosp: I didn't see one in the specs, i only see in the legacy package, that's why I created one inline.

api/v3/handlers/subscriptions/create.go
rolosp: I think this is out of scope but we can add a follow-up ticket for it.

openmeter/productcatalog/subscription/service/create.go
rolosp: pushed a fix for it

api/spec/packages/aip/src/subscriptions/subscription.tsp
gergely-kurucz-konghq: This can be a follow up improvement if needed.

api/v3/handlers/subscriptions/create.go
gergely-kurucz-konghq: Could be a follow up, please create a ticket for this. My understanding is that without fixing this, custom subscriptions won't accept an API provided timing, and will always use "Immediate".

gergely-kurucz-konghq (dismissed):

gergely-kurucz-konghq (approved):

